PR #14, opened 2017-04-11
Ss644 import isc robots

JamesGlover:


app/assets/javascripts/legacy_scripts_b.js
```diff
@@ -235,289 +235,4 @@
 
     $('.create-button').prop('disabled',false);
   });
-
```
JamesGlover: This has been moved into its own javascript file.

PR #21, opened 2017-05-10
Pf limber

JamesGlover:
7262acf is the config additiona
18d617b extends the robot specs
d3afa2f is a bit of refactoring of the specs following some of the http://www.betterspecs.org/ guidelines. The rest of the robot changes are mostly just reorganization of the file structure to avoid sticking too many classes in one file.


JamesGlover:
Merge just pulls in the changes from the other reviewed branches.

PR #26, opened 2017-06-07
Updated tube transfer and custom tubes

JamesGlover:
This branch isn't quite as clean as I'd like, and combines a couple of features.
78fef0b handles creation of tubes from tubes. While this wasn't actively used in limber until now, the code still existed. As a result we're just adding tests.
af22e55 This adds filtering to forms to ensure that they don't appear as options in the menu if pre-requisits aren't met. This primarily ensures that creators recieve the parent they expect, and you can't pool untagged plates. Without this the users just get a something went wrong, which is a bit inelegant.
2aed623- A bug fix. Had to roll this one out in the current UAT branch as well.
4bcef2d - f33ad41: Switch to creating transfer requests in bulk by defining source and target assets directly.
aa4b990 - 01a5297: Our transfer used to name or tubes. Now we do it ourselves on creation. The second commit is a precaution to ensure that we don't inadvertantly switch the order of children as a mismatch between a tubes name and its content would cause major confusion.


app/controllers/creation_controller.rb
```diff
@@ -4,7 +4,7 @@ class CreationController < ApplicationController
   before_action :check_for_current_user!
 
   def form_lookup(form_attributes = params)
```
Proskurina: creator_lookup?

app/controllers/creation_controller.rb
```diff
@@ -4,7 +4,7 @@ class CreationController < ApplicationController
...
-    LabwareCreators::Base.class_for(form_attributes.fetch(:purpose_uuid))
```
Proskurina: creator_attributes?

app/models/labware_creators/pooled_tubes_by_submission.rb
```diff
@@ -43,5 +51,36 @@ def pool_uuids
...
+      wells = pool_details['wells']
+      # Wells SHOULD already be sorted
+      "#{stock_plate_barcode} #{wells.first}:#{wells.last}"
+    end
+
+    def stock_plate_barcode
+      "#{parent.stock_plate.barcode.prefix}#{parent.stock_plate.barcode.number}"
+    end
+
+    delegate :pools, to: :parent
```
Proskurina: move to top

app/models/labware_creators/support_parent/plate_only.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
```
Proskurina: Could this be abstracted or is it a waste of time?

app/models/limber/plate.rb
```diff
@@ -29,16 +29,13 @@ def purpose
   # We know that if there are any transfers with this plate as a source then they are into
   # tubes.
   def transfers_to_tubes?
-    transfers_to_tubes.present?
+    transfer_request_collections.present? || transfers_to_tubes.present?
   end
 
   def tubes_and_sources
```
Proskurina: rename to 'tubes_and_source_wells' (?)
and maybe add some documentation on what it returns and what it is doing?

PR #74, opened 2018-07-05
Seq-668 Added GnT plates and robot steps

radome:
Deploy:
`bundle exec rake config:generate`


JamesGlover:
Merged into mult submission branch

PR #109, opened 2018-11-20
Quad pooling

JamesGlover:
Note: Await UAT before merge
Currently has karma issues


JamesGlover:
Disagree with most of the CC duplication warning as its either:
1) Testing: where it would make tests less explicity
2) Boilerplate for api config: Not really sure there's much to be gained by abstracting out id and name

.babelrc
```diff
@@ -1,17 +1,22 @@
...
     }]
   ],
 
   "plugins": [
-    "syntax-dynamic-import",
-    ["transform-class-properties", { "spec": true }]
+    "@babel/syntax-dynamic-import",
+    ["@babel/plugin-proposal-class-properties", { "spec": true }]
   ]
 }
```
JamesGlover: Updating babel

Gemfile.lock
```diff
@@ -26,63 +26,64 @@ GIT
 GEM
   remote: https://rubygems.org/
```
JamesGlover: Lots of dependency updates

app/javascript/quadrant-stamp/index.js
```diff
@@ -43,6 +43,8 @@ document.addEventListener('DOMContentLoaded', () => {
    # once the majority of our components are vue based.
    */
   if ( document.getElementById('quadrant-stamp-page') ) {
+    axios.defaults.headers.common['X-CSRF-Token'] = document.querySelector('meta[name="csrf-token"]').getAttribute('content')
+    Vue.prototype.$axios = axios
     /* The files-list element isn't on all pages. So only initialize our
```
JamesGlover: Names axios rather than http to make it clearer when its actually used.

PR #293, opened 2019-10-02
Pipeline summaries

JamesGlover:


JamesGlover:
Getting a weid diff due to production being ahead of next-release

PR #370, opened 2020-02-21
Gpl292 duplex seq purposes config

andrewsparkes:
Initial version of purposes configuration.

### DIFF
--- a/config/purposes/duplex_seq.yml
+++ b/config/purposes/duplex_seq.yml
@@ -9,12 +9,6 @@ LDS Cherrypick:
   :presenter_class: Presenters::StockPlatePresenter
 LDS Stock XP:
   :asset_type: plate
-LDS Frag:
-  :asset_type: plate
-LDS A-tail:
-  :asset_type: plate
-LDS Lig:
-  :asset_type: plate
 LDS AL Lib:
   :asset_type: plate
   :label_template: plate_lds_al_lib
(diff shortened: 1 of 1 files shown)


config/purposes/duplex_seq.yml
```diff
@@ -0,0 +1,63 @@
...
+  :input_plate: true
+  :presenter_class: Presenters::StockPlatePresenter
+LDS Cherrypick:
+  :asset_type: plate
+  :stock_plate: true
+  :cherrypickable_target: true
+  :input_plate: true
+  :presenter_class: Presenters::StockPlatePresenter
+LDS Stock XP:
+  :asset_type: plate
```
KatyTaylor: presume the default for all the missing ones here is false? e.g. stock_plate, cherrypickable_target...

config/purposes/duplex_seq.yml
```diff
@@ -0,0 +1,63 @@
...
+LDS AL Lib:
+  :asset_type: plate
+  :label_template: plate_lds_al_lib
+  :file_links:
+  - name: 'Download Hamilton AL Lib to QC1 CSV'
+    id: 'hamilton_al_lib_to_qc1'
+  - name: 'Download Concentration (nM) CSV for Customer'
+    id: 'duplex_seq_al_lib_concentrations_for_customer'
```
KatyTaylor: is there a reason for this one to be prefixed with 'duplex_seq' and the above one not?

KatyTaylor: I guess makes sense if there are other 'AL Libs' that aren't for duplex seq

PR #386, opened 2020-03-10
Gpl295 duplex seq labware creator

KatyTaylor:
Still got to add the 'diluent_volume' into the fields to update, but that will be a tiny change.


KatyTaylor:
changed this to go into the GPL-290 branch, so we can do UAT before merging into develop

PR #387, opened 2020-03-11
Gpl290 duplex seq pipeline

KatyTaylor:
creating a draft pull request so we have an easy link to the diff, and can comment


app/models/labware_creators/pcr_cycles_binned_plate/csv_file/row.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    def pcr_cycles_within_expected_range?
+      in_range?('pcr_cycles', pcr_cycles, @row_config.pcr_cycles_min, @row_config.pcr_cycles_max)
+    end
+
+    def submit_for_sequencing_has_expected_value?
+      return true if empty?
+
+      return true if %w[Y N].include? @submit_for_sequencing_string
```
KatyTaylor: Just realised - shouldn't this be pulling those Y / N values from the config in 'PcrCyclesCsvFileUploadConfig' ?

andrewsparkes: Hmm, maybe, how then would we convert those config values (any string) to boolean?

app/models/labware_creators/custom_pooled_tubes/csv_file.rb
```diff
@@ -53,6 +54,20 @@ def header_row
...
+      # byte order marker will appear at beginning of in first string in @data array
+      s = @data[0][0]
+
+      # NB. had to make byte order marker string mutable here otherwise get frozen string error
+      bom = +"\xEF\xBB\xBF"
+      s_mod = s.gsub!(bom.force_encoding(Encoding::BINARY), '')
+
+      @data[0][0] = s_mod unless s_mod.nil?
+    end
+
```
KatyTaylor, line 70: This looks OK to me - at least it's safe as will work if the BOM is there or not there.

Only other thing I've come across - something like this might work:
@data = CSV.parse(File.read(file, encoding: 'bom|utf-8'))

But I haven't tested it as was struggling to actually produce a file with a BOM.

KatyTaylor: Tried it with your tests and that works, actually - might be a neater solution?

andrewsparkes: Yes I saw that version as well. I didn't test it though. I wasn't sure what would happen if you didn't have a utf-8 encoded file? I think the test files are probably valid for utf-8 so work, but maybe some real files may not be depending on if you're on mac or windows and what app you've used?
There are 2 bom files in the spec/fixtures/files directory in this branch, with bom in the name. Easiest to copy the file, then edit the copy without deleting the first line completely, seemed to work.
fyi We also had some weirdness with emailing files to one another. As if the emailing process itself opened and re-saved the file and changed encoding. Spam/virus filtering?

KatyTaylor: I'm not sure. I think it might act the same as your current code, because I think the File.read will assume it's utf-8 by default. But might be wrong. Maybe worth trying to write another test with a file that is not utf-8, if we have time.

Yes thanks, after writing latest comment I realised you'd provided one in the fixtures directory!

PR #403, opened 2020-03-27
Gpl 351 bed verification fix require robot

KatyTaylor:


app/models/robots/robot.rb
```diff
@@ -74,18 +75,35 @@ def bed_plates=(bed_plates)
...
-        if beds.values.first.plate.custom_metadatum_collection.nil?
-          error_messages << 'Your plate is not on the right robot'
-          return false
-        elsif beds.values.first.plate.custom_metadatum_collection.metadata['created_with_robot'] != robot_barcode
-          error_messages << 'Your plate is not on the right robot'
-          return false
-        end
+      return false unless robot_present_if_required
+
+      return true unless verify_robot? && beds.values.first.plate.present?
```
andrewsparkes: Shouldn't this be return true if rather than unless?

app/models/robots/robot.rb
```diff
@@ -74,18 +75,35 @@ def bed_plates=(bed_plates)
...
+
+      if missing_custom_metadatum_collection || original_robot != robot_barcode
```
andrewsparkes: Does this line work ok if original robot is not set? e.g. nil != '1234'

app/models/robots/robot.rb
KatyTaylor: No, it's the equivalent of the `if verify_robot? && beds.values.first.plate.present?` that was there previously.
i.e. You only want to bother doing the validation to check if the robot is the same as before IF the verify_robot flag is set and there is a plate barcode scanned in

I'll change it back to how it was before as think that's more readable

andrewsparkes: Ah I see :) 
I read verify_robot as a method you'd always want to run rather than a flag, so returning true if the robot was invalid seemed wrong. It's fine as the guard clause.

app/models/robots/robot.rb
KatyTaylor: It's equivalent to what was there before isn't it?

I can see a case that isn't explicitly dealt with where missing_custom_metadatum_collection is false, but there isn't one with the key 'created_with_robot' - looks like that would end up with nil != '1234', so add an error message, which I think is probably correct?

andrewsparkes: Yep it's fine, the guard clause I was confused about above stops it getting there unless we want the robot to be the same. Ignore me.

PR #408, opened 2020-04-01
Gpl 369 rename heron library type

KatyTaylor:
Merge Andrew's PR first, then the diff will make more sense

### DIFF
--- a/config/pipelines/high_throughput_heron.yml
+++ b/config/pipelines/high_throughput_heron.yml
@@ -2,7 +2,7 @@
 Heron:
   filters:
     request_type_key: limber_heron
-    library_type: Heron
+    library_type: PCR amplicon ligated adapters
   library_pass: LHR Lib PCR
   relationships:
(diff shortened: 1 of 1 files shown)


config/pipelines/high_throughput_heron.yml
```diff
@@ -2,7 +2,7 @@
 Heron:
   filters:
     request_type_key: limber_heron
-    library_type: Heron
+    library_type: PCR amplicon ligated adapters
```
harrietc52, line 5: Assume capitals don't matter? :)

KatyTaylor: Yeah David put 'PCR Amplicon ligated adapters' but I thought it was inconsistent having a capital 'A' and not the other ones so I made them all lower case

andrewsparkes: Just a string I think, that is what they asked for.

PR #409, opened 2020-04-01
Gpl 373 bed verify RT to PCR

KatyTaylor:


spec/models/robots/heron_robot_pcr_destinations_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        { 'bed9_barcode' => [source_barcode],
+          'bed4_barcode' => [bed_4_PCR_plate_barcode_2],
+          'bed6_barcode' => [bed_6_PCR_plate_barcode_2] }
+      end
+
+      it 'should not be valid' do
+        is_expected.not_to be_valid
+        expect(subject.message).to include 'The PCR plates must have the same barcode as the RT plate, plus a PP suffix.'
+      end
+    end
```
andrewsparkes, line 146: Maybe also a test on both source and destination robots where the plates are all for the same plate, but PCR 1 and 2 are the wrong way around. Should be invalid. Think we tried this yesterday and it was fine, but worth have a test for as it;s a key reason the bed verifications are there.

PR #417, opened 2020-04-09
Gpl375 pcr plate split and merge

### LINKED ISSUE #410: GPL-375 As R&D we would like to track the 2 PCR plates as there is a handover at this point between labs and teams and it is likely that there will be mistakes especially as throughput increases
**User story**
'We tried printing the labels and we got the RT label along with the PCR1 and PCR2 labels. When we scanned into LIMBER the PCR labels weren't recognised, but the RT label did and was going straight from RT to the LHR XP plate. Should there be a step in between where the RT plate generates the PCR plate and then PCR to XP plate?

PSD decided to not track the 2 PCR plates in Limber for now, as it made development quicker. The bed verification step should still recognise the PCR plates barcodes though.

 We are slightly concerned about not tracking the PCR plates as there is a handover at this point between labs and teams. We can manage it via paperwork and training, but it’s likely ops will ask for changes to this part of the pipeline in the future.'

[
Bed verification does recognise the PCR plates.

Deliberately designed this way for version 1 but clearly not communicated properly to Lesley. Adding the Limber plate creators to make the PCR plates and do the merge of the two PCR plates into the XP plate was too expensive in development time for version 1, plus existing restrictions when merging aliquots means the solution may break other Limber pipelines. 
Andrew.
]

**Who the primary contacts are for this story**
Lesley Shirley

**Acceptance criteria**
1. PCR plates 1 and 2 are both tracked in Limber and can be scanned
2. Bed verifications for RT to PCR 1 and 2, and from PCR 1 and 2 to XP work.
3. Aliquots from PCR 1 and 2 are merged down to 1 aliquot per well in the XP plate.
4. Primer panel choice from the Submission is retained.

**Sub-tasks**
1. Sequencescape and Limber - Research how to do the merge of aliquots without breaking other pipelines. [James has given details of how this is done in another pipeline. Plus here the aliquots will now have the same primer panel name and we can merge two aliquots down to 1, reducing complexity later in the pipeline. **WIP**]
  - Sequencescape currently enforces a uniqueness constraint on all aliquots within a receptacle such that they have unique tag-tag2 combinations. (untagged aliquots are represented by the id -1, as MySQL acts as though null != null) This is enforces at the database level. 
  - If we were to track primer panels independently this would violate this constraint, as primer panels are being used prior to tagging. The safest fix for this would be to add primer_panel_id to the uniqueness key, and use -1 rather than null to indicate no primer panel.
  - This would slightly increase the risk of tag clashes in GBS if the same tags were used for two different primer panels. Although I think we do enforce these checks as part of the plate creation.
  - If we DON'T track two separate primer panels, and continue with the current system, then we should follow a similar practice to that used in the FlexibleCherrypick pipeline. Here we create transfer requests from both parents, but de-duplicate aliquots with the same information.
2. Limber - modify configuration to allow branching at the RT plate and set the purposes for the PCR plates. [**DONE**]
3. Limber - labware creator for PCR plates, one creator to handle both types. Potentially allow override of submission primer panel choice if required. Otherwise this may be a simple stamp. Will they select 2 distinct primer panel ids, one for each plate? [Simple stamp **DONE**]
4. Limber - labware creator for PCR plates to XP plate. Select / scan both PCR plates and merge aliquots to create XP plate. [**DONE**]
5. Limber - modify bed verifications for both steps [**DONE**]
6. Integration Suite - modify heron test to perform these extra steps. [**DONE**]
7. Limber - Help panel on labware creator to take purposes from heron purposes configuration to make more user-friendly help message possible i.e. Scan the following plates: LHR PCR 1 and LHR PCR 2. [**Not Started**]
8. Sequencescape - Fix bug in tag clash rescue code that is allowing silent failure. [**WIP**] Our understanding is that SS is not returning an http error when a tag clash is happening, so Limber thinks that everything is correct. From API v1, TransferRequestCollection.create should raise a Aliquot::TagClash exception inside an after_create callback, however, that exception is either rescued or ignored by API v1 before reaching the ErrorHandler middleware in Sinatra that would generate a http error.
9. Sequencescape - Add de-duplication for identical aliquots in transfer requests for merging plates without breaking other pipelines. [**Not Started**]

**Additional context**
Need to clarify details with R&D / Ops:
1. Do they want the ability to set / override a distinct primer panel selection for PCR 1 and 2? **NO. Naomi clarified on Friday 3rd April.**
2. If 1, do they want to set the two distinct primer panel choices in the Submission? And what is put into MLWH? **N/A**

andrewsparkes:
Added split and merged plate creator and tests for splitting the LHR RT into the LHR PCR 1 and 2 then merging the LHR PCR plates into the XP
closes #410


config/robots.rb
```diff
@@ -1483,62 +1483,47 @@
...
                  },
                  bed(4).barcode => {
-                   purpose: 'LHR RT',
-                   states: ['passed'],
+                   purpose: 'LHR PCR 1',
+                   states: ['pending'],
                    label: 'Bed 4',
-                   display_purpose: 'LHR PCR 1',
-                   override_class: 'Robots::Bed::Heron',
-                   expected_plate_barcode_suffix: 'PP1'
```
KatyTaylor, line 1491: Consider reversing the changes we made for `override_class` in app/models/robots/robot.rb and app/models/robots/bed/base.rb ?

And also for `display_purpose` in app/models/robots/bed/base.rb

config/purposes/heron.yml
```diff
@@ -7,8 +7,18 @@ LHR RT:
   :default_printer_type: :plate_a
   :label_template: plate_lhr_rt
   :presenter_class: Presenters::StockPlatePresenter
+LHR PCR 1:
+  :asset_type: plate
+LHR PCR 2:
+  :asset_type: plate
```
KatyTaylor, line 13: Oh nice, so you didn't need a new creator for the PCR plates, because they are the same as each other, they will just have 2 buttons to click?

app/models/labware_creators/merged_plate.rb
```diff
@@ -0,0 +1,108 @@
...
+    def source_plates_have_same_parent?
+      return if source_plates.map { |sp| sp.parents.map(&:id) }.flatten.uniq.one?
+
+      msg = 'The source plates have different parents, please check you have scanned the correct set of source plates.'
+      errors.add(:parent, msg)
+    end
+
+    # Validation to check the user hasn't scanned the same barcode multiple times
+    def source_barcodes_are_different?
+      duplicates = minimal_barcodes.select { |e| minimal_barcodes.count(e) > 1 }
```
KatyTaylor, line 92:
```suggestion
      duplicates = minimal_barcodes.any? { |e| minimal_barcodes.count(e) > 1 }
```
Could you just return a Boolean instead? Slightly simpler?

app/models/labware_creators/merged_plate.rb
```diff
@@ -0,0 +1,108 @@
...
+      return if duplicates.uniq.empty?
+
+      msg = 'The source plates should not have the same barcode, please check you scanned all the plates.'
+      errors.add(:parent, msg)
+    end
+
+    # Validation to check the user hasn't accidently created multiple plates wirh the same purpose
```
KatyTaylor, line 99:
```suggestion
    # Validation to check the user hasn't accidently created multiple plates with the same purpose
```

app/models/robots/bed/base.rb
```diff
@@ -47,11 +46,7 @@ def barcode
     def load(barcodes)
       # Ensure we always deal with an array, and any accidental duplicate scans are squashed out
       @barcodes = Array(barcodes).map(&:strip).uniq.reject(&:blank?)
-      @plates = if @barcodes.present?
-                  Sequencescape::Api::V2::Plate.find_all({ barcode: @barcodes }, includes: plate_includes)
-                else
-                  []
-                end
```
KatyTaylor, line 54: Don't think you need to reverse this one - think this was an optimisation made by @JamesGlover to avoid an unnecessary API call to SS in the case where there are no barcodes to search on?

PR #420, opened 2020-04-16
Improve factories

JamesGlover:

* Extend documentation of factories
* Remove some unused factories
* Add recording of factory statistics
* Update FactoryBot
* Fix SimpleCov warnings
* Low risk Rubocop fixes


app/controllers/tubes/work_completions_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# Handles the clicking of the 'Charge and Pass Libraries' button for plates
```
KatyTaylor, line 3:
```suggestion
# Handles the clicking of the 'Charge and Pass Libraries' button for tubes
```

spec/factories/barcode_printer_factories.rb
```diff
@@ -1,19 +1,26 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
+  # Build an API v1 barcode printer. YOu probably want to use: plate_barcode_printer or
```
KatyTaylor, line 4:
```suggestion
  # Build an API v1 barcode printer. You probably want to use: plate_barcode_printer or
```

spec/factories/pipeline_factories.rb
```diff
@@ -1,12 +1,17 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
+  # Construct a Pipeline
+  # Use create to automatically push it onto the pipelines array
+  # build just returns the pipleine object
```
KatyTaylor, line 6:
```suggestion
  # build just returns the pipeline object
```

spec/factories/plate_factories.rb
```diff
@@ -5,23 +5,44 @@
...
 
     transient do
+      # The number of wells on the plate. Usualy calculated from
+      # the plate dimensions.
       well_count { number_of_rows * number_of_columns }
+      # The factory to use for wells on the plate
       well_factory { :v2_well }
+      # The factory to use for requests associated with the plate.
+      # For v2_plates this is associated with the aliquot, for v2_stock_plates
+      # these requests are comming from the wells themselves
```
KatyTaylor, line 32:
```suggestion
      # these requests are coming from the wells themselves
```

spec/factories/plate_factories.rb
```diff
@@ -43,39 +68,56 @@
...
+    custom_metadatum_collection { nil }
 
     # Mock the relationships. Should probably handle this all a bit differently
     after(:build) do |plate, evaluator|
-      RSpec::Mocks.allow_message(plate, :wells).and_return(evaluator.wells)
-      RSpec::Mocks.allow_message(plate, :purpose).and_return(evaluator.purpose)
-      RSpec::Mocks.allow_message(plate, :custom_metadatum_collection).and_return(evaluator.custom_metadatum_collection)
+      plate._cached_relationship(:wells) { evaluator.wells }
+      plate._cached_relationship(:purpose) { evaluator.purpose }
+      plate._cached_relationship(:custom_metadatum_collection) { evaluator.custom_metadatum_collection }
```
KatyTaylor, line 117: How does this _cached_relationship work?
Also, is the `# Should probably handle this all a bit differently` comment still accurate after making this change?

spec/factories/plate_factories.rb
```diff
@@ -84,9 +126,14 @@
         evaluator.ancestors.select { |a| parameters[:purpose_name].include?(a.purpose.name) }
       end
       RSpec::Mocks.allow_message(plate, :ancestors).and_return(ancestors_scope)
-      RSpec::Mocks.allow_message(plate, :parents).and_return(evaluator.parents)
+      plate._cached_relationship(:parents) { evaluator.parents }
     end
 
+    # Set up a stock plate. Changes:
+    # - The plate purpose
+    # - The well factory to v2_stock_well which sets requests comming out of the wells,
```
KatyTaylor, line 134:
```suggestion
    # - The well factory to v2_stock_well which sets requests coming out of the wells,
```

spec/factories/plate_factories.rb
```diff
@@ -84,9 +126,14 @@
         evaluator.ancestors.select { |a| parameters[:purpose_name].include?(a.purpose.name) }
       end
       RSpec::Mocks.allow_message(plate, :ancestors).and_return(ancestors_scope)
-      RSpec::Mocks.allow_message(plate, :parents).and_return(evaluator.parents)
+      plate._cached_relationship(:parents) { evaluator.parents }
     end
 
+    # Set up a stock plate. Changes:
```
KatyTaylor, line 132: When it says 'changes', what does it mean? Changes with respect to what? Cheers

spec/factories/plate_factories.rb
```diff
@@ -98,13 +145,19 @@
...
+      initialize_with do
+        Sequencescape::Api::V2::Plate.load(attributes)
+      end
       transient do
         purpose_name { 'Primer Panel example' }
         request_factory { :gbs_library_request }
       end
     end
 
+    # Sets up a plate of 2 isd requests, complete with bait libraries
```
KatyTaylor, line 160: What's an isd request?

spec/factories/plate_factories.rb
```diff
@@ -126,6 +183,8 @@
       end
     end
 
+    # Sets up a plate at the begining of the agregation process
```
KatyTaylor, line 186:
```suggestion
    # Sets up a plate at the beginning of the agregation process
```

spec/factories/purpose_config_factories.rb
```diff
@@ -101,11 +116,14 @@
       end
     end
 
+    # Configuration for an agregation plate
```
KatyTaylor, line 119:
```suggestion
    # Configuration for an aggregation plate
```

spec/factories/request_factories.rb
```diff
@@ -4,15 +4,23 @@
...
 
 FactoryBot.define do
+  # A basic API V2 request. The factories inheriting from this
+  # are probably more useful.
   factory :request, class: Sequencescape::Api::V2::Request, traits: [:uuid] do
     transient do
       pcr_cycles { 10 }
       sequence(:submission_id) { |i| i }
       sequence(:order_id) { |i| i }
+      # Specifies that submision information should be included in the model
```
KatyTaylor, line 14:
```suggestion
      # Specifies that submission information should be included in the model
```

spec/factories/request_factories.rb
```diff
@@ -52,22 +60,29 @@
...
+    # Basic library creation request, such as wgs
     factory :library_request do
       request_type { create :library_request_type }
 
+      # Library request with primer panel information
       factory :gbs_library_request do
         primer_panel
       end
 
+      # Library request with bait libray information
```
KatyTaylor, line 72:
```suggestion
      # Library request with bait library information
```

spec/factories/request_factories.rb
```diff
@@ -52,22 +60,29 @@
...
     end
 
+    # Multiplexing request, representing the pooling steps at the end of most
+    # pipelines.
     factory :mx_request do
       request_type { create :mx_request_type }
     end
 
+    # Aggregation request, representing the transfer of many plates onto
+    # one at the begining of the process
```
KatyTaylor, line 85:
```suggestion
    # one at the beginning of the process
```

KatyTaylor:
Which are the bits that improve the test speed significantly? Cheers

app/sequencescape/sequencescape/api/v2/request.rb
```diff
@@ -3,6 +3,8 @@
 class Sequencescape::Api::V2::Request < Sequencescape::Api::V2::Base
   FragmentSize = Struct.new(:from, :to)
 
+  has_one :submission
```
KatyTaylor, line 6: where does this get used? thanks

lib/tasks/test_factories_lint.rake
```diff
@@ -1,12 +1,41 @@
...
-      puts 'Linting factories...'
-      FactoryBot.lint
-      puts 'Done'
+      require 'webmock'
+      include WebMock::API
+
+      WebMock.enable!
+      # We currently use mocks in the factories to handle building API
+      # objects. We should look at better ways of handling this, but until
+      # then we enable mocks while linting.
```
KatyTaylor, line 16: What are the better ways of handling this?

spec/factories/shared_traits.rb
```diff
@@ -5,50 +5,86 @@
...
 
-  trait :api_object do
+  # Base trait for building V1 API classes, do not use directly, instead use
+  # :api_object or :api_simple_object
+  trait :_api_object do
+    skip_create
+
     transient do
-      api_root { 'http://example.com:3000/' }
+      api_root { 'http://example.com:3000/' } # The root of the Sequenbcescape API V1
```
KatyTaylor, line 14:
```suggestion
      api_root { 'http://example.com:3000/' } # The root of the Sequencescape API V1
```

spec/factories/transfer_request_collection_factories.rb
```diff
@@ -1,13 +1,21 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
+  # API V1 collection of multiple transfer requests
+  # This allows easy creation of multiple transfers between arbitary receptacles
```
KatyTaylor, line 5:
```suggestion
  # This allows easy creation of multiple transfers between arbitrary receptacles
```

spec/factories/transfer_request_collection_factories.rb
```diff
@@ -1,13 +1,21 @@
...
   factory :transfer_request_collection, class: Sequencescape::TransferRequestCollection, traits: [:api_object] do
     json_root { 'transfer_request_collection' }
 
     transient do
+      # The number of transfers to create (Will be generated in column order for wells)
       transfer_count { 2 }
+      # Number of diffent targets expected. For example, setting this to 1 will result in
```
KatyTaylor, line 12:
```suggestion
      # Number of different targets expected. For example, setting this to 1 will result in
```

spec/factories/well_factories.rb
```diff
@@ -20,64 +23,96 @@
...
+      # on outer request.
       requests_as_source { [] }
+      # Requests terminating at the well. Generally only seen on the final
+      # plate of the process after libraries are passed.
       requests_as_target { [] }
+      # The state of the ongoing requests.
       library_state { 'pending' }
+
+      # Set up relationships downstream
+      # In Sequencesacape world these are all populated via the
```
KatyTaylor, line 71:
```suggestion
      # In Sequencescape world these are all populated via the
```

spec/factories/plate_factories.rb
JamesGlover: Good question, I'll pop a description in the code, as it will likely cause similar questions in future.
The # Should probably handle this all a bit differently is still true, although less so now, and the alternatives have their own issues.

JamesGlover: Explanation added

spec/factories/plate_factories.rb
JamesGlover: Relative to the parent factory for a standard plate. Updated the comment.

spec/factories/plate_factories.rb
JamesGlover: Typo, should have said ISC (Indexed Sequence Capture). Fixed typo and added expanded version of the abbreviation to comment.

app/sequencescape/sequencescape/api/v2/request.rb
JamesGlover: Setting up the relationship allows us to extract the submission id in LabwareCreators::WellFilter and the uuids when building the work orders.
This isn't entirely necessary when dealing with actual data coming across the API, as it will automatically configure the relationship based on the data, but it is important it we end up doing something like Sequencescape::Api::V2::request.new, such as when setting up data for tests.
Previously this slight difference in behaviour was hidden by the mocking that was going on.

JamesGlover: Ahh, actually it looks like this was one of the relationships I needed to keep mocking in for. I was having a few issues letting us both extract submission ids from the relationship, and access the actual object itself when we needed it. Works fine in actual code, but had issues setting up items in the right state.

lib/tasks/test_factories_lint.rake
JamesGlover: I'm not 100% sure. The comment was initially made before I made the changes which eliminated most of the mocks in the factories, but a few still remain.

One prospect would be to generate json-api output and get the gem to parse this as though it came in via the API directly. This would ensure that our test object behaved exactly like our real ones, but at the expense of added complexity (and probably reduced speed) in the factories.

It still causes a bit of trouble with the ancestor scope object though, where we end up mocking the results of the query. Not sure how to handle this one without mocks. Could use webmock, but that feels a bit like replacing one smell with another. It probably indicates that ancestors don't belong in the factories, and should be set up separately in the tests themselves.

KatyTaylor: I'm not too familiar with all the options, but opinions from what I've encountered so far:

I enjoyed making changes like this - https://github.com/sanger/asset_audits/pull/132/files#diff-0b60e12263cf050d50a27eda10159846 - where the old mocking was v verbose and not v readable. 
Seems like generating json-api output would bring back in that complexity again? Maybe not, though, if we wrote some nice helper methods.

Eduardo said, in a conversation about the mocking for asset audits, 
"yes, better to use that mocha
webmock only if you need it for any reason, but if not, better avoid it
because any change in http metadata will break the mock"

so probs worth having a wider chat with the team about best way to mock api calls?

PR #435, opened 2020-05-19
GPL-466 Adds qc configuration for heron 384

### LINKED ISSUE #434: GPL-466b New quant type for Heron 384 cDNA - Limber part
**User story**
Companion to issue for Quanthub. Limber needs to print a QC barcode on the XP plate, and have a button for downloading the csv concentrations file on the same plate.

**Who the primary contacts are for this story**
Adam

**Acceptance criteria**
1. I can print a QC label in addition to the normal print label
2. I can download a csv for concentrations in ng/ul

**Additional context**
Can be deployed independently of Quanthub.

andrewsparkes:
See also corresponding Quanthub issue.
Closes #434


app/models/labels/plate_double_label.rb
```diff
@@ -1,26 +1,5 @@
...
-  end
-
-  def default_printer_type
-    default_printer_type_for(:plate_double)
-  end
-
-  def default_label_template
-    default_label_template_for(:plate_double)
-  end
+# Prints the labels for a 384 well plates
```
KatyTaylor, line 3:
```suggestion
# Prints the labels for 384 well plates
```

app/models/labels/plate_double_label_qc_base.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
KatyTaylor, line 1: Why make a `PlateDoubleLabelQcBase` class rather than bunging this code into the `PlateDoubleLabelQc`class? Do you expect to make more subclasses in future?

andrewsparkes: It's done like that for the 96 well plate, wasn't completely sure why, but it is future proof so copied that. What do you think? Should change both 96 and 384 ones to be consistent if we do that. Some risk of 96 plate labels breaking then?

KatyTaylor: I would vote for simplifying them, but don't mind. I think it's fairly safe - we are removing a class so if we search for references of that class should be ok? So search for `PlateLabelXpBase` ?

PR #447, opened 2020-05-27
RT #690877 label printing bug

KatyTaylor:
return an array - this was missed when converting qc_attributes to an array in this pull request - https://github.com/sanger/limber/pull/371/files?file-filters%5B%5D= (GPL-294)


pjvv:
Good spot @KatyTaylor !

PR #454, opened 2020-06-03
Update dependencies

JamesGlover:
- Updates Ruby to 2.6.6
- Updates Rails to 5.2.4.3
- Updates minor gems
- Updates Javascript dependencies
- Simplifies Webpack config
- Inclines Rubocop issues


.gitignore
```diff
@@ -20,3 +20,10 @@ REVISION
 yarn-error.log
 .tern-project
 *.swn
+
```
JamesGlover, line 23: Added by Rails as part of its upgrade process

.browserslistrc
```diff
@@ -0,0 +1 @@
+defaults
```
JamesGlover, line 1: I could swear we had set this up somewhere with different options. Can't track it down though.

Issue #483, opened 2020-09-07
GPL-638 new Library PCR consolidation


rl15:
**User story**
GPL-638 | As a member of R&D (Lesley, Iraad, Emma B) we would like a new Library PCR consolidation to unite the parallel PCR paths for one sample prior to pooling

**Who are the primary contacts for this story**
Lesley S
Iraad B
Emma B

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] add a list of acceptance criteria here
- [ ] ...

**Risk?***
Tag clash (is same sample ID)

**Additional context**
Workflows available: https://ssg-confluence.internal.sanger.ac.uk/x/u4meBg

JamesGlover:
Again, this one should *just work*. Moving into In progress until I discover otherwise.

JamesGlover:
From earlier discussions with Lesley it sounds like they expect to use the same 'one primer panel' option as in the original pipeline.

JamesGlover:
😭  Not free after all, and sadly a little messy.

1. Aliquot deduplication depends on the aliquots sharing the same properties, such that they can be consolidated into a single record.
2. While we share tags, samples, library_types, primer panels, we have different library_ids
3. Library id is set when tagging occurs, and it points to the receptacle in which the tag was first applied.
4. This means that the aliquots can not be de-duplicated, as their attributes are different. (Ie. they aren't de-duplicates)
5. Trying to insert them both into the aliquots table results in a tag clash, as they have identical tags.

Other notes:
1. Tag clashes are enforced at the database level by a uniqueness constraint on receptacle_id,tag_id,tag2_id

Options:
1. Somehow consolidate library_ids at the merge, to only have a single library. (Possibly using the merge well id.)
  Results in single aliquot, but slightly hampers tracking, and will break tag substitution unless we also update the library ids in the parent wells?
2. Set the library ids earlier in the process, such as on the initial transfer using the source receptacle id. This is similar to point 1, but has the advantage that tag substitution works again... as long as both versions of the library need the same substitution. (But if they don't, you'd be in trouble anyway, as suddenly one library indeed becomes two)
2. Add library_id to the uniqueness constraint. This is a pretty simple fix overall, and would result in two aliquots with different library ids, but all other options the same. It feels like the most accurate reflection of what's happened, but I'm not sure it'll gel great with the NPG view of the world, as they don't represent entities that can be deplexed.

I had another idea but its slipped my mind as I'm writing this.

Issue #482, opened 2020-09-07
GPL-637 New Library construction to form the Lib PCR plates


rl15:
**User story**
GPL-637 | As a member of R&D (Lesley, Iraad, Emma B) we would like a new library construction step where tags are added to the PCR plates to form the Lib PCR plates

**Who are the primary contacts for this story**
Lesley S
Iraad B
Emma B

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] add a list of acceptance criteria here
- [ ] ...

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Workflows available: https://ssg-confluence.internal.sanger.ac.uk/x/u4meBg

JamesGlover:
Checked with @andrewsparkes and @KatyTaylor and sounds like the required behaviour maps up to the existing. So we get this for free.

JamesGlover:
Moved to in progress to avoid confusion, will stick it back in Todo is new requirements crop up.

JamesGlover:
Not quite as free as I hoped, but pretty cheap.

The current tag-plate validation was tripping up on the second plate, and tried to enforce UDI plates. However, fortunately the plate was still identified as not a 'cross-plate' pool (Which as far as we're concerned, it isn't, as the samples *will* have the same tags). Just updated UDI enforcement to only active if the submission has an existing tag plate associated with it AND actually needs UDIs.

PR #488, opened 2020-09-17
Gpl 635 new 384 tailed heron workflow

### LINKED ISSUE #480: GPL-635 New 384 tailed heron workflow
**User story**
GPL-635 | As a member of R&D (Lesley, Iraad, Emma B) we would like a new pipeline to support a tailed heron workflow in a 384 well format to reduce TRT

Lesley S
Iraad B
Emma B

**Acceptance criteria**

- [x] Library type must be ~~Heron LTHR-384~~ PCR amplicon tailed adapters 384
- [x]  Request type must be ~~Heron LTHR-384~~ Limber Heron LTHR
- [x] Plate purposes:
~~Heron~~ LTHR-384 RT
~~Heron~~ LTHR-384 PCR 1
~~Heron~~ LTHR-384 PCR 2
~~Heron~~ LTHR-384 Lib PCR 1 [Tagged Plate]
~~Heron~~ LTHR-384 Lib PCR 2 [Tagged Plate]
~~Heron~~ LTHR-384 Lib PCR pool [Merged Plate]
- [x] Tube Purposes
LTHR-384 Pool XP (Picked Pool XP not Lib Pool XP for consistency with existing Heron)

see word document for bed verifications, from link below

**Additional context**
Workflows available:  https://ssg-confluence.internal.sanger.ac.uk/x/u4meBg

JamesGlover:
Closes #480, #482, #483

* Adds Limber LTHR pipeline
* Improves UDI handling on tag plate usage
* Improves handling of large error messages


config/pipelines/high_throughput_heron_384_tailed.yml
```diff
@@ -0,0 +1,26 @@
+---
+Heron-384 Tailed A: # Heron 384-well pipeline specific to PCR 1 plate
```
KatyTaylor, line 2: I feel like I might have asked this before, but do you know why it's got that word 'tailed' in there? What does it mean?

config/pipelines/high_throughput_heron_384_tailed.yml
```diff
@@ -0,0 +1,26 @@
...
+  filters:
+    request_type_key: limber_heron_lthr
```
KatyTaylor, line 4: Why do we need a new request type? Just noticing the current 96 and 384 pipelines have the same, and have forgotten the implications of request types vs library types etc.

Also, could it be called limber_heron_tailed, or is that too long? More human-readable.

app/controllers/concerns/flash_truncation.rb
```diff
@@ -0,0 +1,62 @@
...
+  # @param [Integer] max_size The maximum allowed flash size
+  #
+  # @return [Array, String] The truncated message
+  #
+  def truncate_flash(message, max_size = max_flash_size)
+    return message if message.to_json.bytesize <= max_size
+
+    case message
+    when String
+      message.truncate(max_size - STRING_OVERHEAD)
```
KatyTaylor, line 31: if this is coming from `truncate_flash_array`, and `max_size` here is 2 or below, this will output '...'
Will this be a problem?

app/models/labware_creators/merged_plate.rb
```diff
@@ -76,32 +76,70 @@ def all_source_barcodes_entered?
...
       msg = 'Please scan in all the required source plate barcodes.'
-      errors.add(:parent, msg)
+      errors.add(:base, msg)
     end
 
-    # Validation to check all source plates have the same parent
-    def source_plates_have_same_parent?
-      return if source_plates.map { |sp| sp.parents.map(&:id) }.flatten.uniq.one?
+    # Validation to check all source plates can be merged.
+    # Currently we allow this is there is a maximum of one request associated with each well
```
KatyTaylor, line 83:
```suggestion
    # Currently we allow this if there is a maximum of one request associated with each well
```

app/models/labware_creators/merged_plate.rb
```diff
@@ -76,32 +76,70 @@ def all_source_barcodes_entered?
...
+    end
+
+    #
+    # Build up a hash of all merges that we expect to resolve
+    #
+    # @param [Array] source_plates Array of {Sequencescape::Api::V2::Plate} to be merged
+    #
+    # @return [Hash] Hash of each well/tag combination as a key, with an array of associated aliquots information
+    #                [request_id, suboptimal]
+    #                @example { 'A1' => ['1', false] }
```
KatyTaylor, line 102: This comment isn't correct I think, because the key is just the well position?

app/models/labware_creators/merged_plate.rb
```diff
@@ -76,32 +76,70 @@ def all_source_barcodes_entered?
...
+          merges[merge_index(well, aliquot)] << aliquot.equivalent_attributes
+        end
+      end
+      merges
+    end
 
-      msg = 'The source plates have different parents, please check you have scanned the correct set of source plates.'
-      errors.add(:parent, msg)
+    # All aliquots with the same merge_index should be capable of being merged.
+    # Currently with is restricted to *all* aliquots in the same well location.
```
KatyTaylor, line 116:
```suggestion
    # Currently this is restricted to *all* aliquots in the same well location.
```
?

app/models/labware_creators/merged_plate.rb
```diff
@@ -76,32 +76,70 @@ def all_source_barcodes_entered?
...
-    def source_plates_have_same_parent?
-      return if source_plates.map { |sp| sp.parents.map(&:id) }.flatten.uniq.one?
+    # Validation to check all source plates can be merged.
+    # Currently we allow this is there is a maximum of one request associated with each well
+    # Theoretically we can allow merging of plates with two distinct requests, as long as they have two
+    # different tag sets. See #merge_index for details
+    def source_plates_can_be_merged?
+      expected_merges = expected_merges(source_plates)
+
+      return unless expected_merges.values.any? { |v| v.uniq.many? }
```
KatyTaylor, line 89: Not sure what the `v.uniq.many?` is doing here?
So `v` would be, in your example below,  `['1', false]`
So `v.uniq.many?` would be true

Maybe can chat about it on Monday

config/pipelines/high_throughput_heron_384_tailed.yml
andrewsparkes: Best to keep request types for different pipelines (processes) distinct, plus confusing for user if more than one suggested action visible in Limber. Library types are mainly for NPG to tell libraries apart for analysis, but we can also use them in Limber to further determine suggested action.

config/pipelines/high_throughput_heron_384_tailed.yml
```diff
@@ -0,0 +1,26 @@
...
+    library_type: PCR amplicon tailed adapters 384
+  library_pass: LTHR-384 Lib PCR pool
+  relationships:
+    LTHR-384 RT: LTHR-384 PCR 1
+    LTHR-384 PCR 1: LTHR-384 Lib PCR 1
+    LTHR-384 Lib PCR 1: LTHR-384 Lib PCR pool
+Heron-384 Tailed B:
+  filters:
+    request_type_key: limber_heron_lthr
+    library_type: PCR amplicon tailed adapters 384
```
andrewsparkes, line 14: Is that library type correct for tailed?, looks same as other heron pipeline. Are the libraries essentially the same whether constructed by either pipeline?

config/purposes/heron_lthr_384.yml
```diff
@@ -0,0 +1,50 @@
+---
```
andrewsparkes, line 1: Is there any point where they need extra labels in this pipeline? for QC or the extra tag plates or anything...

andrewsparkes (approved):

config/robots.rb
```diff
@@ -1870,4 +1870,100 @@
...
+                 bed(4).barcode => {
+                   purpose: 'LTHR-384 Lib PCR 2',
+                   states: ['pending'],
+                   label: 'Bed 4',
+                   target_state: 'passed',
+                   parent: bed(3).barcode
+                 }
+               })
+
+  custom_robot('bravo-lhr-384-pcr-1-and-2-to-lhr-384-cdna',
```
KatyTaylor, line 1927: Don't think this robot should be here? Looks like copy-pasted from the existing mosquito version, and it morphed into the one below?

config/robots.rb
```diff
@@ -1870,4 +1870,100 @@
...
+                   label: 'Bed 6'
+                 },
+                 bed(8).barcode => {
+                   purpose: 'LTHR-384 Lib PCR pool',
+                   states: ['pending'],
+                   label: 'Bed 8',
+                   parents: [bed(4).barcode, bed(6).barcode],
+                   target_state: 'passed'
+                 }
+               })
```
KatyTaylor, line 1968: Did you double check the bed numbers here with R&D? (As they are different on the diagram)

config/robots.rb
```diff
@@ -1870,4 +1870,100 @@
...
+                 bed(8).barcode => {
+                   purpose: 'LHR-384 cDNA',
+                   states: ['pending'],
+                   label: 'Bed 8',
+                   parents: [bed(4).barcode, bed(6).barcode],
+                   target_state: 'passed'
+                 }
+               })
+  custom_robot('bravo-lthr-384-pcr-1-and-2-to-lthr-384-lib-pcr-pool',
+               name: 'bravo LTHR-384 Lib PCR 1 and 2 => Heron LTHR-384 Lib PCR pool',
```
KatyTaylor, line 1949: Probably don't need the word 'Heron' in the name here?

config/pipelines/high_throughput_heron_384_tailed.yml
KatyTaylor: Thanks. The existing 96 and 384 pipelines use the same request type, so I guess we avoid the "more than one suggested action visible in Limber" problem by also using the library type to distinguish between them?

So we could theoretically do the same here, but I guess it makes sense to have a new request type because they have developed a new process, whereas the existing 96 and 384 pipelines basically follow the same process (apart from different plate sizes).

app/models/labware_creators/merged_plate.rb
KatyTaylor: Should be array of arrays

config/pipelines/high_throughput_heron_384_tailed.yml
JamesGlover: It's something about the way the adapters work. I think its a case of the tags/adapters being built in to the primer-panel oligos themselves.

config/pipelines/high_throughput_heron_384_tailed.yml
JamesGlover: Yeah, we don't need one, but Lesley specifically requested one, mostly to try and avoid the SSRs accidentally mixing up library types (by pushing the main decision back to the submission template). Initially the requirements actually specified distinct request types for the two plate sizes, but we walked that one back to maintain consistency.

But @KatyTaylor is right, and the library_type would stop the users from seeing the other suggested actions. Its mostly the SSR this will affect. (It mean a separate submission template for the tailed, rather then narrowing down a shared template with specific library type options)

app/controllers/concerns/flash_truncation.rb
JamesGlover: You're right. I've specifically built in some overhead to allow for edge cases like this. I did consider handling this case, but considered:
1) If we reduced the size of '...' in response, it becomes increasingly unclear that the message has been truncated. If max size was 2, we'd just end up with an empty string, which doesn't communicate anything at all.
2) It seemed an unnecessary increase in code complexity, especially if we tried to also solve problem 1.

KatyTaylor: Fair enough. It was making my brain hurt slightly. Where did you build in the extra overhead?

app/models/labware_creators/merged_plate.rb
JamesGlover: Went over this in the zoom, but, to keep the answer on the record:

This was due to a mistake in the comment about expected_merges, which returns an array of arrays, not just an array. Fix coming with updated comment.

config/pipelines/high_throughput_heron_384_tailed.yml
JamesGlover: It has the word tailed, in place of ligated.

config/purposes/heron_lthr_384.yml
JamesGlover: Not that was mentioned in the requirements.

PR #527, opened 2020-10-15
Gpl 679 p wgs 384 pipeline

emrojo:
Closes #

*
*
* ...


config/purposes/pWGS_384_bottom.yml
```diff
@@ -0,0 +1,26 @@
+---
+pWGS-384 Post Shear XP:
+  :asset_type: plate
+  :creator_class: LabwareCreators::QuadrantStamp
+  :merger_plate: true
+  :size: 384
+pWGS-384 End Prep:
+  :asset_type: plate
+  :presenter_class: Presenters::PermissivePresenter
```
emrojo, line 9: Comment in class says: 
```ruby
  # Class PermissivePresenter provides a presenter which allows plate creation
  # even when the plate is pending
```
However the most similar plate to this is LHR-384 End Prep: which it also has this same class. I assume there is a reason why is needed. It would be worth to investigate in which situation is required, just in case we are using something we shouldn't be.

config/purposes/pWGS_384_bottom.yml
```diff
@@ -0,0 +1,26 @@
...
+  :size: 384
+pWGS-384 AL Lib:
+  :asset_type: plate
+  :size: 384
+pWGS-384 Lib PCR:
+  :asset_type: plate
+  :size: 384
+  :creator_class: LabwareCreators::TaggedPlate
+  :tag_layout_templates:
+  - 'TSsc384'
```
emrojo, line 19: The most similar purpose seems to be: LHR-384 Lib PCR. That one is using 
```yaml
  :presenter_class: Presenters::MinimalPcrPlatePresenter
```
That class does not have documentation, but naming suggests it could be related with this type of plate. Did you check this class already?

config/robots.rb
```diff
@@ -2024,4 +2024,103 @@
...
+                   purpose: 'LB Post Shear',
+                   states: ['passed'],
+                   label: 'Bed 15'
+                 },
+                 bed(7).barcode => {
+                   purpose: 'pWGS-384 Post Shear XP',
+                   states: ['pending'],
+                   label: 'Bed 7',
+                   parents: [bed(12).barcode, bed(13).barcode, bed(14).barcode, bed(15).barcode],
+                   target_state: 'passed'
```
emrojo, line 2058: I see robot: hamilton-star-pf-post-shear-to-pf-post-shear-xp-384
with an attribute class: 
```ruby
class: 'Robots::QuadrantRobot'
```
by naming, could it be related with what they want to do (transfer in quadrants of the 4 source plates)?
This is in the docs of that class:
```ruby
  # Quadrant robots take up to four parent plates which
  # are numbered according to their quadrant.
  # eg. A1 => 1, B1 => 2, A1 => 3, B2 => 4
  # While the pooling robot will work for the majority of cases,
  # it exhibits incorrect behaviour if Well A1 is missing on any
  # of the source wells.
```

config/robots.rb
```diff
@@ -2024,4 +2024,103 @@
...
+                   purpose: 'pWGS-384 AL Lib',
+                   states: ['pending'],
+                   label: 'Carousel 3,5',
+                   target_state: 'started'
+                 }
+               })
+
+  custom_robot('bravo-pwgs-384-end-prep',
+               name: 'Bravo pWGS-384 End Prep',
+               verify_robot: true,
```
emrojo, line 2087: I'm trying to understand how is going to verify the robot. It should be scanning its barcode, but how is it mapped with the name?

config/purposes/pWGS_384_bottom.yml
Chris-Friend: Yes there is a reason this was added. As part of the workflow, the pWGS-384 AL Lib plate needs to be created from the pWGS-384 End Prep plate before that plate has `passe`d. (Both the AL Lib and End Prep plates are `started` in the same robot step, and `passed` in subsequent steps on the same robot). As the comment suggests, including this presenter class allows creating the AL Lib plate from the End Prep plate despite it not being `passed`

config/robots.rb
Chris-Friend: As far as I can tell looking at other pipelines, specifically heron_384, and from how the UI behaves, the `require_robot` and `verify_robot` attributes are required when performing different steps on the same robot. This is something that we require as part of this pipeline.

From what I understand, including the `require_robot: true` attribute registers the first robot, with subsequent `verify_robot: true` attributes indicating that the steps are carried out on this same robot

config/robots.rb
Chris-Friend: I will add this class, and also the `destination_bed` field

config/purposes/pWGS_384_bottom.yml
Chris-Friend: As suggested by the name, this class seems to reduce the amount of information displayed in the UI. Given that this presentation style is used for most other, if not all, PCR plates, I will adopt it for the pWGS-384 pipeline too

config/purposes/pWGS_384_bottom.yml
```diff
@@ -0,0 +1,26 @@
+---
+pWGS-384 Post Shear XP:
+  :asset_type: plate
+  :creator_class: LabwareCreators::QuadrantStamp
+  :merger_plate: true
+  :size: 384
```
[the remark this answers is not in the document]
Chris-Friend: Done

config/robots.rb
KatyTaylor: Yep, there's a box for scanning the robot, and I think if you put 'require_robot: true', the page makes sure you've filled that field in. 
Then, as you say, the verify_robot on a later step checks that plate has been previously verified on the same robot.

config/robots.rb
```diff
@@ -2024,4 +2024,109 @@
                    target_state: 'passed'
                  }
                })
+
+  # Robots for pWGS 384 pipeline
+
+  custom_robot('star-pwgs-384-post-shear-to-pwgs-384-post-shear-xp',
+               name: 'STAR pWGS-384 Post Shear => pWGS-384 Post Shear XP',
```
KatyTaylor, line 2031: I see this matches what it says on the Confluence page, but -
I think these names tend to be <robot> <input plate type> => <output plate type>
So I think it should be 'STAR LB Post Shear => pWGS-384 Post Shear XP' 

Make sense do you think?

Chris-Friend: Fixed

KatyTaylor (approved):

PR #532, opened 2020-10-22
Gpl 674 prevent merger pcr plates into pool with different barcode sets

### LINKED ISSUE #516: GPL-674 Prevent merger PCR plates into pool with different barcode sets
**User story**
GPL-674 | As R&D (Lesley) I would like to prevent the merger of LTHR Lib PCR 1 and LTHR Lib PCR 2 plates into a LTHR Lib PCR pool plate, which are tagged with different barcode sets. This will prevent samples being deplexed incorrectly post sequencing.

Lesley S

**Acceptance criteria**

Token for conversation to have

- [ ] add a list of acceptance criteria here
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Discussed in stand up on 2nd October. This issue impacts on the 96 tailed workflow 'blue light' pipeline. Agreement can roll out to production without it - risk managed via process SOP until GPL-674 released.

KatyTaylor:
Closes #516 

* Add toggle to purposes configuration, enforce_same_template_within_pool?, which can be used for Heron tailed pipeline
* Add various comments and whitespace for understanding and readability


app/assets/javascripts/tag_by_tag_plate.js
```diff
@@ -17,21 +17,29 @@
...
       });
       this.monitor = collector.register(!this.required, this);
+
+      // set initial values for the tag plate data
       this.qcable = unkownQcable;
       this.template = unknownTemplate;
     };
 
+    // why is this called `qcLookup`? Shouldn't it be tagPlateLookup or similar?
     qcLookup.prototype = {
```
JamesGlover, line 43: Because it fetches the 'qcable' rather than the plate itself, this could be a tag tube, tag plate (or potentially also a reporter, although that wouldn't get approved).

app/assets/javascripts/tag_by_tag_plate.js
```diff
@@ -66,17 +76,34 @@
         };
       },
       validators: [
+        // `t` is a qcLookup object
+        // The data for t.template comes from app/models/labware_creators/tagging/tag_collection.rb
```
JamesGlover, line 80: Yes, sorry about this short name. I think it was supposed to be short for 'this'.

app/models/labware_creators/tagging/tag_collection.rb
```diff
@@ -43,10 +46,29 @@ def used
...
+    # @return [Bool] true if either no other templates have been used in the submission pool, or
+    #                     if all the templates used are the same as this one
+    #
+    def matches_templates_in_pool(uuid)
+      # if there haven't been any templates used yet in the pool, we say it matches them
+      return true if used.empty?
+
+      # return true if the list is empty after we remove our uuid, meaning no different templates have been used
+      return used.dup.delete(uuid).empty?
+    end
```
JamesGlover, line 70: Could this just be: `used.include?(uuid)`
Behaviour isn't exactly the same, but the difference only occurs in situations where we've somehow ended up with multiple templates associated the the pool (which wouldn't typically be possible if you follow the critical path, but could happen if you dart out into 'other actions'). With the above, we'd allow the template, as long as it was associated with at least one other plate, whereas in your example we'd reject any templates as soon as we ended up in this state.

app/assets/javascripts/tag_by_tag_plate.js
KatyTaylor: OK, and why is a qcable called a qcable? Is a tag plate something to do with quality control? What's a reporter? :)

app/models/labware_creators/tagging/tag_collection.rb
KatyTaylor: Yes OK, will change

JamesGlover (approved):

app/assets/javascripts/tag_by_tag_plate.js
KatyTaylor: (Reply from Slack):
Yep, so tag plates were introduced along with Gatekeeper, which was developed to add a QC process for tags. We had large batches of tags coming in from a supplier, which were then stamped out into individual tag plates, which underwent an internal QC process in Gatekeeper. These were the QCables, which would then either be released for use in labs, or failed, if they showed cross contamination. Reporters are short sequences of DNA that are used in this QC process, and themselves needed to be QCed.

Gatekeeper was so named because it was literally the gatekeeper of the tags that could go into Sequencescape. ANd QCable was anything that passed through its QC process.

Only… well now IDT mostly QC their own tags, and I’m not sure the QC component of Gatekeeper is even used nay-more

And until very recently we weren’t even stamping tags out any-more… and I don’t think the current stamping process actually uses the stamping pipleine.

Instead all plates come in ready stamped, and ready passed. Making the name Gatekeeper a bit irrelevant, and the QC name a bit confusing.

JamesGlover:
Ooh, realised last night that this is adding a new purposes.yml option. Any chance to could add it to the config added in: https://github.com/sanger/limber/pull/528 (which I've just merged in). Bit cheeky to ask I know, as the code was still waiting for a merge when you submitted the pull request.

KatyTaylor:
Yeah no probs, will add it to the new docs

JamesGlover (approved):

PR #535, opened 2020-10-23
Gpl 679 single child presenter

Chris-Friend:
Add a SingleChildPermissivePresenter workflow: This gives the same behaviours per state as the PermissivePresenter, but with the added restriction that only a single child can be created for the labware (plate).

This is adopted, as per feedback from UAT, in the pWGS-384 pipeline where the End Prep and AL-Lib plates are created, started and passed in intertwined steps

PR directed at existing branch to better see the changes

### DIFF
--- a/app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
+++ b/app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
@@ -4,7 +4,7 @@ module Presenters::Statemachine
   # Supports creation of a single child asset in this state
   module StateAllowsSingleChildCreation
     def control_additional_creation
-      yield if self.child_assets.nil? || self.child_assets.empty?
+      yield if child_assets.blank?
     end
   end
 end
--- a/app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
+++ b/app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module Presenters::Statemachine
+  # Supports creation of a single child asset in this state
+  module StateAllowsSingleChildCreation
+    def control_additional_creation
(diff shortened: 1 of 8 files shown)


app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Presenters::Statemachine
+  # Supports creation of a single child asset in this state
+  module StateAllowsSingleChildCreation
+    def control_additional_creation
+      yield if child_assets.blank?
```
emrojo, line 7: So the idea is that it only shows the button if current plate does not have children plate already created?

app/models/presenters/presenter.rb
```diff
@@ -82,6 +82,10 @@ def inspect
     "<#{self.class.name} labware:#{labware.uuid} ...>"
   end
 
+  def child_assets
+    nil
+  end
```
emrojo, line 87: What was the reason of using ```child_assets``` and the previous alias with ```child_plates```? Did child_plates didn't get the right value for the condition at ```app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb```?

emrojo (approved):
Just a question about child_plates and child_assets. Not important.

app/models/concerns/presenters/statemachine/state_allows_single_child_creation.rb
Chris-Friend: Yep that's the one

app/models/presenters/presenter.rb
Chris-Friend: I included `child_assets` on the `Presenter` concern rather than the `PlatePresenter` class in order that the single child creation workflow can be applied to other types of presenters in the future if needed

emrojo: Yes, good idea 👍

PR #564, opened 2020-11-24
Develop into Master

stevieing:
Closes #

*
*
* ...


app/models/print_job.rb
```diff
@@ -4,45 +4,121 @@ class PrintJob # rubocop:todo Style/Documentation
...
+    #   {
+    #     "main_label"=>{
+    #       "top_left"=>"29-OCT-2020",
+    #       "bottom_left"=>"DN9000214K",
+    #       "top_right"=>"DN9000211H",
+    #       "bottom_right"=>"Duplex-Seq LDS Lig",
+    #       "barcode"=>"DN9000214K-LIG"
+    #     }
+    #   },
+    # ]
```
KatyTaylor, line 71: Would be good to move the above 'labels structure' comment to within the `print_to_pmb` method, as that's where it is used. 

The 'labels_sprint structure' one below is good where it is.

app/views/labware/_barcode_printing_form.html.erb
```diff
@@ -1,26 +1,50 @@
...
+    <!-- extra labels for QC plates in several pipelines -->
+    <% label.qc_attributes.each_with_index do |qc_label, index|%>
+      <% qc_label.each do |k, v|%>
+        <%= f.hidden_field "labels_sprint[qc_#{index}][#{k}]", value: v %>
+      <%end%>
+    <%end%>
+  <% end %>
+
   <%= f.label :printer_name, 'Barcode Printer', class: 'col-2' %>
+  <!-- pass the whole printer object through here, if poss, (instead of :name) or just the id and then re-query the printer through API -->
```
KatyTaylor, line 47: Sorry, we left some of the comments a bit messy. The above one can go - will just be confusing as we've now implemented this

config/sprint/label_templates/plate_384.yml.erb
```diff
@@ -0,0 +1,48 @@
+[
```
KatyTaylor, line 1: Do we want to commit these ymls? The real ones will come from the psd-deployment project right?

PR #580, opened 2020-12-04
Add Additional stamp step at top end of pipeline

JamesGlover:
Addtional stamp step at top end of processed missed in first
requirement:
Yes we will need to do the additional stamp of 10 ul. The instrument
used will be the Bravo (same instrument used for the 96 to 384
consolidation). The source bed will be Bed 1 and the destination bed 5.

### DIFF
--- a/config/robots.rb
+++ b/config/robots.rb
@@ -1988,11 +1988,16 @@ ROBOT_CONFIG = RobotConfiguration::Register.configure do
 
   # Heron LTHR 96 Pipeline
 
+  bravo_robot do
+    from 'LTHR Cherrypick', bed(1)
+    to 'LTHR RT-S', bed(5)
+  end
+
   custom_robot('bravo-lthr-96-rt-to-lthr-96-pcr-1-and-2',
                name: 'bravo LTHR RT => LTHR PCR 1 and 2',
                beds: {
                  bed(9).barcode => {
-                   purpose: ['LTHR RT', 'LTHR Cherrypick'],
+                   purpose: ['LTHR RT', 'LTHR RT-S'],
                    states: ['passed'],
                    label: 'Bed 9'
                  },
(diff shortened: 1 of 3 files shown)


JamesGlover:
Accompanying integration suite: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/20

KatyTaylor (approved):

KatyTaylor:
Is this because the volume cherrypicked in the LTHR cherrypick plate is too much, so this will pick only 10ul of that?

Do you need to add the new plate purpose anywhere, like Limber rake task in SS?

config/robots.rb
```diff
@@ -1988,11 +1988,16 @@
 
   # Heron LTHR 96 Pipeline
 
+  bravo_robot do
+    from 'LTHR Cherrypick', bed(1)
+    to 'LTHR RT-S', bed(5)
+  end
+
   custom_robot('bravo-lthr-96-rt-to-lthr-96-pcr-1-and-2',
                name: 'bravo LTHR RT => LTHR PCR 1 and 2',
```
KatyTaylor, line 1997: Change robot name to RT-S ?

KatyTaylor: Ah, but it caters for both types. I guess it's fine then

JamesGlover:
> Is this because the volume cherrypicked in the LTHR cherrypick plate is too much, so this will pick only 10ul of that?
> 
> Do you need to add the new plate purpose anywhere, like Limber rake task in SS?

Yep, the Cherrypick plate has 35ul, and the RT processing required 10ul. Not need for any changes in SS, limber will automatically register the plate purpose.

The only reason we really need the purposes in the limber rake task is due request types validating that they right purpose has been used, so only the input plates are needed. By adding them to the task, it simplifies the setup process, as it means you don't need to run limber:setup / config:generate in a specific order.

config/robots.rb
JamesGlover: It's a bit frustrating that the different purposes at this stage are for our convenience, not the users.

JamesGlover:
Let Lesley know its in UAT, she's also on leave until Wednesday next week. Said she might get a chance to pick it up today.
- All merged into develop.
- Integration suite run, passing and merged to master.

So releasing this simple case of merging to master, tagging and deploying.

JamesGlover:
Oh, and the various dependency updates are intentional and all looked good during integration tests.

KatyTaylor:
Lesley says "Hi Katy, Yes it was UAT'd last week and can be deployed. I dropped James a mail, but I know he was on leave last week"

PR #599, opened 2021-01-12
Hotfix barcode matching

JamesGlover:
Closes #

*
*
* ...


lib/labware_barcode.rb
```diff
@@ -18,18 +18,24 @@ def prefix
...
-  def sbcf
-    @sbcf ||= SBCF::SangerBarcode.from_human(@human)
-  end
-
   def to_s
     @human
   end
 
-  delegate :=~, to: :sbcf
+  def =~(other)
```
andrewsparkes, line 25: Not following this. Is the regex match operator a method name here?

JamesGlover: Yes, pretty much all operators in ruby are method names. Essentially this comes from allowing DN1234K to match 1220001234757. I've always read =~ more as a match operator, than explicitly to do with reg-exp, although given Rubocop seems to follow a similar line of reasoning to you, I should probably change this.

JamesGlover: Agh, will probably need to updated the SBCF gem as well, due to the use of SBCF::EmptyBarcode, do you mind if I hold off that one for now?

andrewsparkes: Sure. Just finding it hard to understand what the method does. Add some comments to explain basic function?

andrewsparkes: And what is @sbcf? sanger barcode formatted?
the different versions of barcodes are pretty confusing

JamesGlover: Added a bit of explanation

JamesGlover: https://github.com/sanger/sanger_barcode_format

andrewsparkes (approved):

PR #604, opened 2021-01-19
First stab at changing LTHR-384 submission to include all 4 plate bar…

emrojo:
…codes

- include a new config parameter 'allowed_extra_barcodes'
- if this is true, display a text box for scanning extra barcodes
- parse this data and include in submission
by kt17 & emr

Closes #

*
*
* ...


emrojo:
And pull request for gitlab integration suite: Number: 22

app/views/plates/_choose_workflow.html.erb
```diff
@@ -28,7 +29,16 @@
             <% end %>
           <% end %>
         <% end %>
-      <%= submit_tag name, class: 'create-submission-button' %>
+
+        <% if submission.allowed_extra_barcodes %>
+          <%= f.label :extra_barcodes, 'Add additional labware barcodes to the submission (Optional)' %>
+          <br/>
+          <%= f.text_area :extra_barcodes, cols: 30, rows: 5, style: 'width: 100%; resize: none;' %>
```
KatyTaylor, line 36: Thanks, the styling was a bit rubbish before :)

app/javascript/plain-javascript/page-reloader.js
```diff
@@ -3,14 +3,14 @@ document.addEventListener('DOMContentLoaded', () => {
   console.log('Loaded')
   const timerElement = document.querySelector('[data-reload-time]')
   if (timerElement) {
-    let timer = parseInt(timerElement.dataset['reloadTime'])
+    let timer = parseInt(timerElement.dataset['reloadTime'], 10)
```
KatyTaylor, line 6: What was the problem here? Is it making it longer, because it was taking a while to do the submission?

app/models/sequencescape_submission.rb
```diff
@@ -62,7 +64,31 @@ def assets=(asset_uuids)
...
   def assets
-    @asset_groups.pluck(:assets).flatten
+    @asset_groups.pluck(:assets)
+  end
+
+  def extra_barcodes_list
+    return nil unless extra_barcodes
+
+    extra_barcodes.split(' ')
+  end
```
KatyTaylor, line 74: About the split - I'm wondering if we should be a bit more permissive in what we allow the separator to be? e.g. space, comma, new line? Like I think it does in LabWhere? 

Or else, we could put text above the box saying please  separate by one space?

app/javascript/plain-javascript/page-reloader.js
emrojo: No change, it will behave the same. Is a "good practice" change for parseInt(), so it declares the string is a base 10 number.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

app/models/sequencescape_submission.rb
emrojo: Yes, I'll fix that to support commas too.

app/models/sequencescape_submission.rb
```diff
@@ -81,10 +107,16 @@ def asset_groups=(asset_groups)
                     end
   end
 
+  def asset_groups_for_orders_creation
+    return asset_groups unless (asset_groups.length == 1) && extra_barcodes
```
KatyTaylor, line 111: I'm interested to know what's the  scenario where asset_groups length is > 1 ?

app/models/sequencescape_submission.rb
```diff
@@ -62,7 +64,31 @@ def assets=(asset_uuids)
   # @return [Array<String>] Array of asset uuids to submit
   #
   def assets
-    @asset_groups.pluck(:assets).flatten
+    @asset_groups.pluck(:assets)
+  end
```
KatyTaylor, line 68: Could getting rid of the `flatten` be problematic? Do you understand what it was needed for?

app/models/presenters/presenter.rb
```diff
@@ -10,7 +10,7 @@ module Presenters::Presenter # rubocop:todo Style/Documentation
 
     class_attribute :summary_items, :sidebar_partial
 
-    attr_accessor :api, :labware
+    attr_accessor :api, :labware, :additional_labwares
```
KatyTaylor, line 13: Where is `additional_labwares` used?

app/models/sequencescape_submission.rb
KatyTaylor: Ah I see you've got a test that checks it works for new line - didn't realise that would work!

app/models/sequencescape_submission.rb
emrojo: All new assets are added to the same asset group defined, that it should be only 1 when we have extra_barcodes because all assets are added to the same asset group, if we found more than 1 asset group and we have extra_barcodes, then there is something that didnt go well/was not expected

app/models/presenters/presenter.rb
emrojo: Yes, no use; I'll remove it

app/models/sequencescape_submission.rb
emrojo: Yes, I think I removed it by mistake, so rolling it back.

PR #620, opened 2021-02-03
Apply sort order to stock plate retrieval

### LINKED ISSUE #619: BUG - Inconsistent stock plate ids used on lables
**Describe the bug**
Limber usually uses the stock plate barcode as a consistent tracking number through the whole process.
In the case of the LTHR process this is complicated by the pooling upfront.
This is resulting in inconsistencies, especially at the pooling to tubes, in which the stock plate barcode used on the plate differs from that that gets assigned to the tube name.

**RT Ticket Number**
#711295

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

JamesGlover:
Stock plate ordering was not specified, so was dependant on the
behaviour of the sequencescape database. This was resulting in
stock plates appearing to change.

This brings the behaviour mostly in line with Sequencescape, although
uses id over created at. This has the advantage of higher resolution,
quicker speed, and not needing a new sort scope added.

That said, we should probably make the same change within sequencescape.

Fixes #619

Closes #

*
*
* ...

### DIFF
--- a/app/sequencescape/sequencescape/api/v2/plate.rb
+++ b/app/sequencescape/sequencescape/api/v2/plate.rb
@@ -97,7 +97,7 @@ class Sequencescape::Api::V2::Plate < Sequencescape::Api::V2::Base
   end
 
   def stock_plate
-    stock_plates.last
+    stock_plates.order(id: :asc).last
   end
 
   def workline_identifier
--- a/spec/sequencescape/api/v2/plate_spec.rb
+++ b/spec/sequencescape/api/v2/plate_spec.rb
@@ -9,13 +9,17 @@ RSpec.describe Sequencescape::Api::V2::Plate do
   it { is_expected.to_not be_tube }
 
   describe '#stock_plate' do
-    let(:stock_plates) { create_list :v2_plate, 2 }
+    let(:stock_plates) { create_list :v2_stock_plate, 2 }
     let(:plate) do
       build :unmocked_v2_plate, barcode_number: 12_345, stock_plates: stock_plates
     end
+
     before do
-      allow(plate).to receive(:stock_plates).and_return(stock_plates)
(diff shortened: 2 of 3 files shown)


app/sequencescape/sequencescape/api/v2/plate.rb
```diff
@@ -97,7 +97,7 @@ def stock_plates(purpose_names: SearchHelper.stock_plate_names)
   end
 
   def stock_plate
-    stock_plates.last
+    stock_plates.order(id: :asc).last
   end
```
JamesGlover, line 101: The fix

spec/sequencescape/api/v2/plate_spec.rb
```diff
@@ -9,13 +9,17 @@
...
     end
+
     before do
-      allow(plate).to receive(:stock_plates).and_return(stock_plates)
+      ancestor_scope = instance_double('JsonApiClient::Query::Builder')
+      allow(plate).to receive(:stock_plates).and_return(ancestor_scope)
+      expect(ancestor_scope).to receive(:order).with(id: :asc).and_return(stock_plates)
     end
+
     it 'returns the last element of the stock plates list' do
```
JamesGlover, line 23: Feels a bit like I'm over-mocking here, but never worked out the best way of dealing with mocking the JsonApiClient.

emrojo (approved):

spec/sequencescape/api/v2/plate_spec.rb
KatyTaylor: It can be horrible, I agree

KatyTaylor (approved):

PR #626, opened 2021-02-05
GPL-851 Make 'choose workflow' page respond better to scanners

### LINKED ISSUE #622: GPL-851 BUG - 'choose workflow' form handles characters from scanner badly
**Describe the bug**
Came out of RT 711410. From Ben F in Slack:

> There was a 384 plate where the scanner hit return so we only had 3 plates in the selection box to create a 384 plate. We then scanned in the 4th plate hit the 384 button, then combined all 4 plates into 1x384 is this the cause?

**RT Ticket Number**
related to #711410 (where two separate submissions were built, one with 3 plates, and 1 with only one)

**To Reproduce**
Fill this out when you try to reproduce it.

**Expected behaviour**
The form should not auto-submit, regardless of what return characters you get back from the scanner.

**Screenshots**
This is the screen we are talking about:
Screenshot 2021-02-04 at 10 08 57

**Additional context**
From James in Slack:

> Must admit, I thought it was a text-area, and that they didn’t convert a return to a submit. So it might be something slightly different we’re seeing.

KatyTaylor:
Closes #622

### DIFF
--- a/app/assets/javascripts/choose_workflow.js
+++ b/app/assets/javascripts/choose_workflow.js
@@ -4,6 +4,6 @@
   'use strict'
 
   $(function(_event) {
-    diable_enter_key_submit('#choose_workflow_card', '#submission_forms')
+    disable_enter_key_submit('#choose_workflow_card', '#submission_forms')
   })
 })(jQuery, window)
--- a/app/assets/javascripts/choose_workflow.js
+++ b/app/assets/javascripts/choose_workflow.js
@@ -1,28 +1,9 @@
+//= require lib/disable_enter_key_submit.js
+
 (function($, exports, undefined){
   'use strict'
 
   $(function(_event) {
-    // If we are not on the choose_workflow card, don't set anything up.
-    if ($('#choose_workflow_card').length === 0) { return }
-
-    // Build up a list of all fields we may want to tab through.
-    var fields = $('#submission_forms').find("input[type=text],input[type=submit]")
-
-    // Then to each of the text fields, capture the enter event,
-    // and stop it from auto-submitting the form.
-    // This supports the scanners which are configured to send
-    // and enter, rather than a tab. We also set the focus
-    // to the next field, to ensure it behaves the same
-    fields.each(function (index, field) {
-      if($(field).is('input[type=text]')) {
-        $(field).on("keypress", function(e) {
-          var code = e.key;
-          if (code === 'Enter') {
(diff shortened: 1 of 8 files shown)


app/views/plates/_choose_workflow.html.erb
```diff
@@ -1,44 +1,52 @@
...
+              <%= asset_groups_form.fields_for group_id.to_s do |ag_form| %>
+                <% asset_group.except(:assets).each do |key, value| %>
+                  <%# Deconstruct the other fields of our asset group, such as study and project %>
+                  <%= ag_form.hidden_field key, value: value  %>
+                <% end %>
+                <%= ag_form.fields_for :assets do |asset_form| %>
+                <% asset_group[:assets].each do |asset_uuid| %>
+                  <%= asset_form.hidden_field nil, value: asset_uuid  %>
+                <% end %>
+                <% end %>
```
KatyTaylor, line 30: This change is just that I put it inside a div with an id

app/views/plates/_choose_workflow.html.erb
```diff
@@ -1,44 +1,52 @@
...
+          <% if submission.allowed_extra_barcodes %>
+            <%= f.label :extra_barcodes, 'Add additional labware barcodes to the submission (Optional)' %>
+            <br/>
+            <% (1..submission.num_extra_barcodes).each do |num| %>
+              <div class="form-group row">
+                <%= f.label :extra_barcodes, "Barcode #{num}", class:"plate-label col-md-2" %>
+                <%= f.text_field :extra_barcodes, multiple: true, class: "form-control col-md-8" %>
+              </div>
+            <% end %>
+          <% end %>
```
KatyTaylor, line 44: Screenshot 2021-02-05 at 16 10 50

KatyTaylor: Now if you hit tab or enter, it moves onto the next field. Works the same way as the merged_plate page.

app/assets/javascripts/choose_workflow.js
```diff
@@ -0,0 +1,9 @@
+//= require lib/disable_enter_key_submit.js
+
+(function($, exports, undefined){
+  'use strict'
+
+  $(function(_event) {
+    diable_enter_key_submit('#choose_workflow_card', '#submission_forms')
```
emrojo, line 7: disable_enter_key_submit

emrojo (approved):
Ok

PR #632, opened 2021-02-10
Improve error message on quad stamp page

KatyTaylor:
Came out of UAT of GPL-851 'choose workflow' page handles characters from scanner badly.

Improves error message 'duplicated transfers' on quad stamp page, giving more information about the root cause of the problem.

R&D discussed whether the behaviour - where you can't process two active submissions in parallel - was OK, and concluded it was OK for now at least.


JamesGlover (approved):
Much more useful error message. I like the inclusion of the problem plates as well.

PR #636, opened 2021-02-12
Update rucocop

JamesGlover:
- Updated rubocop
- Adds bin/update_rubocop for easy updating
- Adds cron GH action to run every Sunday and automatically create a PR

Tested locally, and looks good.


JamesGlover:
Actually, isn't working as well as I thought, the actual bundle update fails....

.github/workflows/update_rubocop.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Correcting and ignoring
+      run: bundle exec rubocop --auto-correct --disable-uncorrectable
+      # https://github.com/marketplace/actions/create-pull-request
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v3
+      with:
+        title: Automatic Rubocop update
+        branch: gh_action/update_rubocop
+        delete-branch: true
+        body: Automatic update via .github/workflows/update_rubocop.yml
```
JamesGlover, line 38: This is the main bit, everything else is rubocop autofixes and mimics the kind of changes this action will generate pull requests for.

.rubocop.yml
```diff
@@ -6,6 +6,7 @@ AllCops:
     - node_modules/**/*
     - bin/**/*
     - vendor/**/*
+  NewCops: enable
 Rails:
```
JamesGlover, line 10: We enable new cops by default.

Gemfile.lock
```diff
@@ -72,7 +72,7 @@ GEM
     addressable (2.7.0)
```
JamesGlover, line 72: All automatic, this is the gem updates

app/controllers/creation_controller.rb
```diff
@@ -40,7 +40,7 @@ def labware_creator(form_attributes)
   def creation_failed(exception) # rubocop:todo Metrics/AbcSize
     Rails.logger.error("Cannot create child of #{@labware_creator.parent.uuid}")
     Rails.logger.error(exception.message)
```
JamesGlover, line 42: Automatic addition of todos

app/models/limber/tag_layout_template/walk_quadrants.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-module Limber::TagLayoutTemplate::WalkQuadrants # rubocop:todo Style/Documentation
+module Limber::TagLayoutTemplate::WalkQuadrants
   include Limber::TagLayoutTemplate::TagWalker
```
JamesGlover, line 4: And some automatic fixes. Not 100% sure why its decided this doesn't need documentation any more.

.github/workflows/update_rubocop.yml
```diff
@@ -0,0 +1,38 @@
+
+name: Update Rubocop
+
+on:
+  # Can be triggered manually from the GH interface or via a dispatches POST
+  # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+  workflow_dispatch:
+  schedule:
+  # * is a special character in YAML so you have to quote this string
+  # Every sunday at 10 am
```
harrietc52, line 10: How does "0 10 * * 0" specify every sunday at 10 🤔

harrietc52: The link suggests it gets built off the default branch, which is develop, is that okay or should it be of master?

JamesGlover: Welcome to the wonders of cron syntax. I need to use a helper tool online every time I need to write one.
Think I used this on here: https://crontab.guru
Yep, develop is the intent. Means we see the advantages earlier as developers, but the risks hit the users later.

.github/workflows/update_rubocop.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  update_rubocop:
+    if: (github.repository_owner == 'sanger') # On the sanger fork only
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      env:
+        BUNDLE_WITHOUT: 'test development default'
```
harrietc52, line 22:
```suggestion
        BUNDLE_WITHOUT: 'test:development:default'
```
Seems they might need to be separated with colons [here](A :-separated list of groups whose gems bundler should not install.)

harrietc52: What does it bundle with, if not any of them? Just the `lint` group?

.github/workflows/update_rubocop.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        bundler-cache: true
+    - name: Update Rubocop
+      run: |
+        bundle config --local deployment false
+        bundle update --group=lint
+    - name: Correcting and ignoring
+      run: bundle exec rubocop --auto-correct --disable-uncorrectable
```
harrietc52, line 30: Rubocop encourages running tests after an auto-correct [here](https://docs.rubocop.org/rubocop/usage/auto_correct.html), maybe worth doing

.github/workflows/update_rubocop.yml
```diff
@@ -0,0 +1,38 @@
...
+      # https://github.com/marketplace/actions/create-pull-request
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v3
+      with:
+        title: Automatic Rubocop update
```
harrietc52, line 35: There is a default `Changes by create-pull-request action` if you didn't want to include title [here](Changes by create-pull-request action)

.github/workflows/update_rubocop.yml
harrietc52: How does the content of the rubocop update, actually get added to the PR created?

harrietc52 (approved):
Looks good! First problem insight into Github Workflows. All comments are minor and mainly just bits I've spotted from reading some docs. We could use workflow badges too!

.github/workflows/update_rubocop.yml
JamesGlover: Yep, just the lint group. Keeps the install time right down low.
Where did you get that colon separation doc from? Space separation is definitely working (Sequencecape builds would fail completely if it wasn't) but it might be that they are in the process of switching from one to another

.github/workflows/update_rubocop.yml
JamesGlover: The tests will run automatically on the pull request

.github/workflows/update_rubocop.yml
harrietc52: https://bundler.io/v1.16/bundle_config.html search `without (BUNDLE_WITHOUT): A :-`

harrietc52:
My fault. Think I was adding single comments by accident instead of one whole one, all good though 👍

.github/workflows/update_rubocop.yml
JamesGlover: Thanks... I wonder why its working? Will definitely make that change.

.github/workflows/update_rubocop.yml
JamesGlover: By default it essentially just does `git add .` so stages any unignored files that have changed.

When importing it to Sequencescape the .gitignore file didn't ignore the vendor directory, so it generated a pull request with the contents of every single gem.

PR #644, opened 2021-02-23
Migrate tube lookup to V2 api

### LINKED ISSUE #634: BUG - Timeouts in GBS pool tube creation
**Describe the bug**
GBS MiSeq Pool tube creation was returning a 404 after 60s, due to the tube search taking too long to return. I've temporarily increased the timeouts in nginx to relieve the problem. But should do the following:

1) Add a 503 error template to avoid the confusion of a 404
2) Update to use the V2 tube endpoint to improve performance
3) Consider further improvements (such as asynchonous calls)

**RT Ticket Number**
#712129

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

JamesGlover:
The V1 API for tubes is incredibly slow, as it always renders the
full content of the tube. By contrast, V1 allows an utterly minimal
representation.

*Depends on* https://github.com/sanger/sequencescape/pull/3075

Closes #634

*
*
* ...

### DIFF
--- a/app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
+++ b/app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 module LabwareCreators
-  # Pools an entire plate into a single tube. Useful for MiSeqQC
+  # Pools one or more source tubes into a single tube.
+  # Provides an inbox list on the left hand side of the page listing
+  # available tubes (tubes of the correct type).
   class PooledTubesFromWholeTubes < Base
     class SubmissionFailure < StandardError; end
 
--- a/app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
+++ b/app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
@@ -11,7 +11,6 @@ module LabwareCreators
 
     self.page = 'pooled_tubes_from_whole_tubes'
     self.attributes += [{ barcodes: [] }]
-    self.default_transfer_template_name = 'Whole plate to tube'
 
     validate :parents_suitable
 
@@ -39,33 +38,28 @@ module LabwareCreators
 
     # TODO: This should probably be asynchronous
     def available_tubes
-      @search_options = OngoingTube.new(purposes: [parent.purpose.uuid], include_used: false, states: ['passed'])
-      @search_results = tube_search.all(
-        Limber::Tube,
-        @search_options.search_parameters
-      )
+      @search_options = OngoingTube.new(purpose_names: [parent.purpose.name], include_used: false)
+      @search_results = Sequencescape::Api::V2::Tube.find_all(@search_options.v2_search_parameters,
+                                                              includes: 'purpose', paginate: @search_options.v2_pagination)
     end
 
     def parents
-      @parents ||= api.search.find(Settings.searches['Find assets by barcode']).all(Limber::BarcodedAsset,
-                                                                                    barcode: barcodes)
+      @parents ||= Sequencescape::Api::V2::Tube.find_all({ barcode: barcodes }, includes: [])
     end
 
     def parents_suitable
-      missing_barcodes = barcodes - parents.map { |p| p.barcode.ean13 }
+      # Plate#barcode =~ ensures different 'flavours' of the same barcode still match.
+      # Ie. EAN13 encoded versions will match the Code39 encoded versions.
+      missing_barcodes = barcodes.reject { |scanned_bc| parents.any? { |p| p.barcode =~ scanned_bc } }
       errors.add(:barcodes, "could not be found: #{missing_barcodes}") unless missing_barcodes.empty?
     end
 
     private
 
     def transfer_request_attributes
-      parents.each_with_object([]) do |parent, transfer_requests|
-        transfer_requests << { source_asset: parent.uuid, target_asset: @child.uuid }
+      parents.map do |parent|
+        { source_asset: parent.uuid, target_asset: @child.uuid }
       end
-    end
-
-    def tube_search
-      api.search.find(Settings.searches.fetch('Find tubes'))
     end
   end
 end
--- a/app/views/sequencescape/api/v2/tubes/_tube.html.erb
+++ b/app/views/sequencescape/api/v2/tubes/_tube.html.erb
@@ -1,0 +1,10 @@
+<li class="labware">
+  <%= link_to(tube, id: "tube_#{tube.barcode.ean13}", class: "list-group-item-action") do %>
+    <h3 class="barcode list-group-item-heading"><%= useful_barcode(tube.barcode) %> <%= state_badge(tube.state) %></h3>
+    <p><%= tube.name %>
+    <dl>
+      <dt>Created</dt><dd><%= tube.created_at.to_formatted_s(:date_created) %></dd>
+      <dt>Type</dt><dd><%= tube.purpose.name %></dd>
+    </dl>
+  <% end -%>
+</li>
--- a/script/limber
+++ b/script/limber
@@ -16,6 +16,7 @@ if ARGV[0].nil?
 end
 
 environment = ARGV[0]
+command = ARGV[1] || 's'
 
 def build_config # rubocop:todo Metrics/MethodLength
   $stdout.puts 'No configuration for environment'
@@ -53,7 +54,7 @@ begin
   $stdout.puts 'Building configuration'
   $stdout.puts `bundle exec rake config:generate`
   $stdout.puts 'Running application'
-  exec('bundle exec rails s')
+  exec("bundle exec rails #{command}")
 rescue Errno::ENOENT
   attempts = (attempts || 0) + 1
   file = File.open(CONFIG_FILE, 'w')
--- a/app/models/asset_search_form.rb
+++ b/app/models/asset_search_form.rb
@@ -8,7 +8,7 @@ class AssetSearchForm
   include ActiveModel::Model
   # We include 'show_my_plates_only' to let us share pagination forms
   attr_accessor :show_my_plates_only, :include_used, :states, :total_results
-  attr_writer :purposes, :page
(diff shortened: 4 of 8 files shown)


app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -11,7 +11,6 @@ class SubmissionFailure < StandardError; end
 
     self.page = 'pooled_tubes_from_whole_tubes'
     self.attributes += [{ barcodes: [] }]
-    self.default_transfer_template_name = 'Whole plate to tube'
```
JamesGlover, line 14: Not relevant for this creator

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -39,33 +38,28 @@ def barcodes=(input)
...
       errors.add(:barcodes, "could not be found: #{missing_barcodes}") unless missing_barcodes.empty?
     end
 
     private
 
     def transfer_request_attributes
-      parents.each_with_object([]) do |parent, transfer_requests|
-        transfer_requests << { source_asset: parent.uuid, target_asset: @child.uuid }
+      parents.map do |parent|
+        { source_asset: parent.uuid, target_asset: @child.uuid }
```
JamesGlover, line 61: No idea why this was each with object.

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -39,33 +38,28 @@ def barcodes=(input)
...
-                                                                                    barcode: barcodes)
+      @parents ||= Sequencescape::Api::V2::Tube.find_all({ barcode: barcodes }, includes: [])
     end
 
     def parents_suitable
-      missing_barcodes = barcodes - parents.map { |p| p.barcode.ean13 }
+      # Plate#barcode =~ ensures different 'flavours' of the same barcode still match.
+      # Ie. EAN13 encoded versions will match the Code39 encoded versions.
+      missing_barcodes = barcodes.reject { |scanned_bc| parents.any? { |p| p.barcode =~ scanned_bc } }
       errors.add(:barcodes, "could not be found: #{missing_barcodes}") unless missing_barcodes.empty?
```
JamesGlover, line 54: In practice tubes are currently ean13, but this kind of code bit us when we updated plates to code39.

script/limber
```diff
@@ -53,7 +54,7 @@ begin
   $stdout.puts 'Building configuration'
   $stdout.puts `bundle exec rake config:generate`
   $stdout.puts 'Running application'
-  exec('bundle exec rails s')
+  exec("bundle exec rails #{command}")
 rescue Errno::ENOENT
```
JamesGlover, line 58: `./script/limber` lets you preconfigure limber to talk to different environments with `./script/limber ENVIRONMENT` this lets you also start up a rails console, via `./script/limber ENVIRONMENT c`

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
KatyTaylor: Just noticed the comment at the top of this file says `Pools an entire plate into a single tube. Useful for MiSeqQC`
Presume should say `Pools an entire tube into a single tube. Useful for MiSeqQC`

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -39,33 +38,28 @@ def barcodes=(input)
 
     # TODO: This should probably be asynchronous
     def available_tubes
```
KatyTaylor, line 40: Does this method find tubes that are appropriate to be the source tubes? (correct purpose)

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -39,33 +38,28 @@ def barcodes=(input)
...
-      @search_options = OngoingTube.new(purposes: [parent.purpose.uuid], include_used: false, states: ['passed'])
-      @search_results = tube_search.all(
-        Limber::Tube,
-        @search_options.search_parameters
-      )
+      @search_options = OngoingTube.new(purpose_names: [parent.purpose.name], include_used: false)
```
KatyTaylor, line 41: Why remove the `states: ['passed']` part?

JamesGlover: The existing search functionality was actually ignoring the filter anyway, but it occurred to me that its probably quite useful to see tubes that are almost ready as well. A pending or started tube in this list is only one step away from being available.

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
JamesGlover: Yep, it tries to exclude those that have been used, but there's another bug elsewhere in Limber that means it isn't always 100% successful in this. (I've added a story here as it seemed to be missing https://github.com/sanger/limber/issues/645)

I also noticed that there seem to be a few tubes 'abandoned' at this stage. Some of them were completed on another fork, others looks just genuinely abandoned. The former we could probably add a scope to filter out, but the latter will always hand about.

There's a limit to confine the returned tubes to the 30 most recent though, so this list shouldn't get too unmanageable.

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
```diff
@@ -39,33 +38,28 @@ def barcodes=(input)
...
-      @parents ||= api.search.find(Settings.searches['Find assets by barcode']).all(Limber::BarcodedAsset,
-                                                                                    barcode: barcodes)
+      @parents ||= Sequencescape::Api::V2::Tube.find_all({ barcode: barcodes }, includes: [])
     end
 
     def parents_suitable
-      missing_barcodes = barcodes - parents.map { |p| p.barcode.ean13 }
+      # Plate#barcode =~ ensures different 'flavours' of the same barcode still match.
+      # Ie. EAN13 encoded versions will match the Code39 encoded versions.
+      missing_barcodes = barcodes.reject { |scanned_bc| parents.any? { |p| p.barcode =~ scanned_bc } }
```
KatyTaylor, line 53: Bit confused by this line. Looks to me like `=~` checks if the string one side is contained within the other, and returns either the offset (integer) or nil. But I would have thought `.any?` wants a boolean returned..?

app/views/sequencescape/api/v2/tubes/_tube.html.erb
```diff
@@ -0,0 +1,10 @@
+<li class="labware">
+  <%= link_to(tube, id: "tube_#{tube.barcode.ean13}", class: "list-group-item-action") do %>
+    <h3 class="barcode list-group-item-heading"><%= useful_barcode(tube.barcode) %> <%= state_badge(tube.state) %></h3>
+    <p><%= tube.name %>
+    <dl>
+      <dt>Created</dt><dd><%= tube.created_at.to_formatted_s(:date_created) %></dd>
+      <dt>Type</dt><dd><%= tube.purpose.name %></dd>
+    </dl>
+  <% end -%>
+</li>
```
KatyTaylor, line 10: where will this be displayed?

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
JamesGlover: Actually that line should be completely re-written, its left over from another creator that I used to template this one.

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
JamesGlover: Barcode in this case isn't a string, but a LabwareBarcode object. https://github.com/sanger/limber/blob/87cb3fbb8ceb0ba4f102b8dc3831ee23814f36f8/lib/labware_barcode.rb

KatyTaylor: ah it's got a special thingamybob

KatyTaylor (approved):

app/views/sequencescape/api/v2/tubes/_tube.html.erb
JamesGlover: `app/views/tube_creation/pooled_tubes_from_whole_tubes.html.erb` renders the partial:
`app/views/search/_inbox.html.erb`
Which contains
`render sorted_labware`
When render receives an array, rails treats it as a collection, and selects the template to used based on the class name. (More accurately various helper methods provided by ActiveRecord/ActiveModel)
In this case we were using the default template for the V1 API stuff, and this just sets up a similar template for the V2 stuff, allowing rails to switch over automatically.

JamesGlover: I've not removed the V1 templates yet, as we'll need to migrate over the inboxes first. (Which we should definitely do)

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
JamesGlover: Updated the comment to remove references to plates (And MiSeq QC)

app/models/labware_creators/pooled_tubes_from_whole_tubes.rb
JamesGlover: Wish I hadn't used the standard match operator, as it caused confusion to Andrew as well. I'll probably rename that at somepoint, or at least add an alias.

app/views/sequencescape/api/v2/tubes/_tube.html.erb
KatyTaylor: Cool OK thanks
Do you think it's worth a comment in this file with an abridged version of that? Just thinking if someone comes across it while on a deleting spree it's not going to be searchable where it's used

Issue #606, opened 2021-01-19
GPL-823 Show help text when a plate is cancelled


KatyTaylor:
**User story**
As a Limber user, I would like to have some help text on the page when a plate is cancelled, to help me understand why.

**Who are the primary contacts for this story**
Katy, James G

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] When on a page where the UI shows that the plate is cancelled, due to the associated transfer request and submission being cancelled, some help text shows that says something like "This plate has been cancelled, along with the associated submission. If you think it's been replaced by a new submission, try going back to the stock plate (link) to start again on this pipeline."

**Dependencies**
This story is blocked by the following dependencies:
- https://github.com/sanger/sequencescape/issues/3019 

**Additional context**
See related (blocking) SS story in dependencies for background.

The screen we're talking about will look something like this currently:
![Screenshot 2021-01-19 at 16 21 26](https://user-images.githubusercontent.com/10407809/105065500-d4a21e80-5a75-11eb-9266-4f04929e9e10.png)

JamesGlover:
Validators on presenters already hook into the warning / message system, so would be the ideal way of handling this.

PR #672, opened 2021-03-17
Issue648/gpl 900 cancel a single plate {heron}

### LINKED ISSUE #608: BUG - Failing wells on a LB Lib PCR-XP post pooling results in a 500
**Describe the bug**
If the user forgets to fail a well before pooling, then they should be able to fail it retrospectively. However attempting to do so on the LB Lib PCR-XP results in a 500 error. The user is able to work around this by failing the well on the parent plate instead.

**RT Ticket Number**
If applicable

**To Reproduce**
1. Vist training limber
2. Scan in the barcode DN709597I
3. Click Fail Wells
4. Select a well
5. Click 'Fail Selected Wells'
6. You will see a failure
The exception generated is:

> There is an issue with the API connection to Sequencescape ({"general":["Event 'fail' cannot transition from 'failed'."]})

This is probably a case of the library creation request being failed twice, chances are the state change is failing the request directly, and then again when attempting to find the outer request associated with the transfer request. 

**Expected behaviour**

It is expected that the action will complete without error, and the well will appear red. Behind the scenes we also expect to see:
1. Failure of the transfer request into the well
2. Transition of the library creation request from passed to failed
3. Ideally a cancelation of the multiplexing request (Unless the library creation request is multiplexed)

**Other Criteria**
1. Well failure on other plate should still fail both the transfer request and the library creation request, regardless if they are failed when the request is in progress, or retrospectively.
2. Retrospective failures walk down the request graph, removing matching aliquots. This behaviour is intentional and allows failure decisions to be made retrospectively.
3. Step 2 will not remove aliquots from QC tubes. Again, this is intentional. This is as the QC tube will inform the failure decisions, and *Will* contain the failed samples. (Albeit at levels not sufficient for QC)

**Screenshots**

* Windows or Mac?

**Additional context**

### LINKED ISSUE #614: BUG - LTHR Cherrypick plates should allow submission creation when canceled
**Describe the bug**
If a submission on an LTHR Cherrypick plate is canceled the plate displays in limber as 'canceled'. However there is no ability to create a new submission. The submission creation behaviour should also be enable for cancelled plates.  

**To Reproduce**
1. Create a submission of an LTHR Cherrypick plate
2. Cancel the submission in limber
3. Refresh the view of the LTHR Cherrypick plate

**Expected behaviour**
1. The plate should display the submission creation sidebar

**Screenshots**

* Windows or Mac?

**Additional context**

### LINKED ISSUE #630: BUG - Cancelling tubes is resulting in a 500
**Describe the bug**
Following tube cancellation we're seeing
```
An ActionView::Template::Error occurred in tubes#show:

  State :cancelled for :state is not a valid context for calling #control_suggested_library_passing
  app/views/tubes/sidebars/_default.html.erb:13:in `block (2 levels) in _app_views_tubes_sidebars__default_html_erb__2933959237873345365_69906776703400'
```

This may be two problems:
1) Cancelling tubes shouldn't cancel the multiplexing requests
2) The above exception shouldn't occur, regardless of whether point 1 is true or not. (Ie. We shouldn't cancel the multiplexing requests, but shouldn't blow up like this if they are cancelled)

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

### LINKED ISSUE #648: GPL-900 Cancel a single plate [Heron]
**User story**
As a lab user I would like to be able to cancel a single plate, without any further repercussions (no requests cancelled etc), to signal that it should not be processed further. It should behave consistently regardless of which plate it is. This will be helpful as a visual indicator that this plate should not be processed further, but allow me to go back to an earlier plate in the pipeline and carry on from there under the same submission.

James G, Katy

**Acceptance criteria**
When I cancel a plate in Limber:
- [ ] It cancels the transfer requests into that plate
- [ ] It does not cancel any Requests or Submissions
- [ ] It does not block processing on any plates other than the one I am cancelling
- [ ] There is some signposting in the app telling me what will happen when I cancel the plate

JamesGlover:
Closes #648 
Closes #630 
Closes #614 
Closes #608

Paritally addresses #606

* Update cancel interface to better support #648
* Update failed/cancel submission reporting
* ...


app/assets/javascripts/state_change_reasons.js
```diff
@@ -1,23 +1,15 @@
...
-  };
-
-  $(document).ready(function(){
-    $('#state-changer').find('#state').on('change', displayReason);
-  });
-
+  var updateDisplay = function(new_state, speed) {
+    $('.reason').not('#' + new_state + '_reasons').slideUp(speed)
+    $('#' + new_state + '_reasons').slideDown(speed)
+  }
```
JamesGlover, line 9: Simplified by removing need to disable form-fields. Also reduces the risk of accidentally submitting the wrong information.

app/helpers/plate_helper.rb
```diff
@@ -56,20 +59,4 @@ def sorted_pre_cap_pool_json(current_plate) # rubocop:todo Metrics/AbcSize
...
-
-    {}.tap do |h|
-      pools.each do |key, value|
-        value['wells'].each do |well|
-          h[well] = pools_by_position[key]
-        end
-      end
-    end
-  end
 end
```
JamesGlover, line 62: No longer used

app/helpers/plate_helper.rb
```diff
@@ -1,10 +1,13 @@
...
+  # Proxy object wrapping the form alongside the presenter.
+  # This allows us to use the shared plate partial, but pass the form
+  # object through to the custom aliquot partial
+  class WellFailingPresenter < BasicObject
     def initialize(form, presenter)
-      @form = form # rubocop:todo Rails/HelperInstanceVariable
-      @_presenter = presenter # rubocop:todo Rails/HelperInstanceVariable
+      @form = form # rubocop:disable Rails/HelperInstanceVariable
+      @_presenter = presenter # rubocop:disable Rails/HelperInstanceVariable
     end
```
JamesGlover, line 11: This is a false positive, as this isn't actually an instance variable in the helper itself.

andrewsparkes (approved):

PR #706, opened 2021-04-16
Add new library type Haplotagging

emrojo:
Closes GPL921

* Add new library type Haplotagging for Limber Bespoke - PCR

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,28 @@
 
 A flexible front end to plate bases pipelines in Sequencescape
 
+## Initial Setup
+
+1. In Sequencescape, configure all Limber required data:
+```shell
+  bundle exec rake limber:setup
+```
+
+1. Start Sequencescape (will start in 3000):
+```shell
+  bundle exec rails s
+```
+
+1. In Limber, connect to Sequencescape to configure required data:
+```shell
+  bundle exec rake config:generate
+```
+
+1. Start Limber (will start in 3001):
+```shell
+  bundle exec rails s
+```
+
 ## Docs
 
 In addition to the [externally hosted YARD docs](https://www.rubydoc.info/github/sanger/limber), you can also run a local server:
--- a/README.md
+++ b/README.md
@@ -11,25 +11,32 @@ A flexible front end to plate bases pipelines in Sequencescape
 
 ## Initial Setup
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,6 +9,28 @@
 
 A flexible front end to plate bases pipelines in Sequencescape
 
+## Initial Setup
+
+1. In Sequencescape, configure all Limber required data:
```
JamesGlover, line 14: It's actually marginally safer to run this after `bundle exec rake config:generate` as it allows limber to register its own plate purposes, which is what actually happens in production. 

I've been hoping to uncouple these steps (It comes from request types validating their plate purposes as the SSRs kept using the wrong plate purpose) but haven't had the chance. (Plan would be to store the plate purpose via name on the submission template, and then have the order perform the validation instead. This would have the added benefit of capturing the mistake earlier.)

README.md
```diff
@@ -9,6 +9,28 @@
...
+```shell
+  bundle exec rake limber:setup
+```
+
+1. Start Sequencescape (will start in 3000):
+```shell
+  bundle exec rails s
+```
```
JamesGlover, line 22: If it worth mentioning the delayed job as well?

README.md
```diff
@@ -9,6 +9,28 @@
...
+1. In Limber, connect to Sequencescape to configure required data:
+```shell
+  bundle exec rake config:generate
+```
+
+1. Start Limber (will start in 3001):
+```shell
+  bundle exec rails s
+```
+
```
JamesGlover, line 33: Are all the steps supposed to be numbered 1? I've seen Phill do something similar, so I'm assuming there's some kind of reasoning behind it.

JamesGlover (approved):
Popped a few suggestions on the setup, only the first of which is important. (It avoids odd differences between development and production).

README.md
emrojo: Yes, I'll add it

README.md
emrojo: It has like an auto-increment, however there are times where it doesnt work well. Just in case, I'll change it to the real numbering.

README.md
emrojo: Ok, I'll reorder it

Issue #710, opened 2021-04-22
GPL-988 investigate queries required for pipeline kanban


KatyTaylor:
**Description**
Look into what queries (API calls) would be needed for a kanban style display of a pipeline within Limber.

**Who the primary contacts are for this work**
Katy, James G, Tom W

**Additional context or information**
This is part of the wider requirements for having an overview of what plates are where / reporting / replacing the Heron Trello board. (see https://github.com/sanger/General-Backlog-Items/issues/139)

TWJW-SANGER:
Link to [UKB - Heron WIP meeting notes and agreements](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/UKB+-+Heron+WIP+Reporting+Meeting)

KatyTaylor:
Used these options:
- Sequencescape V2 API (JSON API)
- join to state_changes (to avoid using transfer requests to determine the state of the plate / tube)
- use '[sparse fieldsets](https://github.com/JsonApiClient/json_api_client#sparse-fieldsets)' to avoid unnecessary queries retrieving fields we don't want
- use maximum page size Sequencescape allows (500) as it reduces the number of API requests we have to make, and massively reduces loading time

This gives a loading time of 30 seconds for a proof of concept page, which lists out all plates and tubes of the LTHR-384 Heron pipeline purposes, grouped by purpose, displaying their barcode and state (542 records at time of writing).

This required modifications as seen in these commits:
- Sequencescape:
    - https://github.com/sanger/sequencescape/commit/ffd4a42e7cd223b235384474f4d8428624e7b294
- Limber:
    - https://github.com/sanger/limber/commit/258034dcebc0d10440382c716d0b97da607f9494
    - https://github.com/sanger/limber/commit/dd656c0235d91187bc46632e56dd6bba0bda41bd

KatyTaylor:
Putting it in code review in case anyone wants to have an early look at the code linked to in previous message. 
It will be incorporated into the code base as part of the actual kanban story - https://github.com/sanger/General-Backlog-Items/issues/139

KatyTaylor:
Closing as is in 'Done' column

PR #726, opened 2021-05-13
Gpl 988 investigate queries for kanban

KatyTaylor:
First iteration of kanban / swimlane view of work in progress, for Heron lab staff / managers.
[GPL-988](https://github.com/sanger/limber/issues/710) and [GPL-953](https://github.com/sanger/General-Backlog-Items/issues/139) 

Linked PR in Sequencescape - https://github.com/sanger/sequencescape/pull/3188


app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+  #   "LTHR Cherrypick" => [
+  #     {
+  #       :record => #<Sequencescape::Api::V2::Labware:@attributes={"type"=>"plates", "id"=>"1234", "uuid"=>"12a34b5c-defg-67hi-8jkl-mn901opq2345", "labware_barcode"=>#<LabwareBarcode:0x00007ff16e5f06b0 @human="DN123456D", @machine="DN123456D", @ean13="1234567890123">, "created_at"=>2021-05-04 12:19:32 +0100}>,
+  #       :state => "pending"
+  #     }
+  #   ],
+  #   "LTHR-384 PCR 1" => [{}]
+  # }
+  def mould_data_for_view(purposes, labware_records)
+    output = {}
```
stevieing, line 125: You can use tap here
`{}.tap do |output|
...
end`

It returns the hash.

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+  def index
+    # TODO: how can we avoid hardcoding this? Problem is there are multiple relevant pipelines we want to display in one.
+    @pipeline = '"Heron"'
+
+    # TODO: test including the 'Heron 384 Tailed MX' pipeline - might cause an issue as there might be loads of tubes in the final purpose
+    pipeline_configs = Settings.pipelines.select{ |pipeline| ['Heron-384 Tailed A', 'Heron-384 Tailed B'].include? pipeline.name }
+    @ordered_purpose_list = combine_and_order_pipelines(pipeline_configs)
+
+    page_size = 500
+    param_date = params[:date]&.to_date
```
stevieing, line 13: I think you can just do
`from_date = params[:date]&.to_date || Date.today.prev_month`

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+class PipelineWorkInProgressController < ApplicationController
+  def index
+    # TODO: how can we avoid hardcoding this? Problem is there are multiple relevant pipelines we want to display in one.
```
stevieing, line 5: Not sure what "Problem is there are multiple relevant pipelines we want to display in one." means

KatyTaylor: It means that ideally, I'd like to just give them a selection of pipelines to choose from, and display the plates for whichever one they choose (for in the future, when it's not limited to Heron).

There's a problem with that though, since for 'Heron' I want the page to display 'Heron-384 Tailed A' and 'Heron-384 Tailed B' - so, for one pipeline there might be multiple pipeline configs, and there's nothing to help the code figure out those groupings. I think it's just a case of adding something into the pipeline config like 'pipeline_group' or something, which this code could read. 

Not a priority for now though, I think hardcoding Heron is OK for first version. I'll make the comment clearer.

app/controllers/pipeline_work_in_progress_controller.rb
KatyTaylor: yes that's neater, thanks

app/controllers/pipeline_work_in_progress_controller.rb
KatyTaylor: looks good, thanks

app/controllers/pipeline_work_in_progress_controller.rb
stevieing: Adding some config sounds like it would be the best way to do it.

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+class PipelineWorkInProgressController < ApplicationController
```
stevieing, line 3: This is just a personal choice but I would generally opt for a service object to do this kind of processing. For various reasons:
- it keeps the controller actions separate from any data processing which makes it easier to debug i.e. if your controller tests fail you will know they are failing because of the request or response.
- it makes the whole thing a bit cleaner and gives you better separation of concerns.
- it makes things easier to test i.e. you can test your controller actions separately from your processing actions.

This is a strategy that I generally use as shown here https://github.com/sanger/traction-service/blob/f36471dc4cfc5e9ec5dcb2bce84c27d1a38079be/app/controllers/v1/pacbio/plates_controller.rb. But again it is personal choice. There is a blog post about it here https://thoughtbot.com/blog/skinny-controllers-skinny-models and this is the approach I generally take.

KatyTaylor: I moved the biggest method into PipelineList as it is generic and seems to belong OK there. I think the others are probably OK here for now.

JamesGlover (changes requested):
Popped a few suggestions on there.

stevieing (approved):

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+  #   ],
+  #   "LTHR-384 PCR 1" => [{}]
+  # }
+  def mould_data_for_view(purposes, labware_records)
+    {}.tap do |output|
+      # Make sure there's an entry for each of the purposes, even if no records
+      purposes.each { |p| output[p] = [] }
+
+      labware_records.each do |rec|
+        next unless rec.purpose
```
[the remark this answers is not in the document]
KatyTaylor: They're not, just being defensive...

app/views/pipeline_work_in_progress/index.html.erb
```diff
@@ -0,0 +1,32 @@
+<%= content do %>
```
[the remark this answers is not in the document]
KatyTaylor: Oops...

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    labware_query = Sequencescape::Api::V2::Labware
+                    .select({ plates: %w[uuid purpose labware_barcode state_changes created_at ancestors] },
+                            { tubes: %w[uuid purpose labware_barcode state_changes created_at ancestors] },
+                            { purposes: 'name' })
+                    .includes(:state_changes, :purpose)
+                    .where(without_children: true, purpose_name: purposes, created_at_gt: from_date)
+                    .order(:created_at)
+                    .per(page_size)
+
+    labware_query = labware_query.includes(:ancestors) unless reduce_information_for_performance(from_date)
```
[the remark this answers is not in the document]
KatyTaylor: Ah I don't think I knew you could do that, thank you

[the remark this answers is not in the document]
KatyTaylor: Really effective - I've got rid of the 'reduced information' thing as it isn't ridiculously slow any more

config/routes.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
+  get 'pipeline_work_in_progress/index'
```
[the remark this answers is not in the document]
KatyTaylor: This is nice, thanks

app/controllers/pipeline_work_in_progress_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    num_retrieved = page_size
+    while num_retrieved == page_size
+      current_page = query_builder.page(page_num).to_a
+      num_retrieved = current_page.size
+      all_records += current_page
+      page_num += 1
+    end
+
+    all_records
+  end
```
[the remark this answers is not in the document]
KatyTaylor: Yep makes sense, will extract it. 

And yes I just tested this, fetching beyond the range returns an empty array, so num_retrieved gets set to 0 and the loop will terminate.

KatyTaylor:
@JamesGlover Thanks for the comments, they were really helpful. I think I've addressed them all. Could you have another look please? Then I'll probably deploy to training so that the users can have a go with it with realistic data.

app/sequencescape/sequencescape/api/v2.rb
```diff
@@ -36,4 +36,23 @@ def self.plate_for_completion(uuid)
...
+    # it does one more iteration and you get an empty array retrieved, stopping the loop
+    while num_retrieved == page_size
+      current_page = query_builder.page(page_num).to_a
+      num_retrieved = current_page.size
+      all_records += current_page
+      page_num += 1
+    end
+
+    all_records
+  end
```
JamesGlover, line 57: Cool. This should let use switch to API::V2 for loading purposes via rake config:generate. A bit of preliminary work, suggests it will drastically speed things up. (API V1 takes plate counts for each purpose, which, even with indexes, is quite slow)

JamesGlover (approved):
Looks good here

PR #747, opened 2021-06-21
DPL-017 Surge testing bed verifications for LTHR pooling stage - Limber components

andrewsparkes:
Changes to allow bed verifications of plates into tubes, plus new robot configuration for 384 well pipeline.

NB. Includes DPL-035 and 036 bed verification robots.


app/controllers/exports_controller.rb
```diff
@@ -93,6 +93,6 @@ def locate_ancestor(plate_id)
   end
 
   def include_parameters
-    csv_details.plate_includes || 'wells'
+    csv_details.labware_includes || 'wells'
```
JamesGlover, line 96: These should probably still be plate includes, as they get used as part of: `Sequencescape::Api::V2.plate_with_custom_includes` so don't currently support tubes. (Although in turn this is wrapped in locate_labware so the code isn't exactly honest about its limitations as is)

app/models/robots/bed/pooling.rb
```diff
@@ -7,23 +7,45 @@ class Pooling < Robots::Bed::Base
...
-    def parent_plates
-      @parent_plates ||= plate.wells.sort_by(&well_order).each_with_object([]) do |well, plates|
+    def parent_labwares
+      return [] if labware.nil?
+
+      @parent_labwares ||= if labware.plate?
+                             parent_labwares_of_plate
+                           else
+                             # currently not used for tubes
+                             []
```
JamesGlover, line 35: Isn't this needed for tubes? How are you validating that a tube is paired with the correct plate?

app/models/robots/bed/splitting.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Robots::Bed
+  # Splitting robots are specific to plates
+  class Splitting < Robots::Bed::Base
+    attr_accessor :parents
+
+    def load(barcodes)
```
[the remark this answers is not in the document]
JamesGlover: Ahh, we already have a separate splitting bed class. Excellent. In which case, definitely go for option 2 and move those methods I mentioned.

spec/models/labware_metadata_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    has_a_working_api
+
+    it 'raises an exception if the barcode is invalid' do
+      stub_asset_search(456, nil)
+      expect do
+        LabwareMetadata.new(api: api, barcode: 456, user: user_uuid)
+      end.to raise_error(Sequencescape::Api::ResourceNotFound)
+    end
+
+    context 'plates' do
```
JamesGlover, line 28: There's a lot of duplication in these specs. Might be cleaner to set-up the two contexts and use some shared examples.

JamesGlover (changes requested):

JamesGlover (approved):

Issue #774, opened 2021-07-14
DPL-nnn replace Artic V3 primer panel with a newer design: Artic V4


neilsycamore:
Main person: Naomi P
Required??

All details copied from RT https://rt.sanger.ac.uk/Ticket/Display.html?id=725579

Hi
Putting in a ticket to request work enabling use of V4 Artic primers for Heron
Scott T mentioned this on a PSD call and suggested that PSD would need to adjust a limber config file to enable.
I'm not sure how much work this is, and how early it needs to commence, or if there are any further requirements PSD side to switch primer panels in heron?

NPG have completed their work, and I have put the new primer panel into sequencescape as nCoV-2019/V4

Thanks
Naomi

_____________________________________________
From: Naomi Park
Sent: 06 July 2021 16:12
To: David Jackson <dj3@sanger.ac.uk>; Rich Livett <rl15@sanger.ac.uk>; Scott Goodwin <sg21@sanger.ac.uk>
Cc: Laura Grout <lg20@sanger.ac.uk>; Sara Stott <ss50@sanger.ac.uk>; Neha Wali <nw6@sanger.ac.uk>; James Mack <jm36@sanger.ac.uk>
Subject: Artic V4 Heron primers

Hi

As you may be aware, the current Artic V3 primer panel in use (albeit a rebalanced V3/B) is to be replaced with a new design, Artic V4
FYI information surrounding reasons and its design here: SARS-CoV-2 version 4 scheme release - Laboratory - ARTIC Real-time Genomic Surveillance<https://community.artic.network/t/sars-cov-2-version-4-scheme-release/312>

I have initiated a change control doc and identified key stages:
https://drive.google.com/file/d/1iWdyQFRVd46PAZ4p4sTo9-AKjdsm7q29/view?usp=sharing

I'm hoping the oligos will arrive within the next 1-2 weeks, after which the pressing need is for R&D to validate for transition to operations before running out of V3 stocks in early Sept
More pressing is to generate reliable coverage of amplicon #72 region in delta so the requirement is to roll this out when possible.
Therefore this work is of a high priority

Happy to call a meeting to discuss, but as a kick off I've identified the following actions and put names to it (James and Sara you have no actions but it's an FYI)

Whilst mostly sitting in R&D -  NPG and PSD have some key actions
@David and @Rich have I missed anything and should I be creating tickets for the highlighted below?

Thanks
Naomi

Key Stage       Task    Owner
[...]

PR #785, opened 2021-08-02
Fix up README

sdjmchattie:
There were quite a few missing steps in setting up.  I've added these in now.


README.md
```diff
@@ -7,36 +7,72 @@
...
 
-2. In Sequencescape, configure all Limber required data:
-```shell
-  bundle exec rake limber:setup
-```
+1. In Limber, ensure the appropriate version of Ruby is installed. The command
+   here is for `rbenv` but you may want to use a different Ruby version manager.
+   If the command fails, you might need to check the list of available Ruby
+   versions using `rbenv install --list-all` and use the one that most closely
+   matches the contents of `.ruby-version`:
```
JamesGlover, line 21: I'm not sure we want to encourage developers to use different ruby versions.

README.md
```diff
@@ -7,36 +7,72 @@
...
+   ```shell
+   yarn install
+   ```
+
+4. In Sequencescape, configure all Limber required data:
+
+   ```shell
+   bundle exec rake limber:setup
+   ```
+
```
JamesGlover, line 44: The previous instructions were advising running config:generate before limber:setup as it better ensures that limber remains the authority on plate purposes. The alternative approach does work, but *should* be showing warnings. (I suspect they are buried in the rest of the logs)

README.md
```diff
@@ -7,36 +7,72 @@
...
+5. In SequenceScape, setup an API key for Limber via the Rails console:
+
+   ```shell
+   bundle exec rails console
+   new_app = ApiApplication.new(name: 'limber', contact: 'developers', key: 'development', privilege: 'full')
+   new_app.save(validate: false)
```
JamesGlover, line 50: This shouldn't be necessary, this should all be set-up automatically, either when running `bin/setup` in sequencescape, or individually as part of `rake db:seed` task (run on both `rake db:setup` and `rake db:create`)

README.md
JamesGlover: Background: Annoyingly we need to have request types validate the plate purpose of their source wells, as otherwise we get a constant stream of user errors. Currently this is via an association on request type (acceptable plate purposes) which means request type setup is dependent to the plate purposes.

I *really* want to decouple these, but haven't had the chance.

JamesGlover (changes requested):

README.md
sdjmchattie: Sorry maybe this reads wrong.  When I tried to run `rbenv install` it told me that there was no Ruby version for `ruby-2.7.2`, which it picks up automatically from the `.ruby-version` file.  So I ran `rbenv install --list-all` and found `2.7.2` without the `ruby-` prefix.  After that installed with `rbenv install 2.7.2`, `rbenv` referred to it as `ruby-2.7.2` so either the `.ruby-version` file will need to be changed, or these instructions need to be here to explain what to do when `rbenv` cannot find `ruby-2.7.2`.

README.md
sdjmchattie: I'll move this step down in the sequence.  I wasn't sure whether we'd prefer this to be done before SS gets fired up, but it sounds like that can't be avoided.

README.md
JamesGlover: Ahh I see... that makes sense. I've been updating most our apps to say just 2.7.2 in `.ruby-version` for precisely this reason, bit surprised that limber hasn't been updated yet.

JamesGlover: I do find it weird that rbenv is more permissive when switching ruby versions than installing them. I know installing is provided by a plugin, but would still expect them to behave the same.

README.md
JamesGlover: Yeah, its pain enough that the 'wrong' way will still work, but I figured the instructions should encourage the 'correct' approach.

README.md
sdjmchattie: I had a lot of trouble running `bin/setup` in SequenceScape and I couldn't get it to run to completion.  However, I've dropped all the tables from MySQL again and now it worked and like you said the key was created for me.

README.md
sdjmchattie: I've updated the `.ruby-version` file now and removed these additional confusing instructions.

sdjmchattie:
Out of interest @JamesGlover why is there a JavaScript test marked as `xit` to make it pending?  It also doesn't pass if you take the `x` away.  It's highlighted in the bottom of the Diff on this PR.

JamesGlover:
> Out of interest @JamesGlover why is there a JavaScript test marked as `xit` to make it pending? It also doesn't pass if you take the `x` away. It's highlighted in the bottom of the Diff on this PR.

I'm not entirely sure. Unfortunately Luca, who disabled it, has left. I know he was complaining about some performance issues he was having, but looks like the disable tests were pre-existing, so a bit unsure why that went out...

sdjmchattie:
> > Out of interest @JamesGlover why is there a JavaScript test marked as `xit` to make it pending? It also doesn't pass if you take the `x` away. It's highlighted in the bottom of the Diff on this PR.
> 
> I'm not entirely sure. Unfortunately Luca, who disabled it, has left. I know he was complaining about some performance issues he was having, but looks like the disable tests were pre-existing, so a bit unsure why that went out...

Maybe I should remove it while I'm working on the next branch, actually creating the change for https://github.com/sanger/limber/issues/774 ?

sdjmchattie:
@JamesGlover you previously requested a change.  I believe this is now ready to go.  If you could check it then I will get it merged.

JamesGlover (approved):

JamesGlover:
> > > Out of interest @JamesGlover why is there a JavaScript test marked as `xit` to make it pending? It also doesn't pass if you take the `x` away. It's highlighted in the bottom of the Diff on this PR.
> > 
> > 
> > I'm not entirely sure. Unfortunately Luca, who disabled it, has left. I know he was complaining about some performance issues he was having, but looks like the disable tests were pre-existing, so a bit unsure why that went out...
> 
> Maybe I should remove it while I'm working on the next branch, actually creating the change for #774 ?

Leave it for now, the describe behaviour is something that we'd want. It might be that the test is invalid due to the way Vue handles updates on a per-tick basis, but I'd need to look at the code more closely to work that out.

PR #798, opened 2021-08-11
DPL-083 Update workflows to automate releases and versioning

sdjmchattie:
Part of https://github.com/sanger/deployment/issues/71


.github/workflows/ruby_build.yml
```diff
@@ -1,33 +0,0 @@
-name: Ruby Build
```
JamesGlover, line 1: Just like with SS, could we restore this with the updated release action please.

lib/deployed_version.rb
```diff
@@ -40,20 +40,20 @@ def minor
...
+      @version_hash ||= /\Av(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)(?<extra>\S*)/.match(label)
     end
 
     def version_label
-      if version_hash
-        "#{major}.#{minor}.#{extra}"
-      else
-        'WIP'
-      end
+      major == 0 && minor == 0 && patch == 0 ? 'WIP' : "#{major}.#{minor}.#{patch}"
```
JamesGlover, line 56: I know you pulled this in from the other repos, but I think the version here (minus the change to 'patch') was actually correct.

If a rexep fails to match, it will return nil, which means the 

```ruby
    def version(rank)
      version_hash ? version_hash[rank] : '0' # String, as it matches what we'd get from the regex
    end
```
Will always return the string '0'. Yet the check here looks for equality with the number.

JamesGlover (changes requested):

lib/deployed_version.rb
sdjmchattie: Thanks!  Good point.  I hadn't appreciate there was a difference here.

.github/workflows/ruby_build.yml
sdjmchattie: 👍 I've done this but it's displayed separately above because of the correct indentation in the new file

JamesGlover (approved):

PR #814, opened 2021-09-02
Issue757/dpl 065 cardinal pbmc isolation and cell count input files

JamesGlover:
This is a first pass refactor to tidy up the ExportsController. Its mostly pulling existing configuration out, without further significant changes.

Long term:
- Consider switching to proper csv encoding, rather than erb files
- Support for other formats
- Support for tubes

But these are probably worth addressing as we come across a need for them.


docs/exports_yaml_files.md
```diff
@@ -0,0 +1,121 @@
...
+
+If this is supplied, the {ExportsController} will identify the first ancestor
+(earliest) of the current plate with this purpose, and will assign it to
+the instance variable `@ancestor_plate`.
+
+Example in `app/views/exports/duplex_seq_pcr_xp_concentrations_for_custom_pooling.csv.erb`
+
+```yaml
+ancestor_purpose: LDS AL Lib Dil
+```
```
andrewsparkes, line 121: Great docs!
Noticed we appear to have tests in specs/views/exports for each of the different 'types' of export file that we create (i.e. not one for every single csv export in the configs). Maybe worth documenting that if you create a new file type/format it needs a test there?

andrewsparkes (approved):

docs/exports_yaml_files.md
```diff
@@ -0,0 +1,121 @@
...
+
+[More details](https://jsonapi.org/format/#fetching-includes)
+
+```yaml
+plate_includes: wells.transfer_requests_as_target.source_asset
+```
+
+Default: `wells`
+
+#### workflow
```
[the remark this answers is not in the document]
andrewsparkes: Ok she says it depends. On pipelines with multiple files they require that field for the robot method to check they are using the right file at the right step. If the pipeline / robot method only requires one file then it is not essential. i.e. validation can be written into the robot method to check for that field value if required.

docs/exports_yaml_files.md
JamesGlover: That makes sense, as you're testing the view at that point. There's also some tests in the controller specs. I'll make a note pointing to these.

PR #819, opened 2021-09-08
Issue757/dpl 065 cardinal pbmc isolation and cell count input files and config

### LINKED ISSUE #757: DPL-065 Cardinal - PBMC isolation and cell count input files
**User story**
As a user of the Cardinal pipeline, I need the plate containing the isolated PBMCs (peripheral blood mononuclear cells) to be recorded in the LIMS, and be able to generate files for the cell-counting Cellaca machine to use.

Stef L

**Acceptance criteria**
- I can create an "LCA PBMC Plate", as a stamp of the "LCA Blood Plate".
- I can print out the barcode for above plate.
- I can print out barcodes for up to 4 additional QC plates.
    - These can have the same barcode as the LCA PBMC plate but with a suffix of QC1,2,3 or 4
    - Only 1 QC barcode will be needed if there are 24 samples or less, only 2 QC barcodes if there are 25-48 samples, etc
- I can generate Cellaca input files
    - The Cellaca can process 24 samples at a time (see input file layout in Figure 1). 
    - The LIMS needs to populate the sample name into the respective field (instead of the 'A1', 'A2' etc shown). 
    - If more than 24 samples are in the plate, additional file(s) need to be created. 
    - They can be labelled QC 1,2,3 or 4. 
    - The files can be downloaded in Limber.

image
_Figure 1. LIMS generated Cellaca upload file (.csv in a 8x3 format.)_

**Dependencies**
- pipeline setup - https://github.com/sanger/limber/issues/772
- LCA Blood Plate type from https://github.com/sanger/limber/issues/756 must be available to finish this story.

JamesGlover:
Closes #757

* Includes behaviour changes to:
  - Add Cellaca label template to print the required number of qc barcodes based on plate occupancy
  - Add Cellaca CSV generation
  - CSV download links can specify additional parameters
* Includes configuration changes to:
  - Use the new Cellaca csv for LCA PBMC
  - Use the new label template for LCA PBMC


app/models/labels/plate_label_cellaca_qc.rb
```diff
@@ -0,0 +1,44 @@
...
+      # Dividing a column by three maps it to its page
+      next if occupied_columns.none? { |col| col / COLS_PER_PAGE == index }
+
+      qc_label(index)
+    end
+  end
+
+  private
+
+  def qc_label(index)
```
andrewsparkes, line 24: This is better than what I had before :)
One comment, does it print them out in reverse order, i.e. 4,3,2,1 rather than 1,2,3,4?
Easy for the user if we order reversed here so when you print them they are in right order top to bottom on the physical labels (as it prints labels upside down essentially).

app/models/presenters/plate_presenter.rb
```diff
@@ -58,8 +59,8 @@ def number_of_wells
       "#{number_of_filled_wells}/#{size}"
     end
 
-    def pcr_cycles
-      pcr_cycles_specified.zero? ? 'No pools specified' : cycles.to_sentence
+    def requested_pcr_cycles
+      pcr_cycles.empty? ? 'No pools specified' : pcr_cycles.to_sentence
```
andrewsparkes, line 63: Is pooling related to PCR cycles? odd text

config/label_templates.yml
```diff
@@ -36,6 +36,6 @@ plate_lds_al_lib:
   :label_class: Labels::PlateLabelLdsAlLib
   :pmb_template: 'sqsc_96plate_label_template_code39'
```
andrewsparkes, line 37: You removed the single quotes around the plate_cellaca_qc pmb_template, maybe be consistent and change others if not needed?

config/purposes/cardinal.yml
```diff
@@ -39,12 +39,28 @@ LCA Blood Bank:
...
   :asset_type: plate
-  :label_template: plate_lca_pbmc # creates QC1 to 4 barcodes
-  # TODO: DPL-065 needs :file_links: csv output: for Cellaca input files for QC1 to 4 depending on number samples
-  # Cellaca output files (1-4) are uploaded into Quanthub to set concentrations on plate (DPL-066)
-  # TODO: DPL-065 needs :file_links: csv output: download of combined Cellaca concentrations uploaded via Quanthub
-  # TODO: DPL-065 Research: how do we then flag on the SAME plate that viability or cell counts are sub standard?
-  # TODO: DPL-065 new :presenter_class: Presenters::?CardinalPBMCPlatePresenter needs a special presenter to pick and display viability and cell count?
+  :label_template: plate_cellaca_qc # creates QC1 up to QC4 barcodes
+  :file_links:
+    - name: "Download Cellaca Input QC1"
```
andrewsparkes, line 44: We should update the Integration Suite test to check for the presence of these file download buttons once merged

spec/models/labels/plate_label_cellaca_qc_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+RSpec.describe Labels::PlateLabelCellacaQc, type: :model do
+  it { expect(described_class).to be < Labels::Base }
+
+  describe '#attributes' do
+    subject(:qc_attributes) { label.qc_attributes }
+    let(:label) { described_class.new(labware) }
+
+    context 'when creating the label of a full plate' do
+      let(:labware) { create :v2_plate, pool_sizes: [96] }
```
andrewsparkes, line 13: Aside: Pooling question: The current multistamp assigns a new pool number for each source plates wells, e.g. multistamp from 3 sources get 3 pools. For cardinal we have up to 96 tubes, so does that equal 96 distinct pools? Or is the whole pool thing irrelevant at the tube arraying stage as we've no submissions yet?

andrewsparkes (approved):

app/controllers/exports_controller.rb
```diff
@@ -9,12 +9,11 @@ class ExportsController < ApplicationController
...
 
   def show
+    @page = params.fetch(:page, 0).to_i
     @workflow = export.workflow
-    if export.ancestor_purpose.present?
-      ancestor_result = @plate.ancestors.where(purpose_name: export.ancestor_purpose).first
-      locate_ancestor(ancestor_result.id) if ancestor_result.present?
-    end
-    render export.csv
+    @ancestor_plate = locate_ancestor
```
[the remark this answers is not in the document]
KatyTaylor: I think the way I've set up the pipeline now, the blood array and blood bank plates should be the entry points to the Cardinal Library Prep and Cardinal Banking pipelines, so yes they should ideally be the stock and input plates.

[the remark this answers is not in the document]
JamesGlover: The flags won't work for tubes, and just get ignored by limber. 

I had a quick chat with Katy about this, and it was decided that the plates would both have the initial and stock flags set, and the submission would then be built from these. The transfer from the tubes to the plates is then tracked by the transfer requests only. This means that each plate is the stock to its own descendants, and as far as each library prep process is concerned is the start of the process.

There didn't seem to be much value in generating a request to go alongside that first transfer, as it wouldn't track any additional information, and would just delay the whole process. (Especially as you probably would want to build it in the plate creator, to avoid the need to create one per tube. Which means the creator would need to twiddle its thumbs while the submission built, only to almost immediately pass it)

app/models/labels/plate_label_cellaca_qc.rb
JamesGlover: Ahh, no, I'd need to add a reverse on the end for that.

app/models/presenters/plate_presenter.rb
JamesGlover: The only scenario where we could reasonably expect no PCR cycles to be specified, is one in which there were no pools defined on the plate.

config/label_templates.yml
JamesGlover: Aggh, sorry, I was fighting with my linter/formatter at this point. It was auto-correcting everything on save, and I wanted to avoid touching more than necessary as I know there are a lot of people working on the same files at the moment. I'd love to get prettier enabled for limber, but want to wait until things are a bit quieter as otherwise I think someone my try to kill me when they merge in their work.

spec/models/labels/plate_label_cellaca_qc_spec.rb
JamesGlover: Yep, pool numbers are a symptom of submission, so are irrelevant here.I'm not quite sure what you mean by multi-stamp assigning pool numbers though?

PR #821, opened 2021-09-08
Switch to jest

JamesGlover:

* Switch to jest for test running
* Cleaner output
* No deprecation warnings
* Vastly faster tests (~10x)


app/javascript/custom-tagged-plate/components/CustomTaggedPlate.spec.js
```diff
@@ -15,7 +15,7 @@ import {
   exampleTag2Group,
   exampleTag2GroupLonger,
   exampleQcableData,
-  exampleTag1GroupChromium
+  exampleTag1GroupChromium,
```
andrewsparkes, line 18: The comma on the last item in a list thing, is that the standard with JS? came up in discussion with another developer and we weren't sure if/why it should be there other than a convenience thing if you want to add more items later...

app/javascript/custom-tagged-plate/components/CustomTaggedPlate.spec.js
```diff
@@ -43,106 +43,106 @@ describe('CustomTaggedPlate', () => {
         expect(wrapper.vm.isChildWellsValid).toEqual(false)
       })
 
-      it('returns false if any wells with aliquots are invalid', () => {
+      it('returns false if any wells with aliquots are invalid', async () => {
```
andrewsparkes, line 46: So when is async required (with await)? Is it best practice anytime you set data? Does that give it chance to run the computed functions etc. before your expect?

babel.config.js
```diff
@@ -53,6 +53,12 @@ module.exports = function(api) {
           useBuiltIns: true
         }
       ],
+      [
+        '@babel/plugin-proposal-private-property-in-object',
+        {
+          loose: true
```
andrewsparkes, line 59: Is this for speed/performance?

andrewsparkes (approved):
Looks good to me but maybe someone with Jest experience should check it too? Steve?

README.md
```diff
@@ -88,12 +88,12 @@ Ruby unit and feature tests:
...
 ```
 
-### Karma
+### Jest
 
 JavaScript unit tests:
 
 ```bash
-yarn karma start --single-run
+yarn test
```
andrewsparkes, line 96: Maybe add an example of how to run a single spec file? Seems to be a common question with testing.

stevieing (approved):

app/javascript/custom-tagged-plate/components/CustomTaggedPlate.spec.js
JamesGlover: It depends on how the linter is configured, as essentially it's optional. It seems most style guides are trending towards including it. I assume ES lint is currently set up not to care either way, and this was probably added by VS Code before I turned of prettier in repos where it isn't explicitly enabled.

But yeah, you're right. The main driver for this is that if you add a new item, the diff will be only one line, rather than showing the new item, and the addition of the comment on the previous line. This also potentially makes merging slightly easier if there is a conflict with two entries added, as you can just accept both changes.

For some reason, the ruby community has settled on the opposite, favouring the lack of a trailing comma. This is despite ruby supporting it just fine:
```
irb(main):002:0> [1,2,3,]
=> [1, 2, 3]
irb(main):003:0> {a:1,b:2,c:3,}
=> {:a=>1, :b=>2, :c=>3}
```

app/javascript/custom-tagged-plate/components/CustomTaggedPlate.spec.js
JamesGlover: This was one of the changes that came out of the vue-test-utils update (sorry to base off that, but I didn't want to migrate the test suite with an outdated library that I knew was causing failures).

But yeah, all the vue documentation examples use await on both setting data and props. Most of these changes were practical (ie. they actually fixed a new failure) rather than theoretical, but I don't really see reason to skip it.

babel.config.js
JamesGlover: It was fixing a deprecation warning. Essentially, this is what it was doing anyway, thanks to a similar configuration option elsewhere. I had a bit of a look at what it actually means, and loose is essentially faster, with slightly better compatibility, as the cost of not being quite so strict. I decided to maintain existing behaviour.

Issue #768, opened 2021-07-13
DPL-068 Cardinal - Sample pooling robot driver file


KatyTaylor:
**User story**
As a user of the Cardinal pipeline, I need to have a file that I can upload to the Hamilton robot, so that it knows how to transfer samples into the LCA PBMC Pools plate, as determined by https://github.com/sanger/limber/issues/759.

**Who are the primary contacts for this story**
Stef L

**Acceptance criteria**
- The volume of each sample to be added to the pool needs to be calculated from the previously uploaded Cardinal_PBMC_Count file / qc results. 
    - 20,000 cells need to be taken from each sample into the pool. The volume that needs to be taken to achieve this needs to be calculated from the live cell count. E.g. 1,000,000 live cells per ml (20,000/1,000,000) = 0.02ml (20ul). 
  
- A resuspension volume also needs to be specified per pool. The volume is fixed and specified in Figure 4.  

![image](https://user-images.githubusercontent.com/10407809/124908187-f6ba3980-dfe0-11eb-8230-05e1d48b4f34.png)

_Figure 4: resuspension volume_

- The LIMS needs to generate a Hamilton driver file that states: source plate, source well, destination plate, destination well, sample volume and resuspension volume. The destination well identifies the pool e.g. all samples going into well A1 are in a single pool. The sample volume reflects the amount of sample volume that needs to go into the pool (20,000 cells per sample). The resuspension volume is taken from Figure 4 and depends on the number of samples in the pool.
**UPDATE: this was actually implemented as resuspension volume = 3.125 * num samples**

Example file: [LIMS_Pool_Example.csv](https://github.com/sanger/limber/files/7084772/LIMS_Pool_Example.csv)
More realistic example file (from running integration suite test after development: [hamilton_pooling_plate_pbmc.csv](https://github.com/sanger/limber/files/7718239/hamilton_pooling_plate_pbmc.csv)

**Dependencies**
pipeline setup - https://github.com/sanger/limber/issues/772
Probably blocked by https://github.com/sanger/limber/issues/759 - or maybe need to be thought about together.

andrewsparkes:
See Limber exports_controller.rb and the file_links in config/purposes files to see how to make the csv file download links.

KatyTaylor:
Info from Slack:

Andrew Sep 3rd

[...]

PR #829, opened 2021-09-13
DPL-068: Hamilton driver file export

sdjmchattie:
For #768 

* Add a new CSV exporter for Hamilton driver file on LCA PBMC Pools plates.
* Export the transfer of samples from the source plate to the pool plate.
* Add the resuspension volumes (3.125µl per pooled sample) to each transfer.

Still to do:

* Put the actual transfer quantity in column 5.


app/sequencescape/sequencescape/api/v2/well.rb
```diff
@@ -107,4 +107,13 @@ def control_info_formatted
...
   end
+
+  def each_source_barcode_and_location_from_transfer_requests
+    transfer_requests_as_target.each do |transfer_req|
+      # NB. Making assumption here that name field on asset is for a plate well
+      # and contains a plate barcode and well position e.g. DN12345678:A1
+      name_array = transfer_req.source_asset.name.split(':')
+      yield name_array[0], name_array[1] if name_array.length == 2
+    end
+  end
```
JamesGlover, line 118: It's probably safer (albeit slightly slower) to include the labware in the request and pull out the plate barcode explicitly. You can configure includes in the exports.yml file. You might find you need to specify the inclusion of 'labware' rather than plate as (schema wise) you can't be sure if the upstream assets are wells or tubes.

JamesGlover: I also wonder if this is a bit too specific/view centric to be on well, and might be better as a helper method for the exports controller/views.

JamesGlover: Oh dear, I've done that thing haven't I. Where a refactor draws attention so something.

config/purposes/cardinal.yml
```diff
@@ -44,19 +44,19 @@ LCA PBMC:
     - name: "Download Cellaca Input QC1"
       id: cellaca_input_file
       params:
-        - page: 0
+        page: 0
```
JamesGlover, line 47: Thought I'd already merged this fix in.

JamesGlover (changes requested):
If you run into major issues pulling back the labware via an include, don't worry about diving into the depths of the API to fix it and we'll run with this (STI and polymorphism support can be a bit troublesome on both sides of the API)

config/purposes/cardinal.yml
sdjmchattie: It is weird that it's showing up here as the change is already on the branch I'm trying to merge into.  I think it's because I branched off before your commit and GitHub is trying to be clever and show you what changed in my branch since it was created.  I'll merge the destination branch into mine and I expect they'll magically disappear.

app/sequencescape/sequencescape/api/v2/well.rb
sdjmchattie: I moved this to a new helper and extracted more of the common logic as well

app/helpers/exports_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+# Helper methods for the Exports controller
+module ExportsHelper
+  def each_source_metadata_for_plate(plate)
+    index = 0
+    plate.wells_in_columns.each do |well|
+      well.transfer_requests_as_target.each do |transfer_req|
+        # NB. Making assumption here that name field on asset is for a plate well
+        # and contains a plate barcode and well location e.g. DN12345678:A1
+        name_array = transfer_req.source_asset.name.split(':')
+        if name_array.length == 2
```
JamesGlover, line 12: I don't fully understand this check, as I'm not quite sure why we're expecting wells wells which don't conform to this format (I guess if we see a transfer from a tube?), or, if we do come across them, why we just ignore them silently.

I know this is previous code, and I'm certainly not suggesting it be removed (as I'm assuming there *was* a reason the check was added) just wondering if you had any insight?

JamesGlover (approved):

app/helpers/exports_helper.rb
sdjmchattie: I've felt the same way about it.  They are of course silently ignored.  The only reason this would not be the case is if the implementation changes for the name of the well, which I think is defined here: https://github.com/sanger/sequencescape/blob/develop/app/models/well.rb#L323

I would recommend putting a comment near where the name is defined to say that changing the format would affect this code, but as it may be in a different repo entirely that might not be a good idea.  Of course if the split didn't work and we try to reference index 1 in a single item array then we're going to generate exceptions.  I don't know if that's preferable to missing out results from the CSV.

JamesGlover: I'm generally one to favour blowing up loudly if we see something unexpected, or at the very least with a little bit of noise. Although I'd probably handle it specifically in this case, with a custom exception. 

But as you say, this is already coupling behaviour to another attribute in a somewhat unpredictable manner (Although I can't imagine this is the only piece of code where something is relying on this, and that goes for customer reports as well). Perhaps the time would be better spent adding the relevant associations to the API so that the information can be pulled out formally. I'll pop a story in the backlog now, as its a pretty minor thing, but its daft that you cant access a plate from its wells.

JamesGlover: PR with warning: https://github.com/sanger/sequencescape/pull/3343
The link is currently broken, but should resolve correctly once this is merged in.

JamesGlover: And for completeness the issue: https://github.com/sanger/sequencescape/issues/3342

PR #833, opened 2021-09-17
Dpl067 pooling random

emrojo:
Closes DPL067

* Added new labware creator to pool samples using different tag depth


app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+# require_dependency 'form'
+# require_dependency 'labware_creators'
+
+# Steps:
+# 1. Create a new empty LCA PBMC Pools plate
+# 2. Get the passed wells from the parent (LCA PBMC) plate
+# 3. For the number of passed wells, get the number of pools from config
+# e.g. if there are 96 passed wells on the parent, the samples get split into 8 pools, with 12 samples per pool
+# 4. Group samples by supplier, to ensure samples with the same supplier are distrubuted across different pools
+# 5. Add the pool to a well in the new LCA PBMC Pools plate
```
harrietc52, line 16:
```suggestion
# 5. Create a Transfer Requests in SS - adding the pool to a well in the new LCA PBMC Pools plate
```

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    def well_filter
+      @well_filter ||= WellFilter.new(creator: self)
+    end
+
+    # Send the transfer request to SS
+    def transfer_material_from_parent!(dest_uuid)
+      dest_plate = Sequencescape::Api::V2::Plate.find_by(uuid: dest_uuid)
+      api.transfer_request_collection.create!(
```
harrietc52, line 53: Do we want to handle if this request to create TransferRequests fails in some way?

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    # returns: a list of objects, mapping source well to destination well
+    # e.g [{'source_asset': 'auuid', 'target_asset': 'anotheruuid'}]
+    def transfer_request_attributes(dest_plate)
+      passed_parent_wells.map do |source_well, additional_parameters|
+        # TODO: revert back to the below
+        request_hash(source_well, dest_plate, additional_parameters)
+
+        # Temp fix below is for testing until we have the tag clash on pooling issue resolved
```
harrietc52, line 67: Can remove temp fix comments here, and comment on line 64

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    def request_hash(source_well, dest_plate, _additional_parameters)
+      {
+        'source_asset' => source_well.uuid,
+        'target_asset' => dest_plate.wells.detect do |dest_well|
+          dest_well.location == transfer_hash[source_well.location][:dest_locn]
+        end&.uuid,
+        # TODO: add concentration/ cell count here?
+        # 'volume' => "12345" #transfer_hash[source_well.location]['volume'].to_s
+        'aliquot_attributes': { 'tag_depth' => tag_depth(source_well) }
```
harrietc52, line 81: I wonder whether this object should be part of additional_parameters? Passing into this method `'aliquot_attributes': { 'tag_depth' => tag_depth(source_well) }` as `additional_parameters` ?

harrietc52: We can also remove volume if not needed

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+          result[source_position] = { dest_locn: destination_well_location }
+        end
+      end
+      result
+    end
+
+    def tag_depth(source_well)
+      @pools.each do |pool|
+        return (pool.index(source_well) + 1).to_s if pool.index(source_well)
+        # index + 1 incase of 0th index
```
harrietc52, line 111: Comment for this method, explaining the +1 too as is isn't currently clear with the current comment

config/purposes/cardinal.yml
```diff
@@ -39,8 +39,11 @@ LCA PBMC:
 # LCA PBMC wells are pooled together into the LCA PBMC Pools plate
 LCA PBMC Pools:
   :asset_type: plate
+  #:presenter_class: Presenters::CardinalPoolsPlatePresenter
   # TODO: DPL-067 :presenter_class: Presenters::?CardinalPoolsPlatePresenter shows pool info for wells, samples in each well
-  # TODO: DPL-067 :creator_class: LabwareCreators::?Plate new creator to make pools of samples automatically according to table, using info from PBMC QC
+  # TODO: DPL-067 :creator_class: LabwareCreators::?Plate
```
harrietc52, line 44: remove this line

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+      api.transfer_request_collection.create!(
+        user: user_uuid,
+        transfer_requests: transfer_request_attributes(dest_plate)
+      )
+      true
+    end
+
+    # returns: a list of objects, mapping source well to destination well
+    # e.g [{'source_asset': 'auuid', 'target_asset': 'anotheruuid'}]
+    def transfer_request_attributes(dest_plate)
```
harrietc52, line 62: Could we also make some of the helper methods here private?

config/initializers/cardinal_config.rb
```diff
@@ -0,0 +1,13 @@
...
+require 'csv'
+cardinal_pooling_csv = CSV.read('config/cardinal_pooling.csv', { encoding: 'UTF-8', headers: true, header_converters: :symbol, converters: :all })
+
+Rails.application.config.cardinal_pooling_config = {}
+
+cardinal_pooling_csv.each do |data|
+  contents = data.to_hash
+  number_of_pools = data.to_hash.values.compact.count - 1
+
+  Rails.application.config.cardinal_pooling_config[contents[:number]] = number_of_pools
```
harrietc52, line 12: Comment from Stuart: 

Those pooling strategies look a lot better than before.  They actually look like you could write a more general rule than having a config file.
Something like, upto 20 samples is a single pool, then multiples of 13 are the transition from x pools to x+1 pool upto 76 samples where it changes to multiples of 11.  

Todo: check whether there should be more than 13 samples in a pool, and that the CSV is correct

harrietc52 (changes requested):
Looking good so far :) just cleaning up left 

Todo
- [ ] Clean up comments etc
- [ ] Update presenter to display pools, and number of samples in each pool, for a destination well
- [ ] Check with Stef about max volume in a pool/ max number of samples up to 20 in a pool, or 13? Possibly refactor config

app/models/labware_creators/cardinal_pools_plate.rb
harrietc52: Add a test to show how the exception is handled

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
...
+      ('A'..'H').to_a.map { |letter| "#{letter}1" }
+    end
+
+    # returns: an object mapping a source well location to the destination well location
+    # e.g. { 'A1': { dest_locn: 'B1' }, { 'A2': { dest_locn: 'A1' }, { 'A3': { dest_locn: 'B1' }}
+    def transfer_hash
+      result = {}
+
+      # Build only once, as this is called in a loop
+      @pools ||= build_pools
```
harrietc52, line 97: test to check empty @pools

Gemfile
```diff
@@ -20,7 +20,7 @@ group :default do
   gem 'json_api_client', github: 'sanger/json_api_client', branch: 'merge_upstream'
 
   # Older Sequencescape API
-  gem 'sequencescape-client-api', '~> 0.6.1', require: 'sequencescape'
+  gem 'sequencescape-client-api', '~> 0.6.1', require: 'sequencescape', path: '/Users/hc6/psd/gems/sequencescape-client-api'
```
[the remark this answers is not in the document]
harrietc52: @harrietc52

app/models/labware_creators/cardinal_pools_plate.rb
BenTopping: Ill do this one

app/models/labware_creators/cardinal_pools_plate.rb
```diff
@@ -0,0 +1,139 @@
+# frozen_string_literal: true
+
+# TODO
+# Update presenter to display pools, and number of samples in each pool, for a destination well
+# Generate CSV
```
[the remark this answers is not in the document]
harrietc52: @harrietc52

app/models/labware_creators/cardinal_pools_plate.rb
BenTopping: done

app/models/labware_creators/cardinal_pools_plate.rb
harrietc52: Leaving them for now and useful to be tested independently

config/initializers/cardinal_config.rb
harrietc52: Going to leave this for now as the CSV makes it clear how many pool are allocated for a number of passed samples

app/models/labware_creators/cardinal_pools_plate.rb
harrietc52: Also, leave for now, think its clearer like this and we aren't sure exactly how additional_parameters are handled

PR #837, opened 2021-09-20
DPL-068: Hamilton driver CSV export

### LINKED ISSUE #768: DPL-068 Cardinal - Sample pooling robot driver file

sdjmchattie:
Closes #768 

* Parse QC results in the output file plus tests


app/sequencescape/sequencescape/api/v2/well.rb
```diff
@@ -31,6 +31,10 @@ def latest_molarity
     latest_qc(key: 'molarity', units: 'nM')
   end
 
+  def latest_live_cell_count
```
stevieing, line 34: The live_cell_count stuff. Just wondering if using the units cells/ml is actually working? The display name is cells/ml but the actual unit is cells.

Also where does the viability qc results come in. Are they needed yet?

sdjmchattie: I probably need to adjust this then.  I didn't know the internal name was different to the display name.

I don't think viability comes into it at all for this CSV file.  They were used to filter samples out of the previous plate so I won't have any with unwanted viability percentages.

sdjmchattie: I've checked the SequenceScape code and I'm still pretty confident that it's `cells/ml` we want here.  I'm basing this on the results of your unit test in https://github.com/sanger/sequencescape/blob/develop-cardinal/spec/models/parsers/cardinal_pbmc_count_parser_spec.rb#L136

I also used the UAT actions to generate some random live cell counts and the Units column was populated with `cells/ml`.

Issue #770, opened 2021-07-13
DPL-072 Cardinal - Custom pooling (M)


KatyTaylor:
**User story**
As a Cardinal pipeline user, I need to pool together samples from multiple LCA Chromium Connect PCR XP plates, into a tube. The wells to be pooled, and the volumes coming from each source well will be set based on the information I upload in a file.

**Who are the primary contacts for this story**
Stef L

**Acceptance criteria**
- Output labware is a tube / tubes called "LCA Custom Pool"
- There is one ~can be one or more~ "LCA Connect PCRXP" input plate~s~ (see additional context)
- Input file format is a csv with headers "Source Well", "Volume to add to pool", and "Dest. well (?)", as in Duplex-seq pipeline
- Look to see if code checks for tag clashes
- I can create more than one child tube by uploading 2 different files.

**Dependencies**
This story is blocked by the following dependencies:
pipeline setup - https://github.com/sanger/limber/issues/772

**Additional context**
Update after into to LIMS with Stef & Cardinal team on  30/09/2021:

Stef is assuming this page only allows pooling from one plate into multiple tubes, rather than multiple plates into multiple tubes - this can be achieved using 'poolings/new' afterwards

andrewsparkes:
Existing custom pooling functionality in Limber should work for this, although not sure about volume to transfer.

KatyTaylor:
research label added for same reason as on https://github.com/sanger/limber/issues/769

PR #851, opened 2021-09-30
DPL-071: Generate pooling CSV for Cardinal PCRXP plate

### LINKED ISSUE #769: DPL-071-1 Cardinal - download template for pooling decision
**User story**
As a user of the Cardinal pipeline, I would like to be able to download information about the wells on a LCA Connect PCRXP plate, so that I can work out the pooling strategy outside the LIMS.

Stef L

**Acceptance criteria**
I can download a file from the LCA Connect PCRXP plate in Limber, containing the following for each well:
- well ID
- If possible, the number of samples that were previously combined into that particular well (in this story - https://github.com/sanger/limber/issues/759)
- tag index
- Depending on difficulty, implement this for multiple source plates
- The hope is that this file can act as a template for the file that is uploaded in https://github.com/sanger/limber/issues/770

**Dependencies**
pipeline setup - https://github.com/sanger/limber/issues/772
https://github.com/sanger/limber/issues/761

harrietc52:
Closes #769

This adds the CSV to the pipeline for using as an input in #770 .  The number of samples in a pool is inaccurate right now (random numbers) but we should probably get the rest of this in the codebase in the mean time so it doesn't sit here forever.


app/views/exports/cardinal_tagging_csv_for_custom_pooling.csv.erb
```diff
@@ -0,0 +1,14 @@
...
+  csv_array = [['Source Well', 'Volume to add to pool', 'Dest. pool', 'Number of samples', 'Tag index', 'Tag 2 index']]
+  @plate.wells_in_columns.each do |well|
+    csv_array << [
+      "#{@plate.human_barcode}:#{well.location}",
+      nil,
+      nil,
+      rand(12) + 1, # TODO: DPL-067 Determine actual number of samples per well
+      well.aliquots[0].tag_index,
+      well.aliquots[0].tag2_index
+    ] unless well.empty? || well.aliquots.size > 1 # TODO: DPL-067 Verify the assumption that a well only has one aliquot
```
KatyTaylor, line 11: Even if the well has multiple aliquots (i.e. not an 'aggregate sample'), they should all have the same tags. Not actually quite sure if the tagging step deals with wells with multiple samples/aliquots, but that's how it should be I think!

app/views/exports/hamilton_pooling_plate_pbmc.erb
```diff
@@ -37,6 +37,4 @@
     data + [samples_in_pool * 3.125]
   end
 %>
-<% rows_array.sort_by{ |a| WellHelpers.well_coordinate(a[1]) }.each do |row| %>
-<%= CSV.generate_line row, row_sep: '' %>
-<% end %>
+<%= rows_array.sort_by{ |a| WellHelpers.well_coordinate(a[1]) }.map(&:to_csv).join %>
```
KatyTaylor, line 40: Did you test this still works? (Just a bit cautious of that `(&:)` syntax as it mysteriously doesn't work in some situations...

sdjmchattie: Yes the unit tests confirm that the generated CSV is identical.

app/views/exports/cardinal_tagging_csv_for_custom_pooling.csv.erb
sdjmchattie: Ah OK so we should not bother checking for number of aliquots then?  Just take the first one?  Or we should validate that we only have a single unique index across all aliquots?  The latter sounds safer.

KatyTaylor (approved):

app/views/exports/cardinal_tagging_csv_for_custom_pooling.csv.erb
KatyTaylor: Yes the latter sounds nice and robust

PR #859, opened 2021-10-04
DPL064 Sample Arraying - adding sample arraying tube layout file download

andrewsparkes:


spec/views/exports/cardinal_sample_arraying_tube_layout.csv.erb_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    expected_plate_header = ['Plate Barcode', barcode]
+    expect(parsed_csv[0]).to eq expected_plate_header
+
+    # check for row headers in column 1
+    expect(get_column(parsed_csv, 0)).to eq([nil, nil, 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H'])
+
+    # chek for tube barcode content in wells
+    expect(get_column(parsed_csv, 1)).to eq([nil, '1', 'NT2P', 'NT3Q', 'NT4R', 'empty', 'empty', 'empty', 'empty', 'empty'])
+    expect(get_column(parsed_csv, 7)).to eq([nil, '7', 'empty', 'empty', 'empty', 'empty', 'NT5S', 'empty', 'empty', 'empty'])
+    expect(get_column(parsed_csv, 12)).to eq([nil, '12', 'empty', 'empty', 'empty', 'empty', 'empty', 'empty', 'NT6T', 'empty'])
```
KatyTaylor, line 43: Why is the first cell blank in every column? Looks like the second cell is the header?

config/purposes/cardinal.yml
```diff
@@ -36,6 +36,9 @@ LCA Blood Array:
         fragment_size_required_from: '200'
         fragment_size_required_to: '800'
       allowed_extra_barcodes: false
+  :file_links:
+    - name: "Download Scanned Tube Layout for LCA Blood Array Plate"
+      id: cardinal_sample_arraying_tube_layout
```
KatyTaylor, line 41: I wonder if you could also add one for the LCA Blood Bank Plate, while you're at it?

KatyTaylor (approved):

spec/views/exports/cardinal_sample_arraying_tube_layout.csv.erb_spec.rb
andrewsparkes: There's a spacer row between the header row with the plate barcode in it and the top of the table of wells, that's all it is.

PR #894, opened 2021-10-28
DPL-110: Group arbitrary tubes

sdjmchattie:
Completes most of #793 

* Add a new page to Limber for creating tube racks.
* Support scanning in tubes in a list of 16 positions.
* Check for duplicate scanned messages.
* Check for tubes that don't match in purpose.

image


app/javascript/shared/components/tubeScanValidators.js
```diff
@@ -53,16 +55,25 @@ const checkDuplicates = (tubeList) => {
...
-// 
-// Returns an object with the right tube scan validation message
-const validTubeScanMessage = () => {
-  return { valid: true, message: 'Great!' }
+// Returns a validator that ensures the purpose namess of all the tubes match
+// the name for the one provided. Typically the one provided should be the one
+// of the purposes from the full set of tubes being validated.
+const checkMatchingPurposes = (purpose) => {
+  return (tube) => {
+    if (tube && purpose && tube.purpose.name !== purpose.name) {
```
JamesGlover, line 68: We could use the safe navigation operator here: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Also, with the current code if you *do* end up with a tube without a purpose (unfortunately they exist) the error message generation will currently fail. I've been using 'UNKNOWN' in the presenters in the case a tube has no purpose. (I'm hoping we won't see them)

app/javascript/tubes-to-rack/components/TubesToRack.vue
```diff
@@ -0,0 +1,131 @@
...
+      </b-card>
+    </lb-main-content>
+    <lb-sidebar>
+      <b-card
+        header="Scan tubes"
+        header-tag="h3"
+      >
+        <b-form-group
+          label="Scan the tube barcodes into the relevant rack coordinates:"
+          style="position:relative; height:460px; overflow-y:scroll;"
```
JamesGlover, line 27: Pop the style in the style section of the component rather than inline.

Also, not entirely sold on the fixed height, and it might be worth using a percentage, although I'm aware with CSS this is sometimes easier said than done when it comes to heights.

app/javascript/tubes-to-rack/components/TubesToRack.vue
```diff
@@ -0,0 +1,131 @@
...
+            v-for="i in tubeCount"
+            :key="i"
+            :api="devourApi"
+            :label="indexToName(i - 1)"
+            :fields="tubeFields"
+            :includes="tubeIncludes"
+            :validators="scanValidators"
+            :colour-index="i"
+            :labware-type="'tube'"
+            :valid-message="''"
```
JamesGlover, line 39: Huh, so the valid message is getting passed in as a prop... (I've been jumping about a bit in this review, so sorry if this comment appears before the one its a reference to)

app/javascript/tubes-to-rack/components/TubesToRack.vue
```diff
@@ -0,0 +1,131 @@
...
+  components: {
+    'lb-labware-scan': LabwareScan,
+    'lb-loading-modal': LoadingModal
+  },
+  props: {
+    // Sequencescape API V2 URL
+    sequencescapeApi: { type: String, default: 'http://localhost:3000/api/v2' },
+
+    // Number of source tubes
+    tubeCount: { type: Number, default: 16 }
```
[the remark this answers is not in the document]
JamesGlover: Combined with some of the other thoughts, it may be better to pass in the rack dimensions, and calculate the size. (I've been going with numberOfRows and numberOfColumns where I can, as it feels less ambiguous than height and width)

I'll also briefly ask how you're planning to deal with the tube-rack purpose? There are currently 3 registered in SS, two 96 size (or which we probably only want to care about TR96) and a 48 well size (TR48). Typically the purpose is passed to the controller as a parameter, which means the page itself will already know what it's creating. Happy to come back to this one though, as there's still a little glue needed to make the buttons also appear on the tube pages.

JamesGlover (commented):
There's obviously the controller side stuff still to come, and there's a few questions about dimensions that I suspect are just being deferred for the moment.

app/javascript/shared/components/scanValidators.js
```diff
@@ -0,0 +1,18 @@
...
+    return aggregate.valid ? validator(item) : aggregate
+  }, validScanMessage())
+}
+
+// Return object for a valid scan
+//
+// Returns an object with a scan validation message
+const validScanMessage = () => {
+  return { valid: true, message: 'Great!' }
+}
```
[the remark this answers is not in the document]
sdjmchattie: Ha yes I see.  Well perhaps you're right.  I know Andrew has overridden the valid message on the component for the new tube scanning array.  I think the "Great!" message should be the default and be defined in the component itself.  I will change this.

app/javascript/tubes-to-rack/components/TubesToRack.vue
sdjmchattie: I'll take a look at this, but these lines of code were copied verbatim from an existing Vue component so I'll need to fix it there as well.

app/javascript/shared/components/tubeScanValidators.js
sdjmchattie: I've used the safe navigation on the `purpose` for the given `tube` but I don't think we should be removing the other checks for `tube` and `purpose` being defined.  The reason is covered in the unit tests.

The `purpose` that was passed in could be `undefined` and this is not an error case.  This happens when the user hasn't provided any tubes yet and then they provide their first tube.  No tube can be valid until at least one is entered and the purpose to match against is taken from valid tubes so you have a chicken and egg unless you allow all tubes to pass when the purpose is `undefined`. 

Similarly, the `tube` can be `undefined` and that's legitimate behaviour.  The empty tube rack positions do submit for validation and they have no tube yet.

app/javascript/tubes-to-rack/components/TubesToRack.vue
sdjmchattie: I wasn't able to find a better solution for the height as it's quite heavily nested and the parent is relying on the child for height constraints.  I tried using viewport units but they didn't really work better than this fixed height does.

JamesGlover (approved):
Although may be a good idea to remove the links on this branch before merging, to reduce the risk of it going live.

PR #905, opened 2021-11-03
DPL-110: Submit tube rack

### LINKED ISSUE #793: DPL-110 - add the ability to group arbitrary tubes together in a tube rack [Heron] [TOR-67]
**User story**
add the ability to group arbitrary tubes together in a tube rack [Heron]. We need to add support to quanthub for tapestation files. These quants are tube based. Users would like to use Limber to create normalisation templates. 

Iraad / Steve / James

**Acceptance criteria**
- [ ] the ability to group arbitrary tubes together in a tube rack
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Users want to generate CSV containing the concentration of the source tubes. (I suggested that we could just generate the robot driver files, rather then them needing to copy-paste the information into an intermediate spreadsheet). However, both approaches require a means of grouping together multiple tubes for processing at the same time.

While this doesn’t need to be a tube rack, we:
- Already have the back-end support for this
- Need to add tube racks shortly for other purposes
- 
A meeting about Tapestation for Stef's work led to a conversation about this work for Naomi as well.  See https://github.com/sanger/quanthub/issues/102#issuecomment-906259713

sdjmchattie:
Closes #793 

* Supplementary changes for creating a tube rack.  Doesn't work yet but this is indicated in the Vue file.


config/routes.rb
```diff
@@ -51,6 +51,11 @@
     resources :qc_files, controller: :qc_files
   end
 
+  scope 'tube_rack', controller: :tube_rack_creation do
+    get  'new', action: :tubes_to_rack, as: :new_tube_rack
+    post 'create', action: :create, as: :create_tube_rack
+  end
```
JamesGlover, line 57: This isn't quite right, especially as we are routing the show etc to limber_tube_racks. However I can't quite work out what the best pattern is, as it doesn't match up with the way thins work for plates etc. However, this is probably best deferred until when we actually have a use.

We could consider removing them for now? But easy either way.

JamesGlover (approved):

PR #908, opened 2021-11-04
Dpl154 cardinal submissions allow source tubes with different studies

### LINKED ISSUE #887: DPL-154- [BUG] Support creating one submission for samples from different studies in Cardinal
**User story**
As a developer I want to be able to create a submission from samples from different studies for Cardinal.

When trying to create the first plate from the vac tubes, if the samples belong to different studies, the submission fails on creation saying 'Study cannot be nil' in Sequencescape. 

Harriet
Eduardo

**Acceptance criteria**
- [ ] After picking VAC tubes at first step of Cardinal, the submission is built on the next plate without problems

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

andrewsparkes:
Closes #887 

Allows tubes with different studies to be arrayed together and submitted.

### DIFF
--- a/app/models/labware_creators/multi_stamp_tubes.rb
+++ b/app/models/labware_creators/multi_stamp_tubes.rb
@@ -78,9 +78,17 @@ module LabwareCreators
       }
     end
 
+    def occupied_wells(wells)
+      wells.each_with_object([]) do |well, occupied_wells|
+        next if well.empty?
+
+        occupied_wells << well
+      end
+    end
+
     def asset_groups(child_plate)
       # split the wells by study id e.g. { '1': [<well1>, <well3>, <well4>], '2': [{<well2>, <well5>}]}
-      study_wells = child_plate.wells.to_a.group_by { |well| well.aliquots.first.study.id }
+      study_wells = occupied_wells(child_plate.wells).to_a.group_by { |well| well.aliquots.first.study.id }
 
       # then build asset groups by study in a hash
       study_wells.transform_values do |wells|
--- a/app/models/labware_creators/multi_stamp_tubes.rb
+++ b/app/models/labware_creators/multi_stamp_tubes.rb
@@ -78,6 +78,19 @@ module LabwareCreators
       }
     end
 
+    def asset_groups(child_plate)
+      # split the wells by study id e.g. { '1': [<well1>, <well3>, <well4>], '2': [{<well2>, <well5>}]}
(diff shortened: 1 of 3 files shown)


app/models/labware_creators/multi_stamp_tubes.rb
```diff
@@ -78,6 +78,27 @@ def request_hash(transfer, child_plate)
       }
     end
 
+    def occupied_wells(wells)
+      wells.each_with_object([]) do |well, occupied_wells|
+        next if well.empty?
+
+        occupied_wells << well
+      end
```
JamesGlover, line 86: This can simply be:
`wells.reject(:empty?)`

Most the useful methods on array actually exist in the Enumerable module. I've actually bookmarked this page, as there's a surprising amount of flexibility there:
https://ruby-doc.org/core-2.7.4/Enumerable.html

app/models/labware_creators/multi_stamp_tubes.rb
```diff
@@ -78,6 +78,27 @@ def request_hash(transfer, child_plate)
...
+    end
+
+    def asset_groups(child_plate)
+      # split the wells by study id e.g. { '1': [<well1>, <well3>, <well4>], '2': [{<well2>, <well5>}]}
+      study_wells = occupied_wells(child_plate.wells).to_a.group_by { |well| well.aliquots.first.study.id }
```
JamesGlover, line 91: You've already have an array here, to `to_a` is redundant

JamesGlover: Also, it may be worth changing the plate lookup to include 'wells.aliquots.study' to improve performance significantly. 

I'd probably suggest `plate_with_custom_includes`. The individual named finders were based on the idea that we could encapsulate a lookup as its own distinct thing, but in practice things seem to diverge more than they share comonalities.

JamesGlover (changes requested):

app/models/labware_creators/multi_stamp_tubes.rb
andrewsparkes: Much more succinct, thx

app/models/labware_creators/multi_stamp_tubes.rb
andrewsparkes: Yep, done

JamesGlover (approved):

PR #909, opened 2021-11-04
Added concern to validate sample manfiest presence

### LINKED ISSUE #906: DPL-158 [BUG] Pools plate creation fails if no sample manifest
**Describe the bug**
If the samples on the plate were not uploaded by sample manifest, the pools plate creation fails with a 500 error in the front end, and this exception:

A NoMethodError occurred in plate_creation#create:

  undefined method `supplier_name' for nil:NilClass

  app/models/labware_creators/cardinal_pools_plate.rb:142:in `block in wells_grouped_by_supplier'

This should either fail with a useful message, or (better I think) it should still work, but without using the supplier information for pooling if it is not available.

This came out of UAT, from Eve and Stephen.

**To Reproduce**
1. Create some LCA Blood Vac tubes without a manifest (e.g. using UAT Actions)
2. Go through the Cardinal pipeline until you reach the button to create the LCA PBMC Pools plate 
3. Click button - see error as described above.

**Expected behaviour**
The plate should be created, without using supplier information in the pooling logic. Or, it should fail gracefully with a useful message.

harrietc52:
Closes #906

### DIFF
--- a/spec/models/labware_creators/cardinal_pools_plate_spec.rb
+++ b/spec/models/labware_creators/cardinal_pools_plate_spec.rb
@@ -25,6 +25,7 @@ RSpec.describe LabwareCreators::CardinalPoolsPlate, cardinal: true do
   end
 
   # TODO: rename throughout to source and dest
+  # SS V2 API Plate
   let(:plate) do
     plate1 = create(:v2_plate, uuid: parent_uuid, well_count: plate_size, aliquots_without_requests: 1)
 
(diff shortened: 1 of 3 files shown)


spec/models/labware_creators/cardinal_pools_plate_spec.rb
```diff
@@ -60,6 +61,22 @@
...
+        expect(subject.errors.messages[:source_plate]).to be_present
+      end
+
+      it 'fails validation when 1 well is missing a sample manifest' do
+        stub_v2_plate(plate, stub_search: false)
+        plate.wells[0].aliquots.first.sample.sample_manifest = nil
+        expect(subject).to_not be_valid
+        expect(subject.errors.messages[:source_plate]).to be_present
+      end
+    end
```
KatyTaylor, line 79: It might be worth an extra test where it has a sample manifest, but that doesn't have the supplier filled out - if it's possible to create a manifest with no supplier? Because that would hit a different bit of your `if well.aliquots.first.sample.sample_manifest&.supplier_name.nil?`

KatyTaylor (approved):

PR #917, opened 2021-11-15
DPL-112 Show tube QC data

### LINKED ISSUE #903: DPL-112 - Display normalisation information on XP tube page [Heron] [TOR-67]
**User story**
In order to inform the user about the amount of material they need to transfer from the LTHR-384 Pool XP tube to the LB Lib Pool Norm the LTHR-384 Pool XP tube page should display:

1) All recorded qc_results, especially the molarity
2) The amount of volume to transfer
3) The amount of buffer to add to the child tube

Parameters are:
1) Target molarity in norm tube is 4nM
2) Target volume in norm tube is 192ul
3) In the event that source molarity is <4nM, pick 192ul from the source
4) In the event that pick volume is <2ul, pick 2ul

Safe Assumptions:
1) We can assume that there is sufficient volume in the source tube (There should be 500ul)

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [ ] Information listed above is displayed
- [ ] Qc_results can be shown for *all* tubes
- [ ] Calculation parameters should be help in config
- [ ] Ideally calculation should be easy to reuse in https://github.com/sanger/limber/issues/799 

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #903 

* Display QC information on all tube pages if it's available.
* Create calculation for transfer volumes if molarity is available and the purpose has other calculation parameters.
* Display transfer volumes and a footnote if the transfer inputs are available.

image


stevieing (approved):
Nice. Do we need to do an end to end run of the tapestation implementation on UAT before we pass it on to ops?

sdjmchattie:
> Nice. Do we need to do an end to end run of the tapestation implementation on UAT before we pass it on to ops?

We definitely should be doing something, but I'm not sure I know all the ins and outs of the Tapestation workflow to know what to test end-to-end.

app/models/presenters/tube_presenter.rb
```diff
@@ -60,5 +61,50 @@ def tubes_and_sources
...
+
+    def qc_summary?
+      labware.receptacle&.qc_results&.to_a.present?
+    end
+
+    def qc_summary
+      labware.receptacle.qc_results.sort_by(&:key).each do |result|
+        yield result.key.titleize, result.unit_value.to_s
+      end
+      nil
```
JamesGlover, line 73: You shouldn't need to return nil here if you use
```
<% presenter.qc_summary do %>
```

app/models/presenters/tube_presenter.rb
```diff
@@ -60,5 +61,50 @@ def tubes_and_sources
...
+      volumes = calculate_pick_volumes(
+        target_molarity: target_molarity,
+        target_volume: target_volume,
+        source_molarity: source_molarity,
+        minimum_pick: minimum_pick
+      )
+
+      yield 'Sample Volume *', "#{volumes[:sample_volume].round} µl"
+      yield 'Buffer Volume *', "#{volumes[:buffer_volume].round} µl"
+      nil
```
JamesGlover, line 107: Similarly here wrt nil

app/models/presenters/tube_presenter.rb
```diff
@@ -60,5 +61,50 @@ def tubes_and_sources
...
+
+    def qc_summary
+      labware.receptacle.qc_results.sort_by(&:key).each do |result|
+        yield result.key.titleize, result.unit_value.to_s
+      end
+      nil
+    end
+
+    def transfer_volumes?
+      [source_molarity, target_molarity, target_volume, minimum_pick].none?(&:nil?)
```
JamesGlover, line 77: Rather than `none?(&nil)` this can just be `.all?` Avoids the double negative

app/models/presenters/tube_presenter.rb
```diff
@@ -60,5 +61,50 @@ def tubes_and_sources
...
+    end
+
+    def source_molarity
+      molarity_qc_result = labware.receptacle.qc_results.detect { |result| result.key == 'molarity' }
```
JamesGlover, line 81: See the `latest_qc` and `latest_molarity` methods on well, which we should probably include on receptacle. Reason being we can have multiple reading taken, and we want to use the most recent reading. This method also double checks the units for us.

JamesGlover (changes requested):
A few suggestions, but the only really important one is making sure we use the most recent QC results.

app/models/presenters/tube_presenter.rb
sdjmchattie: I am using `<% presenter.qc_summary do |key, value| %>`.  I copied the return of `nil` from https://github.com/sanger/limber/blob/develop/app/models/presenters/presenter.rb#L52

Should the `nil` be removed from there as well?  It was added by you on 5th Sep 2019.

sdjmchattie: Turns out when you don't return `nil` the hash generated gets output directly to the UI

sdjmchattie:
One possible improvement that I was thinking of but also didn't do is to extract the `qc_results` relationship and `latest_x` methods to a module that can be included in each of Well and Receptacle.  However I didn't think it was really worth it for a single instance of duplication as it adds complexity with very little gain.  Also, some of the `latest_x` methods simply wouldn't get used on tubes.

app/models/presenters/tube_presenter.rb
JamesGlover: Huh, okay, I thought that was an issue when you were using `<%=` rather than `<%`

I remember having this issue before, and adding the nil stuff as a work-around. I thought switching to `<%` was how I avoided needing to do it, but obviously there's something else at play that I'm forgetting.

JamesGlover: Keep the nil's for now, but I'll see if I can work out what's happening next time I'm in limber, as I'm pretty sure I've been successful in removing them elsewhere in the past.

JamesGlover (approved):

JamesGlover:
> One possible improvement that I was thinking of but also didn't do is to extract the `qc_results` relationship and `latest_x` methods to a module that can be included in each of Well and Receptacle. However I didn't think it was really worth it for a single instance of duplication as it adds complexity with very little gain. Also, some of the `latest_x` methods simply wouldn't get used on tubes.

I think I probably would have gone for the module myself, but its not a strong preference.

PR #920, opened 2021-11-18
Rename file correctly so Rails can find the class

sdjmchattie:
I think this is right.  The missing underscore seems to confuse Rails so that it can't run `constantize` on the string `LabwareCreators::PooledTubesBySubmissionWithPhiX` some of the time.  I don't know why it worked sometimes and others I had to restart the server on the local machine.  Changing the name of this file seems to have solved it, but maybe that's a red herring?


KatyTaylor (approved):
seems to work for me too locally

PR #921, opened 2021-11-22
DPL-074-1 consolidate samples into tubes

andrewsparkes:
Adds fluidx tube barcode file upload and tube creation, and new export for Hamilton csv driver file


config/pipelines/high_throughput_cardinal.wip.yml
```diff
@@ -23,7 +24,7 @@ Cardinal:
 Cardinal Banking:
   filters:
     request_type_key: limber_cardinal_banking
-    library_type: Cardinal Banking
```
KatyTaylor, line 26: It seems to me like this library_type here should match whatever we create as part of the submission. You've removed it as a requirement for going down the pipeline here, but haven't removed it in config/purposes/cardinal.wip.yml on line 128

config/purposes/cardinal.wip.yml
```diff
@@ -26,7 +22,6 @@ LCA Blood Array:
   :stock_plate: true
   :cherrypickable_target: false
   :input_plate: true
```
KatyTaylor, line 24: Should this now be `:input_plate: false` ?

config/purposes/cardinal.wip.yml
```diff
@@ -120,10 +115,8 @@ LCA Custom Pool Norm:
 # LCA Blood Vac tubes are also arrayed into LCA Blood Bank plates (DPL-073)
 LCA Blood Bank:
   :asset_type: plate
-  :stock_plate: true
-  :cherrypickable_target: false
-  :input_plate: true
-  :presenter_class: Presenters::SubmissionPlatePresenter
+  :stock_plate: false
+  :input_plate: false
```
KatyTaylor, line 119: did you get to the bottom of whether it works to have this as:
```
:stock_plate: true
:input_plate: false
```

app/models/labware_creators/multi_stamp_tubes.rb
```diff
@@ -36,17 +36,24 @@ def allow_tube_duplicates?
     private
 
     def create_labware!
```
KatyTaylor, line 38: There's something quite mixed up about this method at the moment. It calls both `create_submission_from_parent_tubes` at the start, and `create_submission_from_child_plate` at the end.
`create_submission_from_parent_tubes` doesn't exist.

KatyTaylor: Ah, it was messed up by this merge: https://github.com/sanger/limber/commit/4d76d3978a667160cdd87e58391c47a087f5cb2c

KatyTaylor: Fixed: https://github.com/sanger/limber/pull/921/commits/f9bf6aa19696f9b9fa523f014327c6636100953d

app/models/labware_creators/multi_stamp_tubes.rb
```diff
@@ -27,29 +27,30 @@ class MultiStampTubes < Base # rubocop:todo Style/Documentation
...
     end
 
     private
 
     def create_labware!
+      create_submission_from_parent_tubes
+
+      # TODO: Hack - change this to poll for status or something
+      puts '*** Waiting for it to build ***'
+      sleep(10)
```
KatyTaylor, line 41: We need to fix this, although perhaps we can get it onto UAT to check the functionality is as expected in parallel

app/models/labware_creators/pooled_tubes_by_sample.rb
```diff
@@ -0,0 +1,261 @@
+# frozen_string_literal: true
+
```
KatyTaylor, line 2: As mentioned on Slack, this could do with a test, and possibly a refactor for some bits (think some bits are overcomplicated, for instance use of 'name_for' and 'name_for_details' methods). But again, we could get it onto UAT and then do this.

KatyTaylor (approved):

PR #927, opened 2021-12-02
DPL-141: Heron verify tube parents

### LINKED ISSUE #799: DPL-141 When scanning the “LTHR-384 Pool XP tube” and the “LB Lib Pool Norm” tube we would like to verify the barcodes on the tubes are from the same parent samples prior to manual transfer of the liquid within [Heron] (M)
**User story**
When scanning the “LTHR-384 Pool XP tube” and the “LB Lib Pool Norm” tube we would like to verify the barcodes on the tubes are from the same parent samples prior to manual transfer of the liquid within.

Currently swaps are occasionally being made during the normalisation step, even though the team are only normalising one pool on a rack at a time.

Potentially this is due to the wrong barcode label being selected to stick to the norm tube.

Naomi / Sarah / Steve

**Acceptance criteria**
- [ ] a simple “scan both tubes and verify for a single pool” page
- [ ] Nice to have: Once tubes verified, display the source molarity, volume to transfer and buffer to add. Do not do this before tubes are verified, to avoid people transferring the wrong volume

**Dependencies**
- #923 

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

### LINKED ISSUE #923: DPL-112-2 Move transfer volumes to norm tubes (C=S, V=5)
**User story**
As Naomi P, I need the transfer volumes to not be visible until the lab worker has verified they have the correct two tubes.

Naomi P
Sara S
Stuart M

**Acceptance criteria**
- [ ] Transfer volumes as shown on LTHR-384 Pool XP tubes are no longer shown.
- [ ] These are instead going to be on another page where you will scan the two tubes first (see linked story in references).

**References**
This story is a prerequisite for #799

sdjmchattie:
Closes #799 and
Closes #923 

* Add a button to the LTHR-384 Pool XP tube page to access a paired tube validation screen.
* Let the user scan tubes into the paired tube validation screen.
* Validate the tubes are indeed paired and are of the right types.
* Show the transfer volumes from the source to the destination tube, only if the tubes are paired.
* Lock the entry of further tubes once the pairing has been made to prevent confusion in scanning.

image


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2011,2012,2013,2014,2015,2016,1017 Genome Research Ltd.
+Copyright (c) 2011,2012,2013,2014,2015,2016,2017 Genome Research Ltd.
```
sdjmchattie, line 1: This licence is repeated in the header of a few of the files, but with the wrong version declared.  We might want to remove those headers elsewhere.

stevieing (approved):
All looks good to me but haven't tested locally.

app/javascript/validate-paired-tubes/components/ValidatePairedTubes.vue
```diff
@@ -0,0 +1,154 @@
...
+        :confirmed-pair="allValid"
+      />
+    </lb-main-content>
+    <lb-sidebar>
+      <b-card
+        header="Validate Tubes"
+        header-tag="h3"
+      >
+        <p class="tv-instructions">
+          Scan the LTHR-384 Pool XP tube and the corresponding LB Lib Pool Norm tube. If they match, the transfer volumes will be shown to the left.
```
JamesGlover, line 20: You've hardcoded the tube names here, but the button will appear whenever tubes have the volume parameters configured. Given how the page is currently accessed, I don't think its possible to make the message dynamic (without passing in additional parameters to the initial request) so maybe just make the message a little less specific.

app/javascript/validate-paired-tubes/components/ValidatePairedTubes.vue
```diff
@@ -0,0 +1,154 @@
...
+        </p>
+        <div class="tube-scan-fields">
+          <lb-labware-scan
+            key="source"
+            ref="sourceScan"
+            :api="devourApi"
+            label="Pool XP"
```
JamesGlover, line 27: Ahh, the hard-coded names here are a little more useful, but source and destination could also be used?

app/javascript/validate-paired-tubes/components/ValidatePairedTubes.vue
```diff
@@ -0,0 +1,154 @@
...
+      // Also the source of transfer volume inputs.
+      destinationTube: { state: 'empty', labware: null }
+    }
+  },
+  computed: {
+    sourceTubeValidators() {
+      return [checkState(['passed']), checkTransferParameters(this.purposeConfigs), checkMolarityResult()]
+    },
+    destinationTubeValidators() {
+      const allowedIdsFromSource = this.sourceTube.labware?.receptacle.downstream_tubes.map(tube => tube.id)
```
JamesGlover, line 115: Ahh, nice, I didn't know operator chaining terminated the evaluation of the rest of the expression. I assumed it followed a similar pattern to ruby where it passed undefined to the right hand side of the expression.

I'm actually having a hard time thinking of a situation where the ruby behaviour would be required.

app/javascript/validate-paired-tubes/components/ValidatePairedTubes.vue
```diff
@@ -0,0 +1,154 @@
...
+      const allowedIds = (allowedIdsFromSource === undefined) ? [] : allowedIdsFromSource
+      return [checkPurpose(['LB Lib Pool Norm']), checkId(allowedIds, 'Does not match the source tube')]
```
JamesGlover, line 117: Oh...

JamesGlover: Actually... do we care? If the main reason is to validate the tubes are related... does it even matter what purpose they are?

app/javascript/validate-paired-tubes/index.js
```diff
@@ -0,0 +1,35 @@
...
+import Sidebar from 'shared/components/Sidebar.vue'
+import axios from 'axios'
+
+Vue.use(BootstrapVue)
+
+Vue.component('lb-main-content', MainContent)
+Vue.component('lb-page', Page)
+Vue.component('lb-sidebar', Sidebar)
+
+document.addEventListener('DOMContentLoaded', () => {
```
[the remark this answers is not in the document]
JamesGlover: Its been on my list for a while!

JamesGlover (changes requested):
Overall looks good, the vast majority of my comments are just personal preference, so feel free to ignore them. (Especially the one about testing)

It is a shame that its a bit tricky to use the state-changer. I think it's worth revisiting at some point in the future, but all this code should be pretty trivial to port across if we go that route. (My suspicion is that there will be value for this in all pipelines with manual tube-tube transfer, which makes our job somewhat easier)

Of my comments, the only thing I will push for a bit more, is the removal of the hard-coded purpose names in the vue file. Rather than pushing them through as request parameters, I think it would be okay to be a bit more general with the terminology. We could probably even loose the bit where we validate an explicit purpose name, as we're pretty much just validating parent-child relationships. (With a some added checks to make sure we have QC data - which is nice)

Another suggestion -> You could add a manual transfer button to the validation page post validation, to avoid the need to go back to the tube page. (Although may be worth pushing down the line due to time constraints)

app/controllers/tubes/validate_paired_tubes_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Handles validating the tubes are in the same heirarchy and then displaying
+# the transfer volume required for manual transfer at a given molarity.
+class Tubes::ValidatePairedTubesController < ApplicationController
+  before_action :check_for_current_user!
+
+  def index; end
+end
```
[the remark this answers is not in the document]
sdjmchattie: There was uncertainty over where it was OK to introduce this and where not.  This workflow allows almost the same functionality and was approved by Sara S

app/javascript/validate-paired-tubes/components/TransferVolumes.spec.js
```diff
@@ -0,0 +1,234 @@
...
+      expect(purposeMinimumPickParameter.mock.calls[0][0]).toBe(mockPurposeConfig)
+    })
+
+    it('calls calculateTransferVolumes with expected arguments', () => {
+      expect(calculateTransferVolumes.mock.calls.length).toBe(1)
+      expect(calculateTransferVolumes.mock.calls[0][0]).toBe(targetMolarity)
+      expect(calculateTransferVolumes.mock.calls[0][1]).toBe(targetVolume)
+      expect(calculateTransferVolumes.mock.calls[0][2]).toBe(sourceMolarity)
+      expect(calculateTransferVolumes.mock.calls[0][3]).toBe(minimumPick)
+    })
```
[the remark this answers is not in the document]
sdjmchattie: That's fair.  I hadn't realised I was comparing different types.  I'll fix it.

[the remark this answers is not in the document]
sdjmchattie: All but the molarity were being returned as numbers already.  I've modified the most recent molarity to return a float instead of change these tests.

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2011,2012,2013,2014,2015,2016,2017 Genome Research Ltd.
+Copyright (c) 2011-2021 Genome Research Ltd.
```
JamesGlover, line 1: Last time I saw it Sanger policy cautioned against date ranges unless they were impossible to avoid (Ie. you didn't know if there had been commits on certain years)

app/javascript/shared/tubeTransferVolumes.js
```diff
@@ -1,22 +1,24 @@
...
-    ?.value
+  // Get the QC results,
+  // find those for molarity entries in nM,
+  // inverse sort by the created_at timestamp,
+  // inverse sort by the id,
+  // take the first item and parse the value to a float.
+  const qcResults = tube?.receptacle?.qc_results
+  const molarityEntries =  qcResults?.filter(result => result.key === 'molarity' && result.units === 'nM')
+  const sortedByCreatedAt = molarityEntries?.sort((resultA, resultB) => -1 * ('' + resultA.created_at).localeCompare(resultB.created_at))
+  const sortedByIds = sortedByCreatedAt?.sort((resultA, resultB) => parseInt(resultA.id) > parseInt(resultB.id) ? -1 : 1)
```
JamesGlover, line 19: Was about to point out that the ECMA script standard didn't specify a stable sort, but it looks like this has changed. I guess it may mean that the won't work on some ancient browsers, but given id will show the correct behaviour in almost all cases, I don't think its worth making things more complicated here. (Especially as I'd be incredibly surprised to find an implementation that wasn't stable)

JamesGlover (approved):

PR #934, opened 2021-12-09
Dockerise Limber

### LINKED ISSUE #933: DPL-nnn Dockerized option for Limber
**User story**
As a PSD dev I want to have a Dockerized setup option so I can run Limber even when I have problems with databases and libraries in my local installation. This is a first step in order to solve the story of running integration suite automatically

Eduardo Martin

**Acceptance criteria**
- [ ] Provide a README with the set of commands I can run to setup a Docker version of Limber that I can use optionally instead of the native installation
- [ ] Provide me a way of starting all services required automatically without me needing to create any specific config setup
- [ ] As long as I have Docker correctly installed and I have an internet connection I should be able to run this version of Limber even if I have problems in my local setup (no dependencies with local installed databases (mysql) or libraries (openssl, etc)

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #933 

* Add Dockerfile and docker-compose files.
* Set up so that `docker build .` only includes production dependencies.
* Ensure local builds automatically generate the config on launch.
* Add a health endpoint for healthcheck to function.
* Ignore unnecessary files to reduce docker image size.


Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# Install bundler based dependencies
+COPY Gemfile .
+COPY Gemfile.lock .
+RUN bundle config set without "${bundlerWithout}"
+RUN bundle install
+
+# Install yarn based dependencies
+COPY package.json .
+COPY yarn.lock .
+RUN rm -rf node_modules && yarn install --frozen-lockfile ${yarnFlags}
```
emrojo, line 37: Did you have to add the removal for any reason? At this moment of the build the folder does not have the node_modules folder yet

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# "items (files, directories) that do not require ADD’s tar auto-extraction capability, you should always use COPY."
+#   https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
+COPY . .
+
+# Set up ENTRYPOINT script as a system binary
+COPY entrypoint.sh /usr/bin/
+RUN chmod +x /usr/bin/entrypoint.sh
```
emrojo, line 45: /usr/bin is more for system packages that belong to the official OS distribution; 3rd party apps would use /usr/local or /opt folder (https://unix.stackexchange.com/questions/8656/usr-bin-vs-usr-local-bin-on-linux)

entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Note that this file is for the exclusive use of the Dockerfile as an ENTRYPOINT
+set -e
+
+# Remove a potentially pre-existing server.pid for Rails.
+rm -f /code/tmp/pids/server.pid
```
emrojo, line 7: I had same problem in SS; why we have a problem with this file when we run Limber in Docker but not in native? What is different?

emrojo (commented):
All changes here good, but I see missing a way of starting the required services for Limber and the README + commands to start and work with this.

sdjmchattie:
> All changes here good, but I see missing a way of starting the required services for Limber and the README + commands to start and work with this.

Yes good point.  It was overlooked but I'll add it in now.

Dockerfile
sdjmchattie: It was move of a "just in case".  It is unnecessary according to the current Dockerfile but it also doesn't do any harm.

Dockerfile
sdjmchattie: I would be inclined to agree but this was taken from the official Docker documentation: https://docs.docker.com/samples/rails/

entrypoint.sh
sdjmchattie: I didn't have a problem.  I just followed the advice of the official Docker documentation: https://docs.docker.com/samples/rails/

Dockerfile
emrojo: Ok

entrypoint.sh
emrojo: Looks good, the problem with this file was the reason why I had to start using exec in the entrypoint, so a docker-compose down would stop all services and remove this file; but this didn't work if the user kills with control-c. With this we cover all.

emrojo (approved):

PR #937, opened 2021-12-10
Docker tweak

### LINKED ISSUE #933: DPL-nnn Dockerized option for Limber

sdjmchattie:
Closes #933 

* JavaScript needs a different localhost URL for the SequenceScape v2 API than the rest of the Rails app.  While requests from the Rails app come from the Docker container Limber is running in, JavaScript requests come from the browser and that doesn't know what the special domain means for Docker containers calling back to the host machine.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
+# Localhost might need to be rewritten if running inside an environment/container.
+# In which case, the value in the LOCALHOST environment variable will be substituted.
+def rewrite_localhost(url)
+  url.gsub(%r{((?<=http://)|(?<=https://))localhost}i, ENV.fetch('LOCALHOST', 'localhost'))
+end
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
-
-  localhost_env = ENV.fetch('LOCALHOST', 'localhost')
 
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
@@ -67,12 +71,12 @@ Rails.application.configure do
   config.api.v1                                  = ActiveSupport::OrderedOptions.new
   config.api.v1.connection_options               = ActiveSupport::OrderedOptions.new
   config.api.v1.connection_options.namespace     = 'Limber'
-  config.api.v1.connection_options.url           = ENV.fetch('API_URL', "http://#{localhost_env}:3000/api/1/")
+  config.api.v1.connection_options.url           = rewrite_localhost(ENV.fetch('API_URL', 'http://localhost:3000/api/1/'))
   config.api.v1.connection_options.authorisation = ENV.fetch('API_KEY', 'development')
 
   config.api.v2                                  = ActiveSupport::OrderedOptions.new
   config.api.v2.connection_options               = ActiveSupport::OrderedOptions.new
-  config.api.v2.connection_options.url           = ENV.fetch('API2_URL', "http://#{localhost_env}:3000/api/v2")
+  config.api.v2.connection_options.url           = rewrite_localhost(ENV.fetch('API2_URL', 'http://localhost:3000/api/v2'))
   config.api.v2.connection_options.js_url        = ENV.fetch('API2_URL', 'http://localhost:3000/api/v2')
 
   config.qc_submission_name = 'MiSeq for QC'
@@ -82,6 +86,6 @@ Rails.application.configure do
   config.request_options = {
     'read_length' => 11
   }
-  config.pmb_uri = ENV.fetch('PMB_URI', "http://#{localhost_env}:3002/v1/")
+  config.pmb_uri = ENV.fetch('PMB_URI', rewrite_localhost('http://localhost:3002/v1/'))
   config.sprint_uri = 'http://sprint.psd.sanger.ac.uk/graphql'
 end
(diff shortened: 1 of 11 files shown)


config/environments/development.rb
```diff
@@ -67,12 +67,13 @@
...
   config.api.v1.connection_options.namespace     = 'Limber'
-  config.api.v1.connection_options.url           = ENV.fetch('API_URL', "http://#{localhost}:3000/api/1/")
+  config.api.v1.connection_options.url           = ENV.fetch('API_URL', "http://#{localhost_env}:3000/api/1/")
   config.api.v1.connection_options.authorisation = ENV.fetch('API_KEY', 'development')
 
   config.api.v2                                  = ActiveSupport::OrderedOptions.new
   config.api.v2.connection_options               = ActiveSupport::OrderedOptions.new
-  config.api.v2.connection_options.url           = ENV.fetch('API2_URL', "http://#{localhost}:3000/api/v2")
+  config.api.v2.connection_options.url           = ENV.fetch('API2_URL', "http://#{localhost_env}:3000/api/v2")
+  config.api.v2.connection_options.js_url        = ENV.fetch('API2_URL', 'http://localhost:3000/api/v2')
```
emrojo, line 76: There something with these using the same API2_URL env that I dont see very clear. Is it a mistake or is it because of any particular reason?

sdjmchattie: No mistake.  If the APIv2 URL were for a different server -- a proper domain name.  You wouldn't want server side requests going to a different instance of SequenceScape than the browser requests.  This problem only exists when you need to talk to the docker host from the server and to carry on using `localhost` in the browser.

sdjmchattie: What wouldn't work, admittedly, is if you specified `http://localhost:1234/api/v2` in the `API2_URL` environment variable.  Maybe the substitution needs to be made to the fetched variable, whether it's default or specified.  I'll update for that.

config/environments/development.rb
```diff
@@ -67,12 +71,13 @@
...
   config.api.v1.connection_options               = ActiveSupport::OrderedOptions.new
   config.api.v1.connection_options.namespace     = 'Limber'
-  config.api.v1.connection_options.url           = ENV.fetch('API_URL', "http://#{localhost}:3000/api/1/")
+  config.api.v1.connection_options.url           = rewrite_localhost(ENV.fetch('API_URL', 'http://localhost:3000/api/1/'))
   config.api.v1.connection_options.authorisation = ENV.fetch('API_KEY', 'development')
 
   config.api.v2                                  = ActiveSupport::OrderedOptions.new
   config.api.v2.connection_options               = ActiveSupport::OrderedOptions.new
-  config.api.v2.connection_options.url           = ENV.fetch('API2_URL', "http://#{localhost}:3000/api/v2")
+  config.api.v2.connection_options.url           = rewrite_localhost(ENV.fetch('API2_URL', 'http://localhost:3000/api/v2'))
```
emrojo, line 79: I think we are trying to reuse API2_URL but would be better to have API_2_JS as another env variable.
If I wanted to change the settings in my local to do something different from the default I would try to redefine all environment variables needed with the values needed, but this rewrite could modify the user settings.
(I'm talking only about the rewrite in line 79)
```suggestion
  config.api.v2.connection_options.url           = ENV.fetch('API2_JS_URL', rewrite_localhost('http://localhost:3000/api/v2'))
```

emrojo (approved):
Just a comment if you think is a good suggestion

config/environments/development.rb
sdjmchattie: If you haven't also declared the environment variable `LOCALHOST` then no rewrite will occur.  Docker Compose sets this environment variable so that Docker will use the correct localhost for the correct places in the code.  I think adding another environment variable such as `API_2_JS` will only make explaining this behaviour even more complicated for developers.  The only reason for the rewrite is so that the endpoint, which should be the same for both server-side and client-side requests, has the correct URL associated with it in each location.  I can't think of a scenario when the developer might want to have server-side talk to one Sequencescape instance and client-side talk to another instance.  If you set `API2_URL` environment variable to:

- https://localhost:9999/api/v2 then it would be changed to https://host.docker.internal:9999/api/v2 for server side requests from inside the container and left as `https://localhost:9999/api/v2` for client side requests in the browser.
- https://sequencescape.com/api/v2 then it would not be modified at all for either server-side or client-side requests.

sdjmchattie: If you haven't also declared the environment variable `LOCALHOST` then no rewrite will occur.  Docker Compose sets this environment variable so that Docker will use the correct localhost for the correct places in the code.  I think adding another environment variable such as `API_2_JS` will only make explaining this behaviour even more complicated for developers.  The only reason for the rewrite is so that the endpoint, which should be the same for both server-side and client-side requests, has the correct URL associated with it in each location.  I can't think of a scenario when the developer might want to have server-side talk to one Sequencescape instance and client-side talk to another instance.  If you set `API2_URL` environment variable to:

- https://localhost:9999/api/v2 then it would be changed to https://host.docker.internal:9999/api/v2 for server side requests from inside the container and left as `https://localhost:9999/api/v2` for client side requests in the browser.
- https://sequencescape.com/api/v2 then it would not be modified at all for either server-side or client-side requests.

PR #940, opened 2021-12-14
Display 2 DP instead of 1 for transfer volumes

### LINKED ISSUE #799: DPL-141 When scanning the “LTHR-384 Pool XP tube” and the “LB Lib Pool Norm” tube we would like to verify the barcodes on the tubes are from the same parent samples prior to manual transfer of the liquid within [Heron] (M)

sdjmchattie:
Closes #799

### DIFF
--- a/app/javascript/validate-paired-tubes/components/TransferVolumes.spec.js
+++ b/app/javascript/validate-paired-tubes/components/TransferVolumes.spec.js
@@ -143,7 +143,7 @@ describe('TransferVolumes', () => {
     })
 
     it('returns the expected sample volume', () => {
-      expect(wrapper.vm.sampleVolume).toBe('150.1')
+      expect(wrapper.vm.sampleVolume).toBe('150.12')
     })
   })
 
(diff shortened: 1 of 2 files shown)


stevieing (approved):
All good for me. Looks like you have some failing tests.

PR #962, opened 2022-01-04
Update primer panel used in submission [Do not release until 10th Jan]

### LINKED ISSUE #939: DPL-194 - As a Senior Staff Scientist (Naomi P) I would like to replace the Artic v4 primer panel with a new version v4.1 so that we can address amplicon dropouts caused by the Omicron variant (C=S, V=5)
**User story**
As a Senior Staff Scientist (Naomi P) I would like to replace the Artic v4 primer panel with a new version v4.1 so that we can address amplicon dropouts caused by the Omicron variant

Naomi P
Sara S
Lesley S

**Acceptance criteria**
- [ ] The primer panel name for Heron plates in Limber should be: 'nCoV-2019/V4.1'
- [ ] A release date & time is agreed with the lab (Sara S) as to when they switch over to the new primers

**Dependencies**
None

References
Similar to #774 

**Additional context**
The change over between primer panels needs to be co-ordinated with the Heron lab (Sara S). 
The use of a runtime feature flag might make this easier.

JamesGlover:
Fixes #939

Omicron variant is resulting in dropouts, so primer panel changed
to reflect this.

Operational switch expected 10th January.

*
*
* ...

### DIFF
--- a/config/purposes/heron_lthr_entry.yml
+++ b/config/purposes/heron_lthr_entry.yml
@@ -17,7 +17,7 @@ LTHR Cherrypick:
         read_length: 150
         fragment_size_required_from: '50'
         fragment_size_required_to: '800'
-        primer_panel_name: nCoV-2019/V4
+        primer_panel_name: nCoV-2019/V4.1alt
       allowed_extra_barcodes: false
     LTHR 384 - NovaSeq:
       template_name: 'Limber - Heron LTHR - Automated'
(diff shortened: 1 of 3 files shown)


andrewsparkes (approved):
How do you toggle this change in Prod? Do you just deploy it at the right time?

JamesGlover:
Unfortunately yes. It probably is worth thinking about more streamlined solutions in future:

Feature flagging in limber: Limber doesn't have any persistence of its own at the moment, so this would still mean a re-deployment with some environmental flags. In turn, this increases the complexity of the configuration.

Move the config to SS: I *think* we might be able to move the request options into Sequencescape, and thus hold the config in the database. This could make the switch much easier to handle, and could even be exposed to operations. I might do some quick experiments to see if this is as simple as I'm thinking.

PR #977, opened 2022-01-18
Configure whether to enable .wip.yml files

KatyTaylor:
Make it possible to configure whether to deploy 'work in progress' .wip.yml files, to make it easier to deploy to UAT with them on, for example

### DIFF
--- a/lib/config_loader/base.rb
+++ b/lib/config_loader/base.rb
@@ -22,6 +22,7 @@ module ConfigLoader
     #
     def initialize(files: nil, directory: default_path)
       path = directory.is_a?(Pathname) ? directory : Pathname.new(directory)
+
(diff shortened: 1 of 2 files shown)


lib/config_loader/base.rb
```diff
@@ -22,12 +22,21 @@ class Base
...
     end
 
     private
 
+    def should_include_file?(files, child)
+      yaml?(child) &&
+        in_list?(files, child) &&
+        (
+          !work_in_progress?(child) ||
+          (work_in_progress?(child) && Limber::Application.config.deploy_wip_pipelines)
```
emrojo, line 36:
```suggestion
          (work_in_progress?(child) && Limber::Application.config.fetch('deploy_wip_pipelines', false))
```

emrojo (approved):

PR #978, opened 2022-01-20
DPL-243 ss gem https

harrietc52:
**Dependencies**
https://github.com/sanger/deployment/issues/164
https://github.com/sanger/sequencescape-client-api/pull/97

Ruby version ruby 2.7.5, so update SS client gem to 2.x


.github/workflows/automated_release.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - dpl-243-ss-gem-https
```
KatyTaylor, line 8: This can be removed now, I guess?

KatyTaylor (approved):

PR #990, opened 2022-01-31
Fix exception when tube not known to Limber

KatyTaylor:
This line was causing the following exception:
- undefined method 'transfer_parameters' for {}:Hash
- when loading a tube page where the tube type wasn't known to Limber i.e. wasn't in a pipeline

I ran the Heron integration suite tests (heron_lthr_pipeline_spec), which passed

I noticed the exception locally, but also in frequent emails from prod.

### DIFF
--- a/app/models/presenters/tube_presenter.rb
+++ b/app/models/presenters/tube_presenter.rb
@@ -72,7 +72,7 @@ module Presenters
     end
 
     def transfer_volumes?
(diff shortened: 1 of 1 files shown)


sdjmchattie:
Nice catch.  Thanks.

sdjmchattie (approved):

PR #997, opened 2022-02-03
Revert babel core.

stevieing:
Closes #

*
*
* ...


config/webpacker.yml
```diff
@@ -4,7 +4,7 @@ default: &default
   source_path: app/javascript
   source_entry_path: packs
   public_root_path: public
-  public_output_path: assets
```
JamesGlover, line 7: I'm curious about this change. Isn't that where the compiled assets should be placed. Howe does nginx see the compiled assets to deliver them if they aren't in public.

JamesGlover: Wait, no that's just assets... still under public.

JamesGlover (approved):

PR #1003, opened 2022-02-14
DPL-256 & DPL-207 cardinal hamilton pooling file tweaks

### LINKED ISSUE #945: DPL-nnn Cardinal driver file minimum volume
**User story**
As an R&D scientist developing automation scripts for Cardinal I need to set a floor for the robot pooling driver file (DPL-068) in order to guarantee there is enough volume for the chromium connect and cell counting processes, with enough excess to allow for compounding pipetting inaccuracies, regardless of the number of donors in the pool.

Scott T

**Acceptance criteria**
The floor should be set to 25µl initially, but may be changed at a later date.

### LINKED ISSUE #986: DPL-nnn sample pooling robot driver file - resuspension volume (C=S, V=4)
**User story**
As a member of the Cardinal team, I would like the resuspension volume to be set to 2.2ul per donor, rather than 3.125, to ensure there are enough cells for the Chromium Connect to work on. 

Stephen W, Scott T, Lesley

**Acceptance criteria**
- [ ] Hamilton driver file for pooling in Cardinal pipeline uses 2.2ul per donor in the calculation, rather than 3.125ul (see additional context for pointer to code).

**Dependencies**
It's definitely worth doing this story at the same time as https://github.com/sanger/limber/issues/945, because they are changing the same bit of code

**Additional context**
Relevant code is here - limber/app/views/exports/hamilton_pooling_plate_pbmc.csv.erb

Until this story is done, the Cardinal team will be manually editing the values in the driver file, introducing risk of human error.

This value may need to be tweaked in future as more samples get put through the pipeline and we get a better idea of the number of cells needed.

KatyTaylor:
Closes #986 
Closes #945

### DIFF
--- a/app/views/exports/hamilton_pooling_plate_pbmc.csv.erb
+++ b/app/views/exports/hamilton_pooling_plate_pbmc.csv.erb
@@ -31,10 +31,11 @@
 %>
 <%
   # Calculate the resuspension volumes from the pooled samples count
+  resuspension_volume_per_sample = 2.2
   samples_by_dest_well = transfer_request_data.group_by { |tr| tr[3] }
   rows_array = transfer_request_data.map do |data|
     samples_in_pool = samples_by_dest_well[data[3]].count
-    data + [samples_in_pool * 3.125]
+    data + [samples_in_pool * resuspension_volume_per_sample]
   end
 %>
 <%= rows_array.sort_by{ |a| WellHelpers.well_coordinate(a[1]) }.map(&:to_csv).join %>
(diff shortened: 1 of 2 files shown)


sdjmchattie (approved):

app/views/exports/hamilton_pooling_plate_pbmc.csv.erb
```diff
@@ -31,10 +31,14 @@
 %>
 <%
   # Calculate the resuspension volumes from the pooled samples count
+  resuspension_volume_per_sample = 2.2
```
andrewsparkes, line 34: we tend to put numbers like these in the purpose config rather than hardcode them like this, easier to find if you want to change later

andrewsparkes: on the other hand it's a pretty specific csv file

andrewsparkes (approved):

PR #1041, opened 2022-03-29
Dpl331 Bug fix for failing rtq plate with failed wells

### LINKED ISSUE #1038: DPL-331 [BUG] 500 error when failing RT-Q plate with failed wells
When the Heron team attempt to fail an RT-Q plate that has failed wells Limber throws a 500 error
**RT Ticket Number**
Most recent RT [746613](https://rt.sanger.ac.uk/Ticket/Display.html?id=746613)

andrewsparkes:
Closes #1038


app/models/state_changers.rb
```diff
@@ -38,7 +38,13 @@ def move_to!(state, reason = nil, customer_accepts_responsibility = false)
     def contents_for(target_state)
       return nil unless FILTER_FAILS_ON.include?(target_state)
 
-      labware.wells.reject { |w| w.state == 'failed' }.map(&:location)
+      # determine list of well locations requiring the state change
+      well_locations_filtered = labware.wells.reject { |w| w.state == 'failed' }.map(&:location)
+
+      # if no wells are in failed state then no need to send the contents subset
+      return nil if well_locations_filtered.length == labware.size
```
JamesGlover, line 45: I'd use `labware.wells.length` rather than `labware.size` as we sometimes have plates where not all the wells actually exist. (Mostly out of cherrypicking)

JamesGlover (changes requested):
Just a minor one

app/models/state_changers.rb
andrewsparkes: fixed

JamesGlover (approved):

PR #1056, opened 2022-04-13
Dependency updates

JamesGlover:
- Update to Rails 7
- Import upstream changes for json-client-api: Unfortunately we still need our fork, as the has_many relationship on the standard gem is still an array, rather than a query builder. Not entirely sure why, the PR to fix it was closed, and the issue was temporarily marked as 'resolved'


Gemfile
```diff
@@ -20,7 +20,7 @@ group :default do
   gem 'hashie'
 
   # Communications with JSON APIs, allows us to begin migration to the new Sequencescape API
-  gem 'json_api_client', github: 'sanger/json_api_client', branch: 'merge_upstream'
+  gem 'json_api_client', github: 'sanger/json_api_client', branch: 'v1.21.0a'
```
JamesGlover, line 23: See https://github.com/sanger/json_api_client/pull/8 for the changes pulled in. Its mostly just improved error handling, and a couple of create/update methods. Importantly it also relaxes the Faraday constraint,  although doesn't move us on to 2.0.

I do wish there were better options available.

Gemfile
```diff
@@ -32,6 +32,7 @@ group :default do
 
   gem 'puma'
   gem 'sanger_barcode_format', github: 'sanger/sanger_barcode_format', branch: 'development'
+  gem 'sprockets-rails'
```
JamesGlover, line 35: No longer a default dependency.

Moving limber on to vite may be a touch complicated by invline_svg which tries to hook into the Rails asset pipeline. The latest versions are configurable though, so I'm sure we'll manage.

bin/rails
```diff
@@ -1,6 +1,4 @@
 #!/usr/bin/env ruby
```
JamesGlover, line 1: Big chunk of app:update changes

stevieing (approved):

PR #1079, opened 2022-05-19
DPL265 - Fix for not clearing the bad bed flags

andrewsparkes:
Fix for bug identified after Training session by Iraad

### DIFF
--- a/app/assets/javascripts/bed_verification.js
+++ b/app/assets/javascripts/bed_verification.js
@@ -80,9 +80,11 @@
         // We don't have any content
         $('#loadingModal').fadeOut(100)
       } else if (response.valid) {
+        // Clear all bed error flags when valid
         clearFlagFromBeds(response.beds)
         pass()
       } else {
+        // Set bed flags according to which are currently valid and which are not
         flagBeds(response.beds, response.message)
         fail()
       }
@@ -94,6 +96,8 @@
         if (!this) {
           $('#bed_list li[data-bed="' + bed_id + '"]').addClass('bad_bed list-group-item-danger')
           bad_beds.push(bed_id)
+        } else {
+          clearFlagFromBed(bed_id)
         }
       })
       SCAPE.message('There were problems: ' + message, 'danger')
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/bed_verification.js
```diff
@@ -94,6 +96,8 @@
         if (!this) {
           $('#bed_list li[data-bed="' + bed_id + '"]').addClass('bad_bed list-group-item-danger')
           bad_beds.push(bed_id)
+        } else {
+          clearFlagFromBed(bed_id)
```
emrojo, line 100: This is a bit complex, What is 'this' for? From jQuery help, they say 'this' is the element, however we are iterating a normal array, not DOM, so I assume 'this' is the element, but then it will never be undefined, son line 97 would never reach?

emrojo: The method name is flagBeds, but it is removing the flag from a bed, and only adding it if this is not defined, but I dont see when that happens.

emrojo: I think this method is never doing the if (!this) part of the condition. Example: 

```$.each([1,2,3], function(a) { console.log("this => "+this)})```

**Review by emrojo — CHANGES_REQUESTED** [review]
Something in the logic I may not understand

File: app/assets/javascripts/bed_verification.js (hunk 1 above)
Comment (andrewsparkes) [reply app/assets/javascripts/bed_verification.js:100]: It does reach the !this statement and set the flag to error so it does something, but I'm not a JS expert so had assumed it was short for if bed not valid

Comment (emrojo) [reply app/assets/javascripts/bed_verification.js:100]: Thanks for checking it with me, I didnt know this behaviour of $.each when using hashes. So when: 
```
$.each({a: 1, b: 3}, function(bed_id) { console.log(this) })
```
The output is 1 and 3 (the value of the element)

Comment (emrojo) [reply app/assets/javascripts/bed_verification.js:100]: And in this case it seems the hash is ```{'bed_1': true, 'bed_2': false)```, so here ```this``` is a boolean

Comment (andrewsparkes) [reply app/assets/javascripts/bed_verification.js]: Fixed

File: app/assets/javascripts/bed_verification.js [hunk 2]
```diff
@@ -102,11 +109,15 @@
     var clearFlagFromBeds = function (beds) {
       $.each(beds, function (bed_id) {
         if (this) {
```
emrojo, line 111: Would it be ok to change it also here?

emrojo (approved):

app/assets/javascripts/bed_verification.js
andrewsparkes: Yes, done

PR #1094, opened 2022-06-07
GPL-403 Fix to allow both arrays and single strings for error message parameter

### LINKED ISSUE #1088: DPL-403 [BUG] Limber error handling - incorrect handling of string error message
**Describe the bug**
Both creation_controller and tube_controller error handling is failing with 'no implicit conversion of String into Array'. It seems sometimes we return an array of messages and other times a single String, and the code is not handling that. User sees an unhelpful 500 error.

**RT Ticket Number**
Example triggered in 752103

**To Reproduce**
1. In above RT in Limber go to plate DN946361K
2. In other actions attempt to create the LB Lib PrePool plate with 6 pools of identical wells
3. 500 error is displayed (core error is it can't handle creating those identical pools (fix to be done in another BUG story), then additionally this error handling fails)

**Expected behaviour**
Display useful error message to the user in the Flash messages on the Limber screen.

**Screenshots**
Not useful.

**Additional context**
Check for numerous occurrences in Kibana for 'no implicit conversion of String into Array'.
See creation_controller line 102

andrewsparkes:
Closes #1088

### DIFF
--- a/app/controllers/creation_controller.rb
+++ b/app/controllers/creation_controller.rb
@@ -98,8 +98,9 @@ class CreationController < ApplicationController
     api_errors_hash.key?('general') ? api_errors_hash['general'] : [api_message]
   end
 
-  def redirect_back_after_error(prefix_message, error_messages)
-    flash_messages = [prefix_message] + error_messages
+  # Use of splat on parameter to cope with both arrays of error messages and single strings
+  def redirect_back_after_error(prefix_message, *error_messages)
+    flash_messages = [prefix_message] + error_messages.flatten
     respond_to do |format|
       format.html do
         redirect_back(fallback_location: url_for(@labware_creator.parent), alert: truncate_flash(flash_messages))
--- a/app/controllers/creation_controller.rb
+++ b/app/controllers/creation_controller.rb
@@ -33,7 +33,7 @@ class CreationController < ApplicationController
     Rails.logger.error(exception.message)
     exception.backtrace.map(&Rails.logger.method(:error)) # rubocop:todo Performance/MethodObjectAsBlock
 
-    redirect_back_after_error('Cannot create the next piece of labware:', *exception.resource.errors.full_messages)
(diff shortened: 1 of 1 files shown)


app/controllers/creation_controller.rb
```diff
@@ -98,8 +98,9 @@ def extract_error_messages_from_api_exception(api_message)
     api_errors_hash.key?('general') ? api_errors_hash['general'] : [api_message]
   end
 
-  def redirect_back_after_error(prefix_message, error_messages)
-    flash_messages = [prefix_message] + error_messages
+  # Use of splat on parameter to cope with both arrays of error messages and single strings
+  def redirect_back_after_error(prefix_message, *error_messages)
+    flash_messages = [prefix_message] + error_messages.flatten
```
JamesGlover, line 103: I think the string is originating from line 36, where we're splatting out the array. If this was hit with an array of multiple errors, it would have been throwing an unknown number of arguments error, but as it was we had one item in the array, so it became a single string.

I'm tempted to suggest:
1) Remove the * from line 36
2) Keep this as a single parameter

This would fix the issue, although would leave is vulnerable if the SS API ever returned a string under 'general', which *shouldn't* happen, but you never know.

If you wanted to proof it agains this, or strings from other resources you could do:

```ruby
flash_messages = [prefix_message] + Array(error_messages)
```

Because:

```ruby
irb(main):001:0> Array('a')
=> ["a"]
irb(main):002:0> Array(['a','b'])
=> ["a", "b"]
```

This discomfort mostly stems from this approach leaving two potential ways to pass in an array of errors:

```ruby
redirect_back_after_error('One way', ['is like this', 'with an array as a single argument'])
redirect_back_after_error('The other way', *['is like this', 'with an array as a single argument'])
# Which is equivalent to
redirect_back_after_error('The other way', 'is like this', 'with an array as a single argument')
```

PR #1164, opened 2022-09-12
Dpl376 duplex tnano seq merged file

### LINKED ISSUE #1074: DPL-376 - Limber - Create merged summary file on PCR XP at end of Duplex Seq and Targeted Nanoseq pipelines
**User story**
As a Team Lead / SSR, we require the ability to merge the named sets of data and be able to relate all data to the original LTN Cherry Pick plate well position within one file, as sample positions change during the course of the process

Named sets:

- [Input amount file - Example_File_A.csv](https://github.com/sanger/limber/files/8729180/Example_File_A.csv)
- [Hamilton driver file - Example_File_B.csv](https://github.com/sanger/limber/files/8729181/Example_File_B.csv)
- [PCR XP concentration file - Example_File_C.csv](https://github.com/sanger/limber/files/8729183/Example_File_C.csv)

The merged file should be populated , presented and downloadable as this set format as regards column order [TNano input order for template for SSR merging.xlsx](https://github.com/sanger/limber/files/8729189/TNano.input.order.for.template.for.SSR.merging.xlsx), in order for the lab team to feed the data into an internal template already present.
 
This information is already present within Limber as these items are uploaded at various stages of process.

Sara W

**Acceptance criteria**
- [x] Download of the merged file at the PCR XP plate step in the Limber Duplex Seq and Targeted Nanoseq pipelines
- [x] The merged file is populated with the correct data as detailed in the example file and comments below
- [x] The merged file is ordered according to the original sample well within the LDS/LTN Cherry pick plate purpose
- [x] The merged file clearly shows the sample relocation which happened at the AL Lib Dil plate step and continues through to the PCR XP plate step
- [ ] Data in the merged file is accepted into the internal Excel template file without issue.

**Dependencies**
- #1078 
- #1077 

References
- sanger/<repo>#<issue_no.>](https://github.com/sanger/limber/issues/1077)

**Additional context**
See comments below

### LINKED ISSUE #1077: DPL-378 - Limber - Modify Duplex Seq and Targeted Nanoseq pipelines to allow for selective progression of samples beyond the AL Lib Dil step
**User story**
As a TeamLead/Laboratory Scientist, we require the ability to filter out some of the samples on the plate by putting zero in the sample volume field at upload of the customer file within Limber when needing to create the LTN & LDS AL Lib Dil plate
(‘Sample volume’ , [Example_File_B.csv](https://github.com/sanger/limber/files/8729216/Example_File_B.csv))

Sara W

**Acceptance criteria**
- [x] Allow upload of customer file on creation of the AL Lib Dil plate with a zero value within the sample field
- [x] Samples with a zero sample_volume should not be included in the AL Lib Dil plate (be filtered out)
- [x] Multiple AL Lib Dil plates may be created from the same AL Lib plate with different selections of wells
- [x] The customer file needs to include at least one additional field 'Hyb Panel' (bait library) (and others may be required see story https://github.com/sanger/limber/issues/1074). The file may need to be converted to Excel to give a Hyb panel dropdown with specific options for the customer (this will increase the cost of the story as it's currently built as a csv).
- [x] Existing charge and pass step needs to be moved earlier to the AL Lib plate.
- [x] A new automated submission with a new submission template and request type should be generated on creation of the AL Lib Dil plate, for the steps between AL Lib Dil through to Lib PCR XP, with a new charge and pass step on the PCR XP plate.

**Dependencies**
- #1078 

**Additional context**
See comments below.

andrewsparkes:
Closes #1074
Closes #1077 

Modification of the duplex seq (and targeted nanoseq) pipelines to allow selection of subset of samples to go forward from Al Lib plate based on what's set in the uploaded customer file, using a new automated submission.
Also a refactor to store parameters needed later in the pipeline on the requests generated in the new automated submission, instead of in the receptacles.


app/models/labware_creators/pcr_cycles_binned_plate/csv_file/row.rb
```diff
@@ -66,33 +83,45 @@ class PcrCyclesBinnedPlate::CsvFile::Row # rubocop:todo Metrics/ClassLength
...
 
-      # initialize supplied fields
+      initialize_sanger_supplied_fields
+      initialize_customer_supplied_fields
+
+      # initialize flag to indicate that sample should not be transferred
+      @do_not_transfer_sample = false
+    end
+
+    # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
sdjmchattie, line 102: Could these issues be resolved rather than suppressed?  What's been added doesn't look like it adds very much if any of these things.

app/models/presenters/plate_presenter.rb
```diff
@@ -151,6 +151,10 @@ def qc_thresholds
 
     private
 
+    def skip_validation_for_single_pcr_cycle_for_all_wells?
+      false
```
sdjmchattie, line 155: This looks like it's for future use or perhaps unimplemented?  Should it be removed or have a comment added to it?

sdjmchattie (approved):
There's a lot here to check so I haven't gone through with a fine tooth and comb, but hopefully someone with more experience in Limber can comment.

app/models/labware_creators/pcr_cycles_binned_plate/csv_file/row.rb
andrewsparkes: I often think rubocop is way too fussy. This class represents a row in a big spreadsheet, so there is a big list of fields to initialise when parsing the file. The field value extractions are not individually complex, there's just a lot of them. Splitting them out into groups as I've done is a bit clearer for the developer to see which values were pre-filled or filled by the customer (they download a csv, fill in some empty columns and re-upload it), but as you say doesn't make rubocop any happier.
If there's a better way to organise this I'd be happy to hear about it.

sdjmchattie: I've had a quick look and can't identify anything easy.  I was hoping we could change the threshold of these checks for just this file but haven't been able to see how that can be done. You might be able to remove the supression of `Metrics/CyclomaticComplexity` since `Metrics/PerceivedComplexity` is a subclass of that.  Other than this, maybe we do just have to accept this.  The problem with doing that is that now it's a free for all for complexity between these disable/enable tags which might mean poor code in the future.

app/models/presenters/plate_presenter.rb
sdjmchattie: Nope that explains it.  Maybe you could just add a comment to this so that it's clear why this is always `false` here.  Normally these booleans are based on properties of the class, but in this case you're just overriding base class behaviour.

andrewsparkes:
Put on hold

PR #1175, opened 2022-10-05
On Cardinal pooling, use sample metadata `collected_by`

harrietc52:
... Instead of sample manifest supplier name to group samples

Dependent on SS PR: https://github.com/sanger/sequencescape/pull/3678

(Partly) Closes https://github.com/sanger/sequencescape/issues/3646

* Update Cardinal pooling validator `RequireWellsWithSampleMetadata` to ensure wells have aliquots with sample, and sample metadata to ensure `collected_by` can be accessed
* Update `CardinalPoolsPlate` to use this validator
* Update tests


app/models/concerns/labware_creators/require_wells_with_sample_metadata.rb
```diff
@@ -1,28 +1,28 @@
 # frozen_string_literal: true
 
-# Can be included in plate creators which requires well aliquots to have a sample manifest
-module LabwareCreators::RequireWellsWithSampleManifest
+# Can be included in plate creators which requires well aliquots to have sample metadata
+module LabwareCreators::RequireWellsWithSampleMetadata
```
KatyTaylor, line 4: I think it would be more informative to call it `RequireWellsWithCollectedBy` (and equivalent changes to the method name etc.), as the current name implies it could be used by anyone who just wanted sample metadata, not that specific field.

KatyTaylor (approved):
Just a small naming suggestion, and a typo from previously.

PR #1269, opened 2023-04-05
Add stock presenter

emrojo:
Closes #

Changes proposed in this pull request:

*
*
* ...

PR #1268, opened 2023-04-04
Dpl566 bioscan library prep

andrewsparkes:
Closes #

Changes proposed in this pull request:
WIP Bioscan pipeline configs

PR #1271, opened 2023-04-06
Bioscan EPIC

harrietc52:
Includes
* https://github.com/sanger/limber/pull/1268
* https://github.com/sanger/limber/pull/1269


config/purposes/bioscan.yml
```diff
@@ -0,0 +1,110 @@
...
+LBSN-384 PCR 2:
+  :asset_type: plate
+  # TODO: May need a specific version of TaggedPlate creator to persist tag plate index 1-24
+  :creator_class: LabwareCreators::TaggedPlate
+  :presenter_class: Presenters::MinimalPcrPlatePresenter
+  :label_template: plate_6mm_double
+  :size: 384
+  # TODO check the names for this list of 24 tag layout templates (see tag plate gen stories)
+  :tag_layout_templates:
+    - Bioscan_384_template_1
```
andrewsparkes, line 54: This is a problem if we want to introduce new versions. this set should be flagged v2, but any new set is going to require a change to this config and a re-deploy.

config/robots.rb
```diff
@@ -2799,4 +2799,86 @@
...
+        }
+      }
+    )
+
+    custom_robot(
+      'beckman-lilys-96-stock-to-lbsn-96-lysate',
+      name: 'Beckman LILYS-96 Stock => LBSN-96 Lysate',
+      verify_robot: true,
+      beds: {
+        # TODO: confirm beds
```
sdjmchattie, line 2823: Does this need a backlog item adding and a reference to it putting here?  Bare TODOs live forever in the code and don't get done.

config/robots.rb
```diff
@@ -2799,4 +2799,86 @@
...
+          child: bed(5).barcode,
+          label: 'Bed 3'
+        },
+        bed(4).barcode => {
+          purpose: 'LBSN-96 Lysate',
+          states: ['passed'],
+          child: bed(5).barcode,
+          label: 'Bed 4'
+        },
+        # TODO: does the order of the parents here define the quadranting behaviour?
```
sdjmchattie, line 2868: Should create a backlog item if this TODO is still needed.

sdjmchattie (approved):
There are a couple of TODOs that I think should be put in the backlog, but otherwise looks good

config/purposes/bioscan.yml
```diff
@@ -0,0 +1,111 @@
...
+  # TODO: do we want any metadata entry?
+  # :custom_metadata_fields:
+  #   - metadatafieldname
+LBSN-9216 Lib PCR Pool XP:
+  :asset_type: tube
+  :target: MultiplexedLibraryTube
+  :type: IlluminaHtp::MxTubePurpose
+  :creator_class: LabwareCreators::TubeFromTube
+  # TODO: may need a modified tube presenter with paging for samples
+  :presenter_class: Presenters::SimpleTubePresenter
```
andrewsparkes, line 108: FinalTubePresenter

config/robots.rb
andrewsparkes: remove comment

config/robots.rb
andrewsparkes: remove comment, it does

config/purposes/bioscan.yml
andrewsparkes: Except there may be 2 sets in progress at a time, e.g. R&D are testing a new version whilst production is on the previous one. So after discussion we decided to build distinctly named tag groups and templates for each version.

PR #1276, opened 2023-04-20
DPL-718: Deep-well stamping

### LINKED ISSUE #1265: DPL-718: Add a pipeline to stamp deep-well plates into shallow-well plates
**User story**
As a lab technician, I want to be able to scan deep-well plates in Limber that were previously added to the system in Lighthouse UI.  This should be so that I can prepare to stamp their contents into a shallow-well plate via Limber.

Stuart M, Andrew S, Adrianne L, Sara S

TBC

**Acceptance criteria**
To be considered successful the solution must:
- [x] Provide an entry point for imported deep well plates.
- [x] Provide a button that creates a submission for a shallow well plate.
- [x] Allows the transfer of samples to the shallow well plate via:
    - [x] Manual transfer.
    - [x] A robot with bed verification.
- [ ] An integration suite has been created that tests the pipeline for deep-well plate stamping including bed verification on the robot.

References
- https://github.com/sanger/General-Backlog-Items/issues/283 -- this is an implementation of option 4 on this research story.

**Additional context**

sdjmchattie:
Closes #1265 

* Add config for a stamping pipeline in Limber.
* Set up bed verification for the Hamilton robot.


config/purposes/deep_well_96_stamping.yml
```diff
@@ -0,0 +1,26 @@
...
+  :cherrypickable_target: false
+  :input_plate: true
+  :creator_class: LabwareCreators::Uncreatable
+  :presenter_class: Presenters::SubmissionPlatePresenter
+  :submission_options:
+    Stamp from Deep-Well to Shallow-Well:
+      template_name: 'Limber-Htp - Deep Well Plate Stamping'
+      request_options:
+        fragment_size_required_from: '50'
+        fragment_size_required_to: '800'
```
emrojo, line 17: Is this needed? If is a config they asked ok, but I would think is not?

config/pipelines/high_throughput_deep_well_stamping.yml
```diff
@@ -0,0 +1,8 @@
+---
+# Stamping of deep well plates imported via Lighthouse UI to shallow well plates
+Deep-Well-96 Stamping:
+  pipeline_group: Deep-Well-96 Stamping
+  filters:
+    request_type_key: deep_well_plate_stamping
```
emrojo, line 6: This is a new request type that needs to be created in SS.

emrojo: Which I see is already there :)

config/purposes/deep_well_96_stamping.yml
sdjmchattie: This is a good point.  Andrew suggested that it may be necessary to provide some kind of value, since it's also used in Cardinal (I think).  But I will try removing it and see if everything still works, because it's true that this was only our suspicion at the time of creating this file.

sdjmchattie: I tried removing `request_options` entirely but that presents a stack trace explaining that you can't iterate with `each` on a `nil` object when trying to load the choose_workflow partial for the plate page due to this code:

```
<%= f.fields_for('request_options') do |ro| %>
  <% submission.request_options.each do |k,v| %>
    <%= ro.hidden_field k, value: v %>
  <% end %>
<% end %>
```

So I set `request_options: {}` which got a bit further but later said that "Request options can't be blank".  I'm pretty sure this error is coming from Sequencescape and not Limber.  However I can't find where in Sequencescape this is actually generated.

The best I could do was to use:

```
request_options:
  null_key: ''
```

which Sequencescape seems to ignore even though there are no other values it is expecting to see.  This feels about as hacky as providing unneeded keys that it does recognise but let me know what you think.

README.md
```diff
@@ -49,43 +49,49 @@ repository, as indicated:
    rbenv install
    ```
 
-2. In Limber, make the Bundler gem install the dependencies for this project:
+1. In Limber, make the Bundler gem install the dependencies for this project:
```
andrewsparkes, line 52: Annoyingly it's reset these numbers all back to 1, can you undo

andrewsparkes (approved):
Looks good to me (and we know it works).
Just revert that numbering change in the readme pls.

README.md
sdjmchattie: As discussed off of GitHub, we'll stick with 1s as it's an option you can use in Markdown and makes future diffs cleaner.

PR #1281, opened 2023-05-03
Prepare release for 4th May 2023

sdjmchattie:


README.md
```diff
@@ -49,43 +49,49 @@ repository, as indicated:
    rbenv install
    ```
 
-2. In Limber, make the Bundler gem install the dependencies for this project:
+1. In Limber, make the Bundler gem install the dependencies for this project:
```
stevieing, line 52: Looks like everything is now 1??

stevieing (approved):

README.md
sdjmchattie: Yes, this is valid MarkDown and means you don't have to change the numbering when you add or remove an item.  The only entry that must be a 1 is the first one.  You could make the rest of the list 6, 3, 2, 7, 1 if you wanted and it'll still render 1, 2, 3, 4, 5, 6

PR #1292, opened 2023-05-24
Bioscan uat updates from testing 24th May

andrewsparkes:
- added Mosquito bed verification for PCR 2 step (it does move the yellow 'perform pcr' down so less visible)
- limited plate scan boxes to 1 on pooling into 384 tube
- changed PCR 1 step to also use the Limber PCR presenter so visibly you see the same screen for both PCR 1 and 2 steps

### DIFF
--- a/app/models/labware_creators/pooled_tubes_from_whole_plates.rb
+++ b/app/models/labware_creators/pooled_tubes_from_whole_plates.rb
@@ -63,5 +63,10 @@ module LabwareCreators
     def plate_search
       api.search.find(Settings.searches['Find plates'])
     end
+
+    def number_of_parent_labwares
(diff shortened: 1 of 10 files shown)


yoldas (commented):
The changes look OK. However, when I executed the integration suite locally using the branch, I found two issues in control randomisation and fixed them.

1.  the well_restrictions rule type in config , which was expected to be well_exclusions in the code. 

2. the control_locations array inside generate_control_well_locations must be reset between retries.

We also needed a change in integration suite for bed-verification. We need to scan lbsn_384_pcr_1_plate to bed 5 as configured.

yoldas (approved):

PR #1293, opened 2023-05-26
DPL-749 Bioscan - traction compatible tube label class

andrewsparkes:
Tube label to have a human readable barcode for Bioscan.
- Merged #1308
- PCR2 Pool, Lib PCR Pool and Final XP tubes have human barcodes
- First line field is parent barcode and wells (asset name) for PCR2 Pool, empty for Lib PCR Pool, and Lib PCR barcode for the XP tube.
- The tubes above use the sprint template tube_label_traction_compatible.


yoldas (approved):
We have reviewed the changes together and I have added updates after the comments. One comment was changing the name plate_384_single as it is specific to Bioscan. I reviewed the code and I do not agree with that. I still keep the name because there is no other 'single sticker' 384-well plate label printing, it is generic enough, it is documented in code, and it is more useful by returning the first instead of last when there are multiple stock plates included.

PR #1308, opened 2023-06-29
Create 384 well plate single label template

yoldas:
Closes #

*
*
* ...

### DIFF
--- a/config/purposes/bioscan.yml
+++ b/config/purposes/bioscan.yml
@@ -45,7 +45,7 @@ LBSN-384 PCR 1:
   :creator_class: LabwareCreators::QuadrantStamp
   :merger_plate: true
   :presenter_class: Presenters::MinimalPcrPlatePresenter
-  :label_template: plate_6mm_double
+  :label_template: plate_384_single
   :size: 384
 LBSN-384 PCR 2:
   :asset_type: plate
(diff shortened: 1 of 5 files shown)


config/purposes/bioscan.yml
```diff
@@ -45,14 +45,16 @@ LBSN-384 PCR 1:
...
   :size: 384
+  :alternative_workline_identifier: LILYS-96 Stock
 LBSN-384 PCR 2:
   :asset_type: plate
   :creator_class: LabwareCreators::TaggedPlate
   :presenter_class: Presenters::MinimalPcrPlatePresenter
-  :label_template: plate_6mm_double
+  :label_template: plate_384_single
   :size: 384
+  :alternative_workline_identifier: LILYS-96 Stock
```
andrewsparkes, line 57: what does this do if there is no LILYS ancestor? Does it use the Lysate plate or blow up?

andrewsparkes (approved):

config/purposes/bioscan.yml
yoldas: It will try that first, if there are multiple lilys, it will return the "last" of them.
If it can't find it, it will return "last" stock plate available.
If it can't find anything, it will return nil, which will print a blank text at top-right of labels.

yoldas:
I am closing this PR because it has been already merged into the PR https://github.com/sanger/limber/issues/1344 and then to develop.

PR #1319, opened 2023-07-05
bioscan - removes v1 tags

andrewsparkes:


emrojo (approved):

andrewsparkes:
R&D not certain they won't be needed so leaving for now.

PR #1321, opened 2023-07-06
Bioscan - change to project name

andrewsparkes:
Request to use Bioscan as Sequencescape Project name in production rather than BIOSCAN UK for flying insects


config/purposes/bioscan.yml
```diff
@@ -31,7 +31,7 @@ LBSN-96 Lysate:
     - control_type: pcr negative
       name_prefix: CONTROL_NEG_
   :control_study_name: BIOSCAN UK for flying insects
```
yoldas, line 33: Do we keep the study name?

yoldas: Yes

yoldas (approved):

PR #1355, opened 2023-08-03
DPL-670-3 Use Sequencescape API key

### LINKED ISSUE #1322: DPL-670-3 Pass a Limber specific API key to Sequencescape calls
**User story**
As a Limber maintainer, I want to authenticate with a unique key when passing requests to Sequencescape.

sm49 hc6

**Acceptance criteria**
- [ ] Update the config for the app to accept an API key for Sequencescape calls on the v2 API.
- [x] Ensure the deployment project sets this key on deployed versions.
- [ ] Update the app to always provide this key in the headers for v2 SS endpoints.
- [ ] Redeploy with an app specific API key for Sequencescape.
- [ ] Check that requests coming from the app into Sequencescape are using the API key.

References
- https://github.com/sanger/deployment/issues/314

sdjmchattie:
Closes #1322 

- Add the API key for Sequencescape to V2 endpoint requests.

### DIFF
--- a/app/javascript/asset-comments/index.js
+++ b/app/javascript/asset-comments/index.js
@@ -44,6 +44,7 @@ document.addEventListener('DOMContentLoaded', () => {
      * Vue app if we actually find it */
     const userId = cookieJar(document.cookie).user_id
     const sequencescapeApiUrl = assetElem.dataset.sequencescapeApi
+    const sequencescapeApiKey = assetElem.dataset.sequencescapeApiKey
     const axiosInstance = axios.create({
       baseURL: sequencescapeApiUrl,
       timeout: 10000,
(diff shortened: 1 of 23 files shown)


andrewsparkes:
To test we think deploy to UAT and run integration suite, and it should run fine.
Could then change the key in deployment project to a non-valid key, and see if it breaks.

andrewsparkes (approved):

sdjmchattie:
> Q. How does SS validate this key, and what happens if it is invalid. e.g. does user get a suitable error message?

See https://github.com/sanger/sequencescape/pull/3812 and https://github.com/sanger/deployment/pull/323

> Q. Is the api key exposed in the Javascript views? does that matter if it is (public key?)

Yes it is and it is a concern.  However the API was otherwise unprotected before, so exposing the key doesn't open any new security holes.  The fix would be to have a server-side proxy for the requests, or to authenticate using the logged in user so that the token is specific to them.  Either way, the key should be changed again after those changes are made so that the old key can be invalidated since it might have been picked up by someone from the JS like you say.

PR #1401, opened 2023-09-28
dpl-873 Added a labware creator allowing partial stamped plate without dilution for 10x manual submission

Sangeetha-Bheeman:
Closes #

- DPL-873 - create new labware creator to allow for partial submissions of wells and to compress child wells to top left


andrewsparkes (changes requested):
This all looks good but you also need to add an rspec test in spec/models/labware_creators
Annoyingly there isn't an existing spec test for the PartialStampedPlate, but look at some of the other labware creator tests in that directory.

spec/models/labware_creators/partial_stamped_plate_without_dilution_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+require_relative 'shared_examples'
+
+# Uses a custom transfer template to transfer material into the new plate
+RSpec.describe LabwareCreators::PartialStampedPlateWithoutDilution do
+  it_behaves_like 'it only allows creation from plates'
+  it_behaves_like 'it has no custom page'
+
+  has_a_working_api
+
+  let(:well_a1) do
```
andrewsparkes, line 15: I think it might be a good idea to have a well which has several requests for different library types so there is some overlap like a real chromium example. 
Plus maybe a well with NO requests.
To give confidence the code still selects the right wells for a given library type filter.

These wells defined here all have one request each.

spec/models/labware_creators/partial_stamped_plate_without_dilution_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  describe '#parent_wells_to_transfer' do
+    it 'returns the filtered wells from the parent by library type' do
+      expect(subject.parent_wells_to_transfer).to eq [well_d1]
+    end
+  end
+
+  describe '#get_destination_location' do
```
andrewsparkes, line 90: Is this enough to test compression? Maybe two or three wells from the source that would have gaps to see that they compress to A1, B1, C1 etc. so can see they go in column order with no gaps.

andrewsparkes (changes requested):
Looks good, just think a couple more tests would be good to show the main features of this labware creator over the others (filtering of wells with multiple library types, compression)

spec/models/labware_creators/partial_stamped_plate_without_dilution_spec.rb
yoldas: I have compressed three wells D1, E1, B2 to A1, B1, C1 in the tests now.

andrewsparkes (approved):

PR #1404, opened 2023-09-28
DPL-811 Adding pbmc to fluidx tube racks labware creator

### LINKED ISSUE #1314: DPL-811 Transfer PBMC isolations into FluidX tube racks
**User story**
As a user of the scRNA Core Cell Extraction pipeline, I would like Limber to support the transfer of PBMC isolations into the FluidX tube rack at the end of the pipeline.

Lesley, Katy

Lesley

**Acceptance criteria**
- [x] Input plate is "LRC PBMC Bank"
- [x] Output is FluidX tubes, in 2 tube racks
- [x] User uploads tube rack scan csv files of the 2 destination tube racks, to tell Limber the destination tube barcodes
    - [x] During upload, user specifies which rack is destined for sequencing, and which for contingency (e.g. two separate labelled upload boxes)
    - [x] Tubes in the sequencing rack become "LRC Bank Seq" tubes
    - [x] Tubes in the contingency rack become "LRC Bank Spare" tubes
- [x] Destination tubes and transfer requests etc. are created in Limber, as per all transfers.
- [x] Failed wells are not transferred.

Transfer logic:
- [x] Transfer is 1:1 - each well in the plate will go into one destination tube
- [x] Labware creator should look at groups of wells that contain the same samples, and...
    - [x] 1 well in each group should go into a "LRC Bank Seq" tube, in the first rack
    - [x] any remaining wells in each group (2 expected) should go into a "LRC Bank Spare" tube, in the second rack
    - [x] If there is only 1 well in the group, the priority is to make a "LRC Bank Seq" tube

Note about reworks:
Sometimes an isolation will fail and will have to be reworked, and they will start again from the LRC Blood Aliquot tube at the start of the pipeline to get more cells. In this case, they may only want to produce "LRC Bank Spare" tubes (as the "LRC Bank Seq" will already have been produced from the first isolation). 
- [x] If they don't upload a rack scan file for the sequencing rack, allow this and only produce "LRC Bank Spare" tubes.

**Dependencies**
- https://github.com/sanger/limber/issues/1310

**Related Stories**
The following are related to this story:
- https://github.com/sanger/limber/issues/1333
- https://github.com/sanger/limber/issues/1333
- https://github.com/sanger/limber/issues/1331

**Additional information**
Although destination tubes are in tube racks, we will probably not represent the tube racks fully in the LIMS - see discussion here - https://github.com/sanger/limber/issues/1333#issuecomment-1682484672

Wireframes showing user experience, using screenshots from a mixture of the two Cardinal pipelines, and notes as to how will differ for this pipeline -https://docs.google.com/presentation/d/1giqhtjqjGErm1iZWbbw4mSLHsNBRSG68S1oAaFsuRlU/edit?usp=sharing 

This is a different process to how it works for Cardinal but it may still be useful to look at how that works, for instance because it also has the tube rack scan upload and the destination labware are FluidX tubes. See documentation for existing Cardinal pooling process here - [Pooling into FluidX tube rack (for cell banking)](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Cardinal+Pipeline+Documentation#CardinalPipelineDocumentation-PoolingintoFluidXtuberack(forcellbanking))

andrewsparkes:
Closes #1314 

Added a new labware creator to split the PBMC Bank plate wells into Sequencing and Contingency tube racks.

See corresponding DPL-811 Integration Suite branch


app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -0,0 +1,453 @@
...
+  #
+  # If after the initial preparation the users feel they need more contingency tubes, then they will go
+  # back to prepare more material for that sample from an earlier step in the pipeline and create just
+  # 'contingency' tubes at this step.
+  #
+  # Inputs:
+  # 1) The parent plate - contains a number of wells containing material prepared from the same sample.
+  #    The first of these parent wells will be transferred into a tube in the 'sequencing' rack if it is
+  #    present, and any remaining parent wells for the same sample will be transferred into tubes in the
+  #    'contingency' rack.
```
KatyTaylor, line 25: It's not quite clear here and in the paragraph starting on line 13 that there will be multiple groups like this. 
`contains groups of wells containing material prepared from the same sample.` - would that be clearer maybe?

app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -0,0 +1,453 @@
...
+    validates_nested :well_filter
+
+    # TODO: can add validations for specific metadata fields here if needed
+
+    # Don't create the tubes until at least the contingency file has been uploaded
+    validates :contingency_file, presence: true
+
+    # N.B. contingency file is required, sequencing file is optional
+    validates_nested :sequencing_csv_file, if: :contingency_file
+    validates_nested :contingency_csv_file, if: :sequencing_file
```
KatyTaylor, line 62: Are these meant to be swapped round? validates sequencing file if contingency is uploaded, and vice vera? Probably just misunderstanding something!

config/purposes/scrna_core_cell_extraction.wip.yml
```diff
@@ -86,14 +86,20 @@ LRC Bank Seq:
   :asset_type: tube
   :target: SampleTube
   :type: Tube::Purpose
-  :creator_class: LabwareCreators::PooledTubesBySample
+  :creator_class:
+    name: LabwareCreators::PlateSplitToTubeRacks
+    args: &fluidx_tube_creation_config
+      child_seq_tube_purpose_name: LRC Bank Seq
+      child_spare_tube_purpose_name: LRC Bank Spare
   :ancestor_stock_tube_purpose_name: LRC Blood Vac
```
KatyTaylor, line 94: Just wondering if we might want this to be the LRC Blood Aliquot... that will be the one that they go back to to repeat isolations from, and _think_ it'll be the one the requests come out of.

But then, the Vac tube barcode is the one that'll be in the Supplier Name - so this is probably correct 😅

app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -0,0 +1,453 @@
...
+
+          # TODO: put prefixes in the purpose config
+          name_for_details = name_for_details_hash('SPR', ancestor_tube_barcode(sample_uuid), tube_posn)
+
+          tube_name = name_for(name_for_details)
+          @contingency_wells_to_tube_posns[well] = tube_name
+
+          { name: tube_name, foreign_barcode: contingency_csv_file.position_details[tube_posn]['barcode'] }
+        end
+    end
```
KatyTaylor, line 399: Seems like a fair bit of code duplication between `generate_sequencing_tube_attributes` and `generate_contingency_tube_attributes` - could probably do with a refactor.

app/models/labware_creators/plate_split_to_tube_racks.rb
andrewsparkes: changed

config/purposes/scrna_core_cell_extraction.wip.yml
andrewsparkes: I think it's ok, the ultimate ancestor is always the Vac tube.
They definitely only go back to the Blood Aliquot tube, there is enough volume left there? The Vac has gone to another lab and is not accessible?
I'm using the barcode in the tube 'pool name' currently.
We may need to understand better how they plan to tell all these tubes apart given we aren't printing labels on them anyway.

app/models/labware_creators/plate_split_to_tube_racks.rb
andrewsparkes: refactored

config/purposes/scrna_core_cell_extraction.wip.yml
KatyTaylor: Yep think the Vac tube will have gone to faculty by this point, so it will be going back to the aliquot tube. 
I guess I assumed they would print barcodes for the aliquot tubes - but maybe they're FluidX ones - will check with Lesley.

app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -0,0 +1,602 @@
...
+    # @return [Boolean] true if the child tubes were created successfully.
+    def create_labware!
+      @child_sequencing_tubes = create_child_sequencing_tubes
+      @child_contingency_tubes = create_child_contingency_tubes
+      add_child_tube_metadata
+      perform_transfers
+      true
+    end
+
+    # Creates a single child sequencing tubes for each parent well containing a unique sample.
```
KatyTaylor, line 113:
```suggestion
    # Creates a single child sequencing tube for each parent well containing a unique sample.
```

app/models/labware_creators/plate_split_to_tube_racks.rb
KatyTaylor, line 1: It is huge, but generally really nice small methods and tons of documentation 👍

KatyTaylor (approved):

config/purposes/scrna_core_cell_extraction.wip.yml
KatyTaylor: Yep Lesley confirmed they'll print barcodes for the aliquot tubes

PR #1425, opened 2023-10-18
scRNA - fix for validation triggering prematurely when custom page present

andrewsparkes:
Fixed the Alert showing early on the PlateSplitToTubeRacks labware creator custom page


yoldas (approved):
Including CustomPage changes the creator button to make a link to a **new** labware page with params, instead of a button to **create** with params, which would call **save** and trigger validations early. The include changes which method to call on CreationController .

The if condition on sufficient_tubes_in_racks? validation is appropriate as without the contingency file, it would not make sense.

KatyTaylor (approved):

PR #1453, opened 2023-11-06
scRNA - fixes for file validations

andrewsparkes:
Changes:

1. Refactored sufficient tubes validation to not trigger if csv files are parsed incorrectly


spec/models/labware_creators/plate_split_to_tube_racks_spec.rb
```diff
@@ -242,13 +250,12 @@
 
     context 'when a contingency file is not present' do
       it 'does not call the validation' do
-        expect(subject).not_to receive(:sufficient_tubes_in_racks?)
+        subject.valid?
```
emrojo, line 253: maybe check that is false?
```suggestion
        expect(subject).not_to be_valid
```

spec/models/labware_creators/plate_split_to_tube_racks_spec.rb
```diff
@@ -291,32 +298,191 @@
...
-      before { allow(subject).to receive(:require_contingency_tubes_only?).and_return(false) }
+      before do
+        allow(subject).to receive(:require_contingency_tubes_only?).and_return(false)
+        subject.must_have_sufficient_tubes_in_rack_files
+      end
 
       context 'when there are enough tubes' do
-        it 'returns true' do
-          expect(subject.sufficient_tubes_in_racks?).to be true
+        it 'is valid' do
```
emrojo, line 307: shouldn't we have to do ```subject.validate``` ?

emrojo (approved):
All looks good

PR #1466, opened 2023-11-16
scRNA file validation fix - added check for valid files

andrewsparkes:
To add check for file valid?

### DIFF
--- a/app/models/labware_creators/plate_split_to_tube_racks.rb
+++ b/app/models/labware_creators/plate_split_to_tube_racks.rb
@@ -175,7 +175,7 @@ module LabwareCreators
     #
     # Sets errors if the tube rack barcodes are the same
     def check_tube_rack_barcodes_differ_between_files
-      return unless contingency_file_correctly_parsed? && sequencing_file_correctly_parsed?
+      return unless contingency_file_valid? && sequencing_file_valid?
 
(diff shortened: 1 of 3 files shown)


app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -240,17 +240,17 @@ def must_have_sufficient_tubes_in_rack_files
...
+      return contingency_file_valid? if require_contingency_tubes_only?
 
-      contingency_file_correctly_parsed? && sequencing_file_correctly_parsed?
+      contingency_file_valid? && sequencing_file_valid?
     end
 
-    def sequencing_file_correctly_parsed?
-      sequencing_file.present? && sequencing_csv_file.correctly_parsed?
+    def sequencing_file_valid?
+      sequencing_file.present? && sequencing_csv_file&.correctly_parsed? && sequencing_csv_file&.valid?
```
BenTopping, line 249: I think you can remove the check for `correctly_parsed?` here as it is part of the file validations so should be checked when `valid?` is called.

andrewsparkes: Ok will do thx

BenTopping (approved):

PR #1468, opened 2023-11-17
DPL-963: Fail pending tubes validation

### LINKED ISSUE #1437: DPL-963 [BUG] Don't allow 'pending' tubes to progress
**Describe the bug**
In the sample arraying screen in the scRNA Core Cell Extraction pipeline (transfer from LRC Blood Aliquot to LRC Blood Bank), it doesn't error if you scan in tubes that are in a 'pending' state.

**To Reproduce**
1. Make an LRC Blood Vac tube using the UAT Actions tab in Sequencescape.
2. Scan it into Limber
3. Click 'Add an empty Blood Aliquot tube'
4. Save the barcode somewhere NT1686895U
5. Make a second LRC Blood Vac tube using the UAT Actions tab in Sequencescape.
6. Scan it into Limber
7. Click 'Add an empty Blood Aliquot tube'
8. Click 'Manual Transfer'
9. Go back to the tube page (copy the barcode from the message and search for it again)
10. Save the barcode somewhere NT1686897W
11. Click 'Add an empty LRC Blood Bank plate'
12. Paste the two barcodes you saved into the boxes
13. Click 'Create'
14. Expected behaviour: it should error, saying something like 'These tubes are not yet passed: <barcode>'
15. Actual behaviour: it saves

StephenHulme:
Closes #1437 

- Reduce verbosity of unit-test output
- Refactor mock_api to fix long-outstanding bugs
- Add 'pending' state validation to tubes
  Screenshot 2023-11-17 at 09 18 03

NOTE:
This is currently failing against Intergration-Suite due to:

1. Long-outstanding error-feedback bugs have worked their way into the tests there 
   - `ENVIRONMENT=local bundle exec rspec spec/limber/bespoke_pcr_free_pipeline_spec.rb`
2. It expects 'pending' tubes to be added to racks - this has now been disallowed. 
   - `ENVIRONMENT=local bundle exec rspec ./spec/limber/cardinal_pipeline_spec.rb`


app/javascript/shared/components/LabwareScan.vue
```diff
@@ -103,13 +105,6 @@ export default {
       type: Object,
       default: null,
     },
-    validators: {
-      // An array of validators. See plateScanValidators.js and tubeScanValidators.js for examples and details
-      // defaults are set based on labwareType, in method 'computedValidators'
-      type: Array,
-      required: false,
-      default: null,
-    },
```
KatyTaylor, line 112: So was this prop not actually being used anywhere?

app/javascript/shared/components/LabwareScan.vue
```diff
@@ -184,12 +179,8 @@ export default {
       }
     },
     computedValidators() {
-      if (this.validators) {
-        return this.validators
-      }
-
       if (this.labwareType == 'tube') {
-        return []
+        return [checkState(['passed'])]
```
KatyTaylor, line 183: So this is changing the behaviour for everywhere that this component is used - do we think that's OK?

app/javascript/shared/components/LabwareScan.vue
```diff
@@ -42,6 +42,8 @@
 <script>
 import { checkSize } from './plateScanValidators'
 import { aggregate } from './scanValidators'
+import { checkState } from './tubeScanValidators'
```
KatyTaylor, line 45: Handy that this method already existed :)

app/javascript/shared/devourApiValidators.js
```diff
@@ -1,24 +1,43 @@
...
+          valid: false,
+          message: 'Results objects do not contain expected properties',
         }
       }
-      return { valid: true, message: 'Results valid' }
-    } else {
-      return { valid: false, message: 'No results retrieved' }
     }
+
+    return { valid: true, message: 'Results valid' }
```
KatyTaylor, line 39: This is more readable 👍 Is there test coverage to check it works the same?

app/javascript/shared/components/LabwareScan.vue
KatyTaylor: I'm not sure if it's safe. I'd have to look at where else it's being used and even then I'm not sure I could say whether we can assume it's always passed or not. I think it's quite likely that tag plates are scanned in in a 'pending' state... but then, that's a plate not a tube. Possibly the only other place it's being used is in Cardinal, in which case this would be fine. 
I guess the alternative would be to pass the validation rule in as a prop (the thing you got rid of)?

app/javascript/test_support/mock_api.js
```diff
@@ -1,3 +1,10 @@
+/*
+Mock the devour api to allow us to test the UI without making real requests to
+the server. This is done by adding a mock middleware to the devour api. The
+mock middleware intercepts requests and returns a mocked response. The mocked
+response is defined by calling mockGet() or mockFail() on the mock middleware.
+*/
+
```
KatyTaylor, line 7: 🙏 thanks for the docs

KatyTaylor (approved):
Approving as I believe Cardinal is the only other pipeline using the LabwareScan component with tubes rather than plates, and the change should be fine for that. However, if it was me, I think I would keep it more flexible by passing in the validation rule, in case in future we want to use it in other places where the validation is not appropriate.

app/javascript/shared/components/LabwareScan.vue
KatyTaylor: Yep I think it's fine to add this for Cardinal too, because they're doing the same thing in the lab in both pipelines.

PR #1472, opened 2023-11-22
ScRNA Core - UAT feedback fixes - driver and tube scan counts

andrewsparkes:
Fix for hamilton export file not showing latest set of tubes.
Added validation that tube rack scan files should have exactly the number of tubes they require.
Fixed some tests and added more specific error message checks.


app/views/exports/hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare.csv.erb
```diff
@@ -32,7 +32,7 @@
     next if source_well.downstream_tubes.empty?
 
     # expecting 1 sample tube
-    destination_tube = source_well.downstream_tubes.first
+    destination_tube = source_well.downstream_tubes.last
```
yoldas, line 35: We talked about state checking after `source_well.downstream_tubes` and then calling `last` and 
skipping if we can't find the destination tube.

app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -218,23 +217,24 @@ def extract_barcodes(file)
...
 
       if require_contingency_tubes_only?
-        add_error_if_insufficient_tubes(:contingency_csv_file, num_contingency_tubes, num_parent_wells)
+        add_error_if_wrong_number_of_tubes(:contingency_csv_file, num_contingency_tubes, num_parent_wells)
       else
-        add_error_if_insufficient_tubes(
+        add_error_if_wrong_number_of_tubes(
           :contingency_csv_file,
           num_contingency_tubes,
           num_parent_wells - num_parent_unique_samples
```
yoldas, line 235: We talked about if `num_parent_wells - num_parent_unique_samples` is zero, then `required_tubes` argument will be zero, rejecting any number of contingency tubes.

spec/models/labware_creators/plate_split_to_tube_racks_spec.rb
```diff
@@ -909,6 +1030,7 @@
           stub_asset_search(child_tube_4_v2.barcode.machine, child_tube_4_v1)
         end
 
+        # TODO: this one
```
yoldas, line 1033: We could remove this comment.

yoldas (approved):
Three comments: Of three, we have talked about two already. The last one is minor (remove TODO comment).

app/views/exports/hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare.csv.erb
andrewsparkes: I've realised there is already a check on the parent well state being 'passed' or we skip the position.
A non-passed well should not have created any chid tubes anyway.
For passed wells at this point the child tubes should have nothing in them and should be in 'pending' state, which is based on the state of the transfer requests into them (the transfer has not happened yet so pending).
So a state check on the tube.state = pending or skip should be all we need for now until we add the uuid stories

spec/models/labware_creators/plate_split_to_tube_racks_spec.rb
andrewsparkes: done

PR #1492, opened 2023-12-04
scRNA fix - to add barcodes to export filenames

andrewsparkes:
Adds barcode to filenames for the two hamilton driver files


yoldas (approved):
This change adds the labware barcode **to the end** of the CSV export (before the .csv extension). The purpose config for the labware contains an entry for the export in the file_links section.

PR #1494, opened 2023-12-06
FIX added plate purpose type to LILYS plate config

andrewsparkes:
Minor config change to LILYS plate config in Bioscan pipeline


stevieing:
closed as stale.

PR #1500, opened 2023-12-11
DPL-937 - Updates to RVI BCL pipeline

BenTopping:

- Adds RVI prefix to bait capture library pipeline
- Updates robot verifications for rvi bcl pipeline
- Adds custom presenter to RVI cDNA plate to allow custom 'Manual transfer' state change.

### DIFF
--- a/app/models/presenters/rvi_cdna_xp_presenter.rb
+++ b/app/models/presenters/rvi_cdna_xp_presenter.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+module Presenters
+  # Used for RVI cDNA XP plates. Used for a custom 'Manual Transfer' state transition.
+  # N.B. 'Manual Transfer' is used for the RVI BCL pipeline for the cDNA XP Plate because the plates
+  # are inaccessible in the bed at the time of transfer so cannot be transferred via
+  # bed verifications. Also we cannot set its state to 'Passed' on initial bed verification
+  # because there are steps on other plates that need to be completed before the cDNA XP Plate can be transferred.
+  # and we do not want to misreport the plate being passed if an issue arises with the steps before hand.
+  class RviCdnaXpPresenter < PlatePresenter
+    include Statemachine::Shared
+
+    # Same state machine as permissive plate presenter, but with a custom 'Manual Transfer' state transition.
+    state_machine :state, initial: :pending do
+      event :take_default_path, human_name: 'Manual Transfer' do
+        transition pending: :started
+        transition started: :passed
+      end
+
+      event :transfer do
+        transition %i[pending started] => :passed
+      end
+
+      event :cancel do
+        transition %i[pending started passed] => :cancelled
+      end
+
+      # We use `fail_labware` here as `fail` is defined on Object (its an alias for `raise`)
+      # Statemachines ends up throwing a warning, which while we can disable, we're probably
+      # safer avoiding the conflict, especially as we don't actually call these methods directly
+      # anyway.
+      event :fail_labware, human_name: 'Fail' do
+        transition %i[pending started passed] => :failed
+      end
+
+      # These are the states, which are really the only things we need ...
+      state :pending do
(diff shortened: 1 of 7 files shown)


app/models/presenters/rvi_cdna_xp_presenter.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module Presenters
+  # Used for RVI cDNA XP plates. Used for a custom 'Manual Transfer' state transition.
+  # N.B. 'Manual Transfer' is used for the RVI BCL pipeline for the cDNA XP Plate because the plates
+  # are inaccessible in the bed at the time of transfer so cannot be transferred via
+  # bed verifications. Also we cannot set its state to 'Passed' on initial bed verification
+  # because there are steps on other plates that need to be completed before the cDNA XP Plate can be transferred.
+  # and we do not want to misreport the plate being passed if an issue arises with the steps before hand.
+  class RviCdnaXpPresenter < PlatePresenter
```
andrewsparkes, line 10: Minor thing but the capitalisation in the classname looks odd to me.
`RVIcDNAXpPresenter`?
RVI and DNA are already initialisms of longer titles.
c in cDNA means complimentary and is written in lower case.
Dunno if doing that causes any class loading issues mind.

app/models/presenters/rvi_cdna_xp_presenter.rb
```diff
@@ -0,0 +1,85 @@
...
+      # Statemachines ends up throwing a warning, which while we can disable, we're probably
+      # safer avoiding the conflict, especially as we don't actually call these methods directly
+      # anyway.
+      event :fail_labware, human_name: 'Fail' do
+        transition %i[pending started passed] => :failed
+      end
+
+      # These are the states, which are really the only things we need ...
+      state :pending do
+        include Statemachine::StateAllowsChildCreation
```
andrewsparkes, line 38: Just checking, at the point they want to click on the 'Add Ligand' plate button is the cDNA plate still in pending state (or in started or another later state)? I'd delay allowing ChildCreation as long as possible. But if it is in pending state this looks ok.

app/models/presenters/rvi_cdna_xp_presenter.rb
```diff
@@ -0,0 +1,85 @@
...
+  # bed verifications. Also we cannot set its state to 'Passed' on initial bed verification
+  # because there are steps on other plates that need to be completed before the cDNA XP Plate can be transferred.
+  # and we do not want to misreport the plate being passed if an issue arises with the steps before hand.
+  class RviCdnaXpPresenter < PlatePresenter
+    include Statemachine::Shared
+
+    # Same state machine as permissive plate presenter, but with a custom 'Manual Transfer' state transition.
+    state_machine :state, initial: :pending do
+      event :take_default_path, human_name: 'Manual Transfer' do
+        transition pending: :started
```
andrewsparkes, line 16: should there be 2 transitions here? if this presenter is for one specific plate shouldn't there by only one state change by manual transfer? or is the first backup for if the robot (bed verification) is not available for the first state transition?

andrewsparkes (approved):

app/models/presenters/rvi_cdna_xp_presenter.rb
BenTopping: I agree, not the nicest capitalisation, but having the name as anything else causes loading issues unfortunately.

app/models/presenters/rvi_cdna_xp_presenter.rb
BenTopping: Yes good spot, will change so only 'started' and 'passed' can have child creation.

app/models/presenters/rvi_cdna_xp_presenter.rb
BenTopping: Yeah i included started: :passed in the case the bed verification cant be completed. But that should never happen so probably safe to remove?

PR #1513, opened 2023-12-13
Revert "Bioscan barcode label changes"

yoldas:
Reverts sanger/limber#1493

This PR is for reverting the changes to bioscan label printing changes for PCR 1 and PCR 2 plates (showing LYSATE ancestor top-left) and PCR2 Pool tube (immediate parent in first line). The reason for the revert is that UAT is not finished yet and we have to deploy Limber to production before the end of year future freeze.


yoldas:
This PR is not necessary anymore as the changes are accepted in UAT and there is no need to revert the merge.

PR #1587, opened 2024-02-19
docs: add some troubleshooting tips to readme

StephenHulme:
Closes some long-standing debugging niggles

- Add some troubleshooting tips to the readme

**Please feel free to add any known issues and fixes that you are aware of**

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,8 +121,8 @@ yarn prettier --write .
 If during development changes do not seem to be taking effect, try:
 
 - Restart the application:
-- Restart the Docker container `docker-compose down && GENERATE_CONFIG=false docker-compose up -d`
-- Rebuild the Docker image `GENERATE_CONFIG=false docker-compose up --build -d`
+- Destroy and recreate the Docker container `docker-compose down && GENERATE_CONFIG=false docker-compose up -d`
+- Rebuild the Docker image, particularly useful for changing dependencies
 - Clobber local resources `rails assets:clobber`
 
 ## Note about the remainder of this document


README.md
```diff
@@ -116,6 +116,15 @@ yarn prettier --check .
 yarn prettier --write .
 ```
 
+## Troubleshooting
+
+If during development changes do not seem to be taking effect, try:
+
+- Restart the application:
+- Restart the Docker container `docker-compose down && GENERATE_CONFIG=false docker-compose up -d`
+- Rebuild the Docker image `GENERATE_CONFIG=false docker-compose up --build -d`
```
sdjmchattie, line 125: I sometimes find Docker compose can be a bit obsessed with caching and that the best fix is actually to rebuild without caches.  Unfortunately this means using `docker-compose build --no-cache` and then `docker-compose up -d` afterwards.

README.md
```diff
@@ -116,6 +116,15 @@ yarn prettier --check .
 yarn prettier --write .
 ```
 
+## Troubleshooting
+
+If during development changes do not seem to be taking effect, try:
+
+- Restart the application:
+- Restart the Docker container `docker-compose down && GENERATE_CONFIG=false docker-compose up -d`
```
sdjmchattie, line 124: If you intend to destroy and recreate the container this sequence is fine, but you labelled it "restart" which would be `docker-compose restart`.  Down destroys the containers and networks.  Up creates them from the image and compose file.

README.md
dasunpubudumal: Agreed. But then again, this would mean that **all the images** would have to be re-pulled from docker repositories (hub). Takes time for the process but seems like the best fix.

sdjmchattie: No it still uses the cached version of images pulled from the hub.  It just ensures you repeat every step in your `Dockerfile` and don't use cached version of those steps.  It does take time to redo the steps, but it also ensures a clean image.

sdjmchattie: > What I find valuable about the steps above is the rebuilding forces packages to be reinstalled and the down-and-up re-evaluates everything after the code-sync step.
> 
> Maybe it's worth breaking these steps down into the specific things that each one addresses and add the no-cache option where it's appropiate?

You can't get it to re-do all the steps in your `Dockerfile` when you do `docker-compose up --build -d`.  If it thinks it has a cached version of the final image in the sequence it will do none of the build steps.  You have to use `docker-compose build --no-cache` which builds (and caches) the image afresh, but doesn't run a container.

sdjmchattie: The problem is that it only has the contents of the `Dockerfile` to go by when determining whether a cache of a step is appropriate.  If you've changed the dependencies file in the repo, Docker compose is unlikely to realise and will not get your packages installed the versions you expected.

Note that `--no-cache` cannot be added to the `docker-compose up` command.  It MUST be `docker-compose build --no-cache`.  Only the `build` command supports the flag.  I've never found `docker-compose up --build` to be a reliable way to ensure a fix.  When it comes to fixing an unknown issue, nuke and pave gets you there the fastest.

sdjmchattie (approved):

PR #1595, opened 2024-02-26
DPL-1054 Add Vac tube barcode to the Aliquot tube labels

### LINKED ISSUE #1546: DPL-1054 Add Vac tube barcode to the Aliquot tube labels
**User story**
As a user of the scRNA Core Cell Extraction pipeline, I would like the LRC Blood Aliquot labels to display the parent LRC Blood Vac tube barcode (as well as the current information), so that it is easier to see the relation between the two tubes.

Abby, Katy

Abby

**Acceptance criteria**
- [ ] LRC Blood Aliquot barcode label should display the parent LRC Blood Vac tube barcode (as well as the current information)
- [ ] Printing label is tested using the relevant printer (r152-bc1)

**Dependencies**
Probably best to implement https://github.com/sanger/limber/issues/1536 first, as it might change how this story is implemented.

**Additional context**
This request came out of the training sessions.

KatyTaylor:
Closes #1546 

Add workline reference functionality for tubes as well as plates
- add LRC Blood Vac as the alternative workline identifier for LRC Blood Aliquot tubes
- the top right section is blank on the current LRC Blood Aliquot labels, so this will fill it in


app/sequencescape/sequencescape/api/v2/shared/has_workline_identifier.rb
```diff
@@ -0,0 +1,21 @@
...
+      return stock_plate if alternative_workline_identifier_purpose.nil?
+
+      ancestors.where(purpose_name: alternative_workline_identifier_purpose).last
+    end
+
+    def workline_identifier
+      workline_reference&.barcode&.human
+    end
+  end
+end
```
yoldas, line 21: This code seems to be extracted from plate so it can be included in other labware. The common use of the identifier is using it at the top-right field if it is a Plate barcode. Label templates delegate the workline_identifier method to their labware. One can still override it inside label template or labware if this concern is included. Making this a concern is a very good approach. For tubes though, the printed label can use it for one of the lines in the label. Maybe the comment of workline_reference of the Concern could reflect that.

yoldas (approved):
Looks good. Thanks for refactoring it as a Concern.

app/sequencescape/sequencescape/api/v2/shared/has_workline_identifier.rb
```diff
@@ -0,0 +1,21 @@
...
+  module HasWorklineIdentifier
+    extend ActiveSupport::Concern
+
+    # Finds the labware whose barcode will be printed in the label at the top_right field.
+    # Uses alternative_workline_identifier from the purpose config if present, otherwise uses the stock plate.
+    def workline_reference
+      alternative_workline_identifier_purpose = SearchHelper.alternative_workline_reference_name(self)
+      return stock_plate if alternative_workline_identifier_purpose.nil?
+
+      ancestors.where(purpose_name: alternative_workline_identifier_purpose).last
```
sdjmchattie, line 14: I wonder if this would be easier for someone to follow if we avoided the inline if on the first line and made it clear that the other line would be returned in the else case?

```suggestion
      if alternative_workline_identifier_purpose.nil?
        return stock_plate
      else
        return ancestors.where(purpose_name: alternative_workline_identifier_purpose).last
      end
```

I think you can even drop the `return` key words then, but that leaves scope for more code being added below this block and creating an unexpected bug.

sdjmchattie (approved):
Minor suggestion for clarity

app/sequencescape/sequencescape/api/v2/shared/has_workline_identifier.rb
KatyTaylor: Thanks, I think I agree it's clearer but Rubocop doesn't agree - https://docs.rubocop.org/rubocop/cops_style.html#styleguardclause

So I think I'll stick with how it is in the interests of keeping the code style consistent.

KatyTaylor: Maybe it can't tell the difference?

PR #1609, opened 2024-03-03
DPL-1024 PBMC donor pooling - driver file

yoldas:
Closes #

- 

### DIFF
--- a/app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
+++ b/app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
@@ -1,0 +1,17 @@
+<%# This is a picking file for a Hamilton robot for LRC PBMC Defrost plate to LRC PBMC Pools plate transfers %>
+<%= CSV.generate_line [
+    'Workflow', 
+    @workflow
+  ], 
+  row_sep: '' 
+%>
+<%= CSV.generate_line [
+    'SourcePlate',
+    'SourceWell',
+    'DestinationPlate',
+    'DestinationWell',
+    'SampleVolume',
+    'ResuspensionVolume',
+  ],
+  row_sep: ''
+%>
--- a/app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
+++ b/app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
@@ -22,21 +22,19 @@
   maximum_sample_volume = 50.0  # microlitres
   resuspension_volume_per_sample = 2.2  # microlitres
   minimum_resuspension_volume = 10.0  # microlitres
-  MILLILITRES_TO_MICROLITRES = 1_000.0
+  millilitres_to_microlitres = 1_000.0
+
+  # Make a mapping { src_barcode => { src_location => src_well } }
+  ancestor_plates_wells = @ancestor_plate_list.each_with_object({}) do |ancestor_plate, hash|
+    hash[ancestor_plate.labware_barcode.human] = ancestor_plate.wells.index_by(&:location)
+  end
 
   transfer_request_data = []
-  ancestor_plates_wells = {}  # { src_barcode => { src_location => src_well } }
   each_source_metadata_for_plate(@plate) do |src_barcode, src_location, dest_well|
-
-    # Fetch ancestor plate specified by src_barcode and index wells by src_location.
-    ancestor_plates_wells[src_barcode] ||= plate_by_barcode(src_barcode).wells.index_by(&:location)
-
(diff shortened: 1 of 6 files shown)


app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
KatyTaylor, line 1: Since we might not be UAT-ing for a little while yet, it might be worth generating a filled in example file and pasting it in the Slack channel - so that R&D can check it - because I think we didn't really discuss this driver file after the change to allow 2 source plates? (In the interests of early feedback)

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
```diff
@@ -0,0 +1,58 @@
...
+  # Make a mapping { src_barcode => { src_location => src_well } }
+  ancestor_plates_wells = @ancestor_plate_list.each_with_object({}) do |ancestor_plate, hash|
+    hash[ancestor_plate.labware_barcode.human] = ancestor_plate.wells.index_by(&:location)
+  end
+
+  transfer_request_data = []
+  each_source_metadata_for_plate(@plate) do |src_barcode, src_location, dest_well|
+    src_well = ancestor_plates_wells[src_barcode][src_location]
+    cell_count = src_well.latest_live_cell_count  # cells / millilitres
+    next if cell_count.nil?
```
KatyTaylor, line 36: There seems to be a bit of an inconsistency here, because the LIMS will look as if that sample was transferred, right? Because the sample is already present in the destination well?

Not sure how this should behave, if the cell count data is not present.

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
KatyTaylor: Am I right in saying this won't generate rows for failed wells? Because it's driven by transfer requests, and a transfer request won't have been generated for a failed well?

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
```diff
@@ -0,0 +1,58 @@
...
+  ], 
+  row_sep: '' 
+%>
+<%= CSV.generate_line [
+    'SourcePlate',
+    'SourceWell',
+    'DestinationPlate',
+    'DestinationWell',
+    'SampleVolume',
+    'ResuspensionVolume',
```
KatyTaylor, line 14: Would be nice to have the unit in the headings for `SampleVolume` and `ResuspensionVolume`?

KatyTaylor (commented):
A few questions & ideas

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
yoldas: I have been thinking about generating it using integration suite.  I will ask with a made-up example file this week.

yoldas: Yes, failed wells won't be transferred.

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
yoldas: I don't think there is an inconsistency. The plate will be created. However, driver file download means samples are not transferred yet. Bed-verification will be done (when the story is done) and transfer will be started. 

We did not include the count data in the transfer request. We collect it from source wells. Whenever the QC data are available upstream, a fresh new driver file download will include the latest QC. Are you suggesting that we should have some kind of verification?

At the moment, I am skipping the rows if cell count is not available, similar to Cardinal.

app/views/exports/hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools.csv.erb
yoldas: None of the exports in Limber has units headers. It made think that maybe it is on purpose. I have added the units as you suggested now, also made the headers more readable.

KatyTaylor (approved):
Thanks for answering my comments. Approving as I think it's fine to merge in. As discussed, a couple of things to check with R&D:
- Behaviour if cell count data is missing for some wells (no transfer / take default volume / error ?) - if no transfer, how can we reflect this in the LIMS (e.g. fail transfer request)
- units in headers - if we're not confident about the units, we could check this with R&D (but I think they're probably fine)

PR #1625, opened 2024-03-14
DPL-1024-2 PBMC donor pooling - ancestor plates

yoldas:
Closes #

Updates for using multiple ancestor plates in driver file download

Summary:

The ancestor_purpose setting of an export configuration will make an additional controller instance variable`@ancestor_plate_list` populated with all ancestor plates of that purpose and available in the view template.

Background:

The existing export controller allows making an ancestor plate available to the export view template by specifying the ancestor plate purpose in the export configuration. This is useful in several cases, for example, access to parent plate barcode and relating source wells specified in transfer templates in driver file exports. However, only one ancestor plate is made available to the view, the first, which assumes a single parent for the current plate being viewed and download requested.

In cases where a plate has (or may have) multiple ancestors of a specific purpose, such as pooling from multiple source plates, as in scRNA donor pooling, we need access to more than one parent in the view to interpret the transfer request and generate a column with parent barcode. This PR enables that.

### DIFF
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -14,6 +14,7 @@ class ExportsController < ApplicationController
     @workflow = export.workflow
     @ancestor_plate = locate_ancestor_plate
     @ancestor_tubes = locate_ancestor_tubes
+    @ancestor_plate_list = locate_ancestor_plate_list
 
     set_filename if export.filename
 
@@ -47,6 +48,20 @@ class ExportsController < ApplicationController
     return nil if ancestor_result.blank?
 
     Sequencescape::Api::V2.plate_with_custom_includes(include_parameters, id: ancestor_result.id)
+  end
+
+  # Returns an array of all ancestor plates of @plate that match a specific
+  # purpose (defined in the export configuration). If no such plates are found,
+  # an empty array is returned. The result array is made available to the view
+  # that generates the export as @ancestor_plate_list by the show method.
+  #
+  # @return [Array, Sequencescape::Api::V2::Plate] An array of Plate records if
+  #   any are found, otherwise an empty array.
+  def locate_ancestor_plate_list
+    # Collect plate ids from Asset resource results to fetch Plates later.
+    ids = @plate.ancestors.where(purpose_name: export.ancestor_purpose).map(&:id)
+    return [] if ids.empty?
+    Sequencescape::Api::V2::Plate.includes(include_parameters).find({ id: ids })
   end
 
   def include_parameters
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -60,7 +60,8 @@ class ExportsController < ApplicationController
   def locate_ancestor_plate_list
     return [] if export.ancestor_purpose.blank?
 
-    # Collect plate ids from Asset resource results to fetch Plates later.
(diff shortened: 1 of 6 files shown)


app/controllers/exports_controller.rb
```diff
@@ -43,16 +44,28 @@ def locate_labware
...
 
+  # Returns an array of all ancestor plates of @plate that match a specific
+  # purpose (defined in the export configuration). If no such plates are found,
+  # an empty array is returned. The result array is made available to the view
+  # that generates the export as @ancestor_plate_list by the show method.
+  #
+  # @return [Array, Sequencescape::Api::V2::Plate] An array of Plate records if
+  #   any are found, otherwise an empty array.
+  def locate_ancestor_plate_list
+    return nil if export.ancestor_purpose.blank?
```
andrewsparkes, line 61: should that return empty array instead of nil to be consistent?

app/controllers/exports_controller.rb
```diff
@@ -43,16 +44,28 @@ def locate_labware
...
   end
 
+  # Returns an array of all ancestor plates of @plate that match a specific
+  # purpose (defined in the export configuration). If no such plates are found,
+  # an empty array is returned. The result array is made available to the view
+  # that generates the export as @ancestor_plate_list by the show method.
+  #
+  # @return [Array, Sequencescape::Api::V2::Plate] An array of Plate records if
+  #   any are found, otherwise an empty array.
+  def locate_ancestor_plate_list
```
andrewsparkes, line 60: Please add an rspec test for this method.
I think it's useful and will be used in other pipelines.

andrewsparkes (changes requested):
Looks good, just add a test and see comment about nil / empty array.

KatyTaylor:
Could you add a bit more to the PR description, explaining what the problem was that you're fixing? Thanks

app/controllers/exports_controller.rb
```diff
@@ -43,16 +44,28 @@ def locate_labware
...
+    return nil if export.ancestor_purpose.blank?
+
+    # Collect plate ids from Asset resource results to fetch Plates later.
```
KatyTaylor, line 63: Can we avoid using the word Asset? 
1. Isn't @plate actually coming from a V2 Plate API query in `locate_labware`?
2. Asset is a historical term from before it was split into Labware and Receptacle, and I'd like to get rid of all references to the word 'Asset' from SS & Limber to avoid confusion.

app/controllers/exports_controller.rb
yoldas: Thanks for spotting that. Yes, it should return an empty array as described in the method documentation.

app/controllers/exports_controller.rb
yoldas: I have added test for the behaviour of the controller when dealing with multiple ancestor plates using this method and configuration.  It verifies that @ancestor_plate_list is correctly assigned and the controller makes it available to the view template and the response body matches the expected output, in two contexts when it is configured and not configured.

yoldas:
> Could you add a bit more to the PR description, explaining what the problem was that you're fixing? Thanks

I have updated the PR description to explain the problem being fixed.

app/controllers/exports_controller.rb
yoldas: I have updated the method documentation to make it clear.

KatyTaylor (approved):
Thanks for the more detailed description

PR #1635, opened 2024-03-19
DPL-1099 Added double label for bioscan lysate plate

andrewsparkes:
Closes #4019

Adds a second label for the Bioscan lysate plate which has the supplier partner plate id barcode on it that was stored in the sample metadata in sample_description.

Needs physical testing onsite.


app/models/labels/plate_label_lbsn_96_lysate.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    # replace any underscores with hyphens (printer software can't handle underscores)
+    partner_id = partner_id.tr('_', '-')
+
+    # add the required suffix
+    [partner_id, PARTNER_ID_SUFFIX].compact.join('-')
+  end
+
+  # Fetch the partner id from the first well that has a sample.
+  # Assumption: The partner id is stored in the sample description, and all wells woth
```
yoldas, line 51: typo: all wells **with**

app/models/labels/plate_label_lbsn_96_lysate.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  # Fetch the first sample from the first well that has a sample (not a control)
+  def sample_from_first_populated_well
+    labware.wells_in_columns.each do |well|
+      return well.aliquots.first.sample if well.aliquots.present? && well.aliquots.first.sample.control != true
+    end
+
+    raise StandardError, 'No wells with aliquots found in this labware to fetch a sample'
+  end
```
yoldas, line 69: Do we need to change plate includes for this to work? Is `wells.aliquots.sample.sample_metadata` already included?

app/models/labels/plate_label_lbsn_96_lysate.rb
```diff
@@ -0,0 +1,82 @@
...
+  # Assumption: The partner id is stored in the sample description, and all wells woth
+  # samples on the plate have the same partner id.
+  def fetch_partner_id_for_plate
+    partner_id = sample_from_first_populated_well&.sample_metadata&.sample_description
+
+    raise StandardError, 'Unable to fetch partner id' if partner_id.blank?
+
+    # trim to max length if needed
+    partner_id.truncate(MAX_LENGTH_PARTNER_ID, omission: '')
+  end
```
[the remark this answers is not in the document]
yoldas: "Sample manifest on creating LILYS contains partner id in sample metadata 'sample_description'. Likewise if they do faculty input manifest at Lysate, so should always be present."

app/models/labels/plate_label_lbsn_96_lysate.rb
yoldas: "Works on label printer in morgan reception and includes has sample_metadata yes"

yoldas (approved):
Looks good.

app/models/labels/plate_label_lbsn_96_lysate.rb
```diff
@@ -0,0 +1,70 @@
...
+  PARTNER_INFO_TEXT = 'PARTNER ID LABEL'
+  PARTNER_ID_SUFFIX = 'SDC'
+
+  # Define the standard first label for the lysate plate
+  def attributes
+    super.merge(barcode: labware.barcode.human)
+  end
+
+  # Define the second label for the lysate plate
+  def intermediate_attributes
```
sdjmchattie, line 18: Would it be worth a refactor on the name of these attributes?  It seems like this name was perhaps relevant to another type of label printing but I think in this case you're trying to print two labels to capture more information in the same piece of labware?

I just wouldn't know what intermediate attributes means if I was looking at this code for the first time.

sdjmchattie (commented):
Unsure whether a restructure/refactor is needed here.  I won't approve or request changes.

app/models/labels/plate_label_lbsn_96_lysate.rb
andrewsparkes: Have renamed a couple of those methods in the label class to (I think) make more sense.
I don't really follow why `extra_attributes` is separated (it seems to make an extra 6mm label for 384-well plates but why is that different to additional_attributes?) or what `sprint_attributes` is doing.

PR #1655, opened 2024-03-28
DPL-1099 Fix to handle scenario where sample description has not been set

andrewsparkes:
Fix after UAT
500 server error when sample description was not set. Better to put message on label.


docs/creators.md
```diff
@@ -26,8 +26,8 @@ Labware creators are responsible for creating new labware from a parent labware.
...
+  Used directly in 110 purposes:
```
andrewsparkes, line 29: why has this changed when I did nothing with labware creators? why? why?

KatyTaylor (commented):
Looks good apart from one typo (not in this PR)!

PR #1661, opened 2024-04-04
Y24-042 scRNA - Fix bug in file handling

### LINKED ISSUE #1659: Y24-042 'Cryostor' tube rack scan file improvements
**User story**
As a user of the cDNA Prep part of the scRNA Core pipeline, I would like the rack scan format (used for the initial Seq/Spare tubes --> Cryostor plate/rack step) to match that required by a previous step in the pipeline, so that the tube rack scanner only has to have one output format, reducing scope for user error.

Abby, Katy

Abby

**Acceptance criteria**
- [x] Upload page accepts files that include an extra (first) column, containing the tube rack barcode (example file - [Cryostor plate rack scan_3col_correct (1).csv](https://github.com/sanger/limber/files/14837111/Cryostor.plate.rack.scan_3col_correct.1.csv)).
- [ ] It could also accept the previous format, but does not need to. [Not done as not easy to implement]
- [x] This first column of data does not need to be saved (plate barcode is printed out and stuck on the tube rack).
- [x] Currently, if the attached file is uploaded, the LIMS throws an exception. It should handle errors gracefully if the file format is not as expected.

**References**
This is an enhancement to https://github.com/sanger/limber/issues/1388

**Additional context**
This came up during UAT.

andrewsparkes:
Closes #1659 

Fix for bug to do with routing when displaying errors when creating Crystor plate.
Change to expect csv file upload to contain a tube rack barcode column.


app/models/labware_creators/multi_stamp_tubes_using_tube_rack_scan.rb
```diff
@@ -74,7 +74,7 @@ def create_labware!
 
     # Returns a CsvFile object for the tube rack scan CSV file, or nil if the file doesn't exist.
     def csv_file
-      @csv_file ||= CommonFileHandling::CsvFileForTubeRack.new(file) if file
+      @csv_file ||= CommonFileHandling::CsvFileForTubeRackWithRackBarcode.new(file) if file
```
KatyTaylor, line 77: nice small change here!

app/models/labware_creators/common_file_handling/csv_file/row_for_tube_rack_with_rack_barcode.rb
```diff
@@ -24,6 +25,10 @@ def initialize_context_specific_fields
         @tube_position = (@row_data[1] || '').strip.upcase
         @tube_barcode = (@row_data[2] || '').strip.upcase
       end
+
+      def expected_number_of_columns
+        EXPECTED_NUMBER_OF_COLUMNS
+      end
```
KatyTaylor, line 31: I don't like this because you're defining the exact same method in the subclass (RowForTubeRackWithRackBarcode) and the superclass (RowForTubeRack). So, it looks like you could just get rid of this method in the subclass and it would inherit it instead. However, if you do this, it looks like it would behave incorrectly and always read the `EXPECTED_NUMBER_OF_COLUMNS` constant from the superclass (https://www.ruby-forum.com/t/class-inheritance-and-class-constants/210683)

So personally I would just get rid of the constants and put the number 2 or 3 directly in the methods.

KatyTaylor (approved):
Approving, although I have a code style comment.

app/models/labware_creators/common_file_handling/csv_file/row_for_tube_rack_with_rack_barcode.rb
andrewsparkes: Ok will change it.
It does do that inheritance thing with the constants, found that out the hard way before. very confusing.

Issue #1529, opened 2024-01-05
DPL-1037 Handle 'EMPTY' values from tube rack scan


KatyTaylor:
**User story**

> We have finally set up our tube rack barcode reader and I would just like to make a slight change to the expected tube scan layout. Instead of 'NO READ' values for empty tube positions in the rack we get 'EMPTY' with our version of the software (attached is an example). Would it please be possible for you to amend the LIMS at the tube rack scan upload stage to accommodate this?

Refers to the scRNA Core Cell Extraction pipeline, at the step where LCA Blood Bank plate is transferred to LCA Bank Seq and LCA Bank Spare tubes. See `NO_TUBE_TEXTS = ['NO READ', 'NOSCAN'].freeze` in `app/models/labware_creators/plate_split_to_tube_racks/csv_file.rb`

**Who are the primary contacts for this story**
Abby, Katy

**Who is the nominated tester for UAT**
Abby

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Tube rack scan upload must treat 'EMPTY' values in the same way as 'NO READ' values.

KatyTaylor:
> @KatyTaylor @andrewsparkes Looking at [Katy's comment in an earlier story](https://github.com/sanger/limber/pull/1666#issuecomment-2045012900), and inspecting the code, as well as uploading a test file - it seems that this has already been implemented. Would you please review and confirm?

Yeah it looks like it's in common code, and sounds like you've been pretty thorough in checking. I think we can put this into Done.

andrewsparkes:
I can confirm the file handling has already been changed to include EMPTY as a possible value to indicate a missing tube. I did it when I did the refactoring.

The second requirement is to add EMPTY as an option in the help text on the relevant views where the files are uploaded, e.g. I can see it included in these 2 views:
app/views/plate_creation/multi_stamp_tubes_using_tube_rack_scan.html.erb
app/views/tube_creation/plate_split_to_tube_racks.html.erb

If those are the views used here (please check) then it's done.

PR #1666, opened 2024-04-08
scRNA UAT fix - added EMPTY to options for missing tubes

andrewsparkes:
Added EMPTY to the list of possible values indicating an empty position in a tube rack scan file


KatyTaylor:
Am I right in saying that this also ticks off this story - https://github.com/sanger/limber/issues/1529 ?

It's the same requirement, but for the rack scan earlier in the pipeline.

KatyTaylor (approved):

PR #1668, opened 2024-04-09
Y24-028 Targeted Nanoseq - fix to allow metadata update

### LINKED ISSUE #1628: Y24-028 [BUG] Targeted NanoSeq - Fix metadata insert issue preventing creation of additional child plates from the LTN AL Lib plate
**Describe the bug**
It is not currently possible to create any additional child LTN AL Lib Dil plates from an LTN AL Lib parent plate in the Targeted NanoSeq pipeline due to a limitation in the metadata creation code. It throws an exception when it detects a duplicate key on attempting insert. This prevents users from looping back to the LTN AL Lib plates and creating additional child plates (desired normal behaviour).

**RT Ticket Number**
NA

**To Reproduce**
1. Attempt to make more than one child plate from an LTN AL Lib plate in Limber
2. Get errors regarding duplicate metadata
3. In addition, child plate is made anyway, but binning is done incorrectly so needs to be cancelled

**Expected behaviour**
Having discussion about this, see research story https://github.com/sanger/limber/issues/1623

Short term fix for this story: 
Allow request metadata to be overridden. See code in the create_request_metadata method in file `app/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq.rb` which expects to insert fresh keys. We should check and overwrite the value if the key already exists. 
This is not ideal as it wipes out the record for previous runs if they loop without a new submission, but users are Ok with that for short term. Longer term better to have a new submission with different requests for follow up runs, on just the samples that they want to move forward, plus allowing partial transfers in the labware creator. See research story above.
Possibly in the short term we could store history by re-writing each of the metadata to be overwritten with a prefix key e.g. date. Example 'original_plate_barcode' could become '20240321_original_plate_barcode'.

**Screenshots**
NA

**Additional context**
We currently make 9 metadata records for each well with a sample on the plate. This is already slow so adding additional queries to check whether each metadata key exists before creating or updating it will slow it further. Check it's not going to time out for a full plate.

Longer term we should only be uploading data for rows they want to take forward, and only recording metadata for those rows (a partial transfer to the child plate), but this is not the case at the moment and the code is writing metadata for all rows with samples.

andrewsparkes:
Closes #1628 

Adds check for whether metadata with same key already exists on the request, if it does update it (if value different than current).

### DIFF
--- a/app/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq.rb
+++ b/app/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq.rb
@@ -50,37 +50,42 @@ module LabwareCreators
     # NB. makes assumption that metadata from previous iterations can be safely overwritten
     def create_or_update_request_metadata(request, request_metadata, child_well_location)
       request_metadata.each do |metadata_key, metadata_value|
-        # first select the polymetadatum by request id and key, if it exists, then update it
-        existing_metadata_v2 =
-          Sequencescape::Api::V2::PolyMetadatum.find(key: metadata_key, metadatable_id: request.id).first
+        existing_metadata_v2 = find_existing_metadata(metadata_key, request.id)
 
         if existing_metadata_v2.present?
-          next if existing_metadata_v2.value == metadata_value # no need to update if the value is the same
-
-          # update the existing metadata
-          next if existing_metadata_v2.update(value: metadata_value)
-
-          raise StandardError,
-                "Existing metadata for request (key: #{metadata_key}, value: #{metadata_value}) " \
-                  "could not be updated for request at child well location #{child_well_location}"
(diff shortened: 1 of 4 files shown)


app/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq.rb
KatyTaylor, line 1: Think the `# update fields on each well with various metadata` comment is misleading as we're no longer updating fields on `Well`.

KatyTaylor (approved):

app/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq.rb
andrewsparkes: changed, thx

Issue #1444, opened 2023-10-31
DPL-971 Create 007c version of 'Bespoke Chromium 5 prime' pipelines


KatyTaylor:
**User story**
Duplicate the 'Bespoke Chromium 5 prime' pipeline in Limber, but with new purposes containing the word 'GEM-X'. This is so we can make the change in https://github.com/sanger/limber/issues/1397 in just the GEM-X version, and so that plates going down the GEM-X route can be distinguished visually from plates going down the standard route. In the future, we will probably want 'GEM-X' versions of the other Chromium pipelines (3', BCR, TCR), but just make the 5 prime version for now.

**Who are the primary contacts for this story**
Katy, Lesley, Abby

**Who is the nominated tester for UAT**
It can be tested along with https://github.com/sanger/limber/issues/1397 when that's ready.

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Replace references to 'HT' in new aggregation pipeline, purposes, submission templates etc. with references to 'GEM-X' - see this PR - https://github.com/sanger/limber/pull/1673/files - for the scope.
- [ ] 'Bespoke Chromium 5p' pipeline is duplicated, and called 'Chromium GEM-X 5p'.
  - [ ] Purposes are:
      - [x] LRC GEM-X 5p Cherrypick (new)
      - [x] LRC GEM-X 5p Dil (new) [We named this as LRC GEM-X 5p GEX Dil]
      - [x] LRC GEM-X 5p Frag 2XP (new)
      - [x] LRC GEM-X 5p LigXP (new)
      - [x] LRC GEM-X 5p PCR 2XP (new)
  - [x] Library type is 'Chromium single cell 5 prime GEM-X v2' (new)
  - [x] Request type is 'limber_chromium_gem_x' (new) [We named it as limber_scrna_core_chromium_gem_x_5p]
  - [ ] New submission template - 'Limber - GEM-X Chromium' (plus standard variants including multiplexing and sequencing) [We named this as Limber-Htp - scRNA Core Chromium GEM-X]
- [ ] 'Bespoke Chromium 5p MX' multiplexing pipeline is duplicated, and called 'Chromium GEM-X 5p MX'. [See config/pipelines/bespoke_chromium_5p.yml in Limber]
  - [x] Purposes (all new except final) are:
      - LRC GEM-X 5p GEX PCR 2XP
      - LRC GEM-X 5p GLibPS
      - LRC GEM-X 5p Pool Norm
[...]

PR #1673, opened 2024-04-19
DPL-924 - scRNA - added configuration and labware creator variant for makng the cDNA Dil plate

### LINKED ISSUE #1397: DPL-924 10X PCR binning strategies change
**User story**
As a user of the scRNA Core pipeline, I would like Limber to generate a robot driver file for the "10X HT 5' Cherrypick --> 10X HT 5' cDNA Dil" transfer (exact naming to be finalised).

This should be similar to the current 10x pipeline (“LBC Cherrypick --> LBC 5p GEX Dil” transfer/ step to ‘Download Cherrypick to Sample Dilution CSV’) however the current process sets volumes to pick based on concentration and bins samples into either 12 or 14 PCR cycles. For the scRNA pipeline, the volume to pick should be set at 20ul and the range of cycles to bin into should be increased (see additional context for rules).

Lesley, Abby

Lesley, Abby

**Acceptance criteria**
- [x] Hamilton robot driver file can be downloaded, to tell the robot which transfers to make, between the "10X HT 5' Cherrypick" and “10X HT 5' cDNA Dil” plates.

Driver file example for existing pipeline, showing layout:
image

- [x] The samples in the "10X HT 5' Cherrypick" plate should be binned for PCR into different columns across the “10X HT 5' cDNA Dil” plate according to their concentrations (see table in ‘Additional context’ for details).
- [x] The new positions should be displayed to the user in Limber on the “10X HT 5' cDNA Dil” plate page and ordered so that each column only allows one number of PCR cycles (see Additional context’ for an example in a pre-existing line).
- [x] The new well locations should be captured in the ‘Destination Plate Well’ column of the driver file.
- [x] The ‘Sample Vol’ column should be set to ‘20’ and ‘Dilution Vol’ to ‘0’ for all samples.
- [x] Integration Suite test is updated

**Dependencies**
Blocked by:
- https://github.com/sanger/limber/issues/1444 

**Additional context**
Table of corresponding sample concentrations and PCR cycles. **We still need to determine the exact cycle numbers where a range is given. I will update you once we know.**
image

Image of an existing line sharing similar logic:
image

andrewsparkes:
Closes #1397 

Adds labware creator for creating the cDNA Dil plate.


config/purposes/scrna_core_cdna_prep.wip.yml
```diff
@@ -112,3 +112,73 @@ LRC HT 5p cDNA PCR XP:
...
+  :presenter_class: Presenters::ConcentrationBinnedPlatePresenter
+  :creator_class: LabwareCreators::ConcentrationBinnedSingleCellDilPlate
+  :dilutions:
+    :source_volume: 20
+    :diluent_volume: 0
+    :bins:
+      - max: 5.0
+        pcr_cycles: 16
+        colour: 1
+      - min: 5.0
```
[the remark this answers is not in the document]
andrewsparkes: I think the idea was to ensure you get the same colour (for say 14 PCR cycles) each time, rather than it dynamically assigning numbers depending on the binning range in that particular plate, so you'd need some config or table somewhere that persisted that relationship (PCR cycles to colour). The presenter uses the purpose config as well as the binning calculation code in the labware creator. And the binning ranges and PCR cycles could change so made sense it be in one place.
I get the point though.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
```diff
@@ -26,3 +26,24 @@ scRNA Core cDNA Prep Input:
     LRC HT 5p Chip: LRC HT 5p GEMs
     LRC HT 5p GEMs: LRC HT 5p cDNA PCR
     LRC HT 5p cDNA PCR: LRC HT 5p cDNA PCR XP
+scRNA Core Chromium HT Aggregation:
+  pipeline_group: scRNA Core cDNA Prep
```
KatyTaylor, line 30: Can we have a new 'pipeline group' called 'scRNA Core Library Prep'? We're out of the 'cDNA Prep' phase here. (Or, two new ones - one for aggregation and one for library prep)

They're just string values so just need to change it here, and in the below pipeline.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
```diff
@@ -26,3 +26,24 @@ scRNA Core cDNA Prep Input:
     LRC HT 5p Chip: LRC HT 5p GEMs
     LRC HT 5p GEMs: LRC HT 5p cDNA PCR
     LRC HT 5p cDNA PCR: LRC HT 5p cDNA PCR XP
+scRNA Core Chromium HT Aggregation:
```
KatyTaylor, line 29: When we came up with the naming before, we didn't include 'scRNA Core'.

See all names planned out here: https://github.com/sanger/limber/issues/1444  

Don't feel strongly about it though. We did include '5p' though, which seems like it could matter.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
```diff
@@ -26,3 +26,24 @@ scRNA Core cDNA Prep Input:
...
+  pipeline_group: scRNA Core cDNA Prep
+  filters:
+    request_type_key: limber_scrna_core_chromium_ht_aggregation
+  relationships:
+    # NB. alternate faculty input plate route included here
+    # TODO: do we need to allow skipping or bypassing the Aggregate plate? Config allows for a 2-step
+    # aggregation process, but this is not always necessary depending on the number of samples
+    # selected from the source plates. i.e. if 10 or fewer source plates are needed you could make the
+    # cherrypick plate directly from the source plates.
```
KatyTaylor, line 38: Feels like this comment should be captured in a backlog story or similar, instead of in the code. It applies to the existing aggregation pipeline as well as this one.

KatyTaylor (commented):
Haven't finished looking but stopping for now so I'll add these comments

config/purposes/scrna_core_cdna_prep.wip.yml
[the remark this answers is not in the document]
KatyTaylor: Ah OK, fair enough!

yoldas:
Is it possible to extract the TODOs from the code into a separate story?

config/exports/exports.yml
```diff
@@ -188,3 +188,7 @@ hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools:
     - wells.qc_results
     - wells.transfer_requests_as_target.source_asset
   ancestor_purpose: LRC PBMC Defrost PBS
+hamilton_lrc_ht_5p_cherrypick_to_lrc_ht_5p_gex_dil:
```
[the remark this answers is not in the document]
andrewsparkes: The dil plate is the first plate in the GEX library prep part of the pipeline. i.e. it will branch in future at the parent cherrypick plate into GEX, TCR and BCR library preps. Up till the cherrypick it is 5p for scRNA, then we get more specific for the library preps. For phase 1 we just have GEX library prep. This purpose naming is the same in the Bespoke pipeline where GEX is also included in the purpose names.
So I think it is right.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
andrewsparkes: I added scRNA Core because it's going to be aggregating with scRNA Core plate purposes. It's not generic to just any aggregation. I feel this is correct in terms of Limber naming conventions.

The **5p** I was less sure about, and initially had it in there.
I think I will re-add it. Say we add a 3p chip later, this means we will need a separate aggregation pipeline for the 3p plate purposes.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
andrewsparkes: yes will change.
also wondered if they should be in separate config files, but will leave together for now, we can discuss.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
andrewsparkes: Yes, I can make one.
I recall a conversation with an SSR about the bespoke aggregation, and that they would create a cherrypick plate directly rather than use the aggregation pipeline if fewer sources were needed. So it might be that SS cherrypicking needs to allow picking from either LRC HT 5p cDNA Input or LRC HT 5p cDNA PCR XP plates into an LRC HT 5p Cherrypick directly.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
```diff
@@ -26,3 +26,24 @@ scRNA Core cDNA Prep Input:
...
+    LRC HT 5p cDNA Input: LRC HT 5p Aggregate
+    LRC HT 5p cDNA PCR XP: LRC HT 5p Aggregate
+    LRC HT 5p Aggregate: LRC HT 5p Cherrypick
+scRNA Core Chromium HT 5p:
+  pipeline_group: scRNA Core cDNA Prep
+  filters:
+    request_type_key: limber_scrna_core_chromium_ht
+    library_type: Chromium single cell 5 prime HT v2
+  relationships:
+    LRC HT 5p Cherrypick: LRC HT 5p GEX Dil
```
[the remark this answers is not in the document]
andrewsparkes: No I think GEX is correct from Dil plate on as we are now in GEX specific 5p library prep. The Cherrypick plate parent is the branch point for the different variants of 5p library prep (GEX, TCR, BCR). See bespoke pipeline for similar naming.

app/models/labware_creators/well_filter_allowing_partials.rb
```diff
@@ -15,7 +15,9 @@ def filter_by_state(requests)
   end
 
   def filter_requests(requests, well)
-    return nil if well.requests_as_source.empty?
+    # Line changed, previously was checking for well.requests_as_source.empty? but not sure why
+    # there was that limitation. Here requests contains all active requests for the well.
+    return nil if requests.empty?
```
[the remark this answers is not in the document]
andrewsparkes: I too am confused by this filtering.

If you look at extract_submission it attempts to get the submission id from the well.aliquot.request. In our plate well.aliquot.request is nil as we're on the submission plate not the child plate. To get submission id we'd have to look at the request in well.requests_as_source.
I think when we are in a well filter allowing_partials to move forward we are always on a submission plate. So if the well is to go forward it needs to have a request_as_source.
But I looked at the well.active_requests method and that does fetch requests_as_source too.
I think I'll change it back to `return nil if well.requests_as_source.empty?` as that's more specific for a partially submitted plate.

config/pipelines/high_throughput_scrna_core_cdna_prep.wip.yml
andrewsparkes: Added 5p

KatyTaylor (approved):

yoldas (approved):
I commented about the GEX part because of a discussion in 007 channel. Using GEX and GEM-X together may be an issue for the user but for now this PR looks good.

PR #1694, opened 2024-05-17
Y24-074 scRNA - pooling tab removed and samples tab enhanced

### LINKED ISSUE #1679: Y24-074 [BUG] Pooling tab improvements for "LRC PBMC Pools" plate
N.B. see update section at bottom.

**Describe the bug**
From Abby during UAT of the scRNA Core pipeline:

> I have attached the Limber page for an LRC PBMC Pools plate that has 21 samples pooled across 2 wells.
> The display under the pooling tab shows a 'pool' for every sample the plate contains i.e. 21 pools. Pools 3-21 do not exist though so the information is a bit redundant. What would be ideal would be if it displayed which sample well from the previous plate went into which pool. Is that possible?

See screenshot further down.

N.B. we have another bug open relating to this same part of the Limber UI - https://github.com/sanger/limber/issues/1667 

**To Reproduce**
1. Run the scRNA Core integration suite pipeline up to the LRC PBMC Pools plate. You might have to increase the numbers of samples being used so that there are enough samples to make >1 pool.
2. On the page for the LRC PBMC Pools plate, click on the pooling tab.
3. See redundant information as described above.

**Expected behaviour**
The information shown here should be useful. 
- See Abby's suggestion above (display source wells for a pool)
- Look into what information this tab is meant to display - I presume that it's not working correctly for this plate, maybe because these are not traditional 'pools', because the samples are not tagged.

**Screenshots**
image (1)

**Update**
On looking into this it appears the 'pooling' tab is intended to show the planned future pooling of the samples in the plate according to the submissions / requests set up by the SSR in the submission. It is NOT intended to show the contents for current pools on this plate. i.e. it is not suitable for this plate or pipeline, and was probably intended to show Multiplex pooling strategies for the older sequencing pipelines.

The 'samples' tab is probably better suited for the information that Abby is requesting. This tab is supposed to list the sample composition of each well. The current code will display each sample in a well as `<sanger_sample_id>-<species>`. 

Unfortunately both these fields were not being set on sample creation for tube manifests in the Sequencescape UAT action used for testing this pipeline, so the tab was not displaying anything except the list of occupied wells (fix for that 
created here - https://github.com/sanger/sequencescape/pull/4114). 
This sample information does not include parent well either, so we will likely need to override the sample tab content for this plate in the scRNA pipeline. 

The story then became partly a research story to investigate and document how the tabs are currently constructed and capture possible ways to improve these summary tabs so that they are only displayed if relevant to the labware.

The research output of the story is a document capturing this investigation. See here:
[Document](https://docs.google.com/document/d/1MJHUyrIwwWIgBAhiRFp-NlYzk9GkF2iPutnrlF7n_9Q/edit?usp=sharing)

andrewsparkes:
Closes #1679 

Adds new Presenter variant which:
- hides pooling tab
- enhances samples tab


app/models/presenters/pooled_wells_plate_presenter.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Presenters
+  #
+  # The PooledWellsPlatePresenter overrides the way the pooling tab is displayed.
+  #
+  #
+  class PooledWellsPlatePresenter < StandardPresenter
+    # The pooling tab is not relevant for this presenter as the wells are already pooled (tab shows future pooling
+    # by submission strategy)
+    self.pooling_tab = nil
```
yoldas, line 11: Is it supposed to be nil or empty string? Presenter sets this to empty string.

app/views/plates/_pooled_wells_samples_tab.html.erb
```diff
@@ -0,0 +1,11 @@
+<dl id="samples-information" class="card-body">
+  <% plate_presenter.wells.each do |well| -%>
+    <% next if well.aliquots.empty? %>
+    <dt class="col-sm-1"><%= well.location %> <%= state_badge(well.state) %></dt>
+    <dd class="col-sm-11"><ul class="list-inline">
+    <% well.upstream_wells.each do |parent_well| -%>
+      <li class="list-inline-item"><%= parent_well.aliquots.first.sample.sanger_sample_id %> (source:
```
andrewsparkes, line 7: each here instead of first

yoldas: parent wells have single aliquots for LRC PBMC Pools plate. It is OK for this story.

yoldas (approved):

PR #1711, opened 2024-05-24
Y24-035 tnanoseq update configs for pipeline

### LINKED ISSUE #1646: Y24-035 Targeted NanoSeq - update configuration for pipeline and add submission for ISC prep
**User story**
As PSD we would like to fix the targeted nanoseq pipeline configuration to better reflect the reality of what the lab users are doing, and allow them to loop back to the LTN AL Lib and process different subsets of samples as many times as they like.

Sara W and team
Andrew S

Sara W team

**Acceptance criteria**
- [x] update config for Targeted NanoSeq to better reflect reality 
    * only re-ISC end route is required, remove other duplex seq route
    * shearing may or may not be done at the beginning, leave the two routes for that part
- [x] allow the submission for re-ISC to be done on the LTN Lib PCR XP plate rather than require a cherrypick into an LB Lib PCR-XP plate - this is likely to require the plate to be added to the acceptable plate purposes for the re-ISC submission template (SS config change and rails console fix in databases), plus add the relationship for transfer from LTN Lib PCR XP plate into LB Lib PrePool plate
- [x] add an automated close off of the library prep requests at creation of the LTN Al Lib plate (use :state_changer_class: StateChangers::AutomaticPlateStateChanger and   :work_completion_request_type lines in the purpose config) - [Edit: maybe Ok to use Charge and Pass as currently. But we MUST close these before continuing.]
- [x] add an automated submission at creation of AL Lib Dil for a customer request 'name TBC' [Edit: see comments below - have decided manual is the way to go, given the submission the name: **Targeted NanoSeq ISC Prep**, but TBC]
- [x] automatically close the new customer request at creation of the LTN Lib PCR XP plate
- [x]  improve code that identifies correct request to use when inserting poly-metadata on creating the LTN Al Lib Dil plate, as it currently only looks in aliquot.request, needs to look in the receptacles.requests_as_source for the newly created active requests [CORRECTION: no change required as it looks in aliquot.request on the CHILD well, which is right]
- [x] update the integration suite test for this pipeline to match what the users are doing (subsets of samples and a loop back)

**Dependencies**
None

References
- #1628 
- 

**Additional context**
Currently the centre part of this pipeline where they loop back to the LTN AL Lib plate is not tracked with a submission. 
Plus all samples are tracked as moving from LTN Al Lib plate through to LTN Lib PCR XP despite only a subset physically being transferred through.
Plus they are editing the Hamilton driver file to remove the unneeded samples manually.

andrewsparkes:
Closes #1646 

Updates config to allow Targeted NanoSeq to use ReISC and to loop with submissions.


yoldas (approved):
The latest changes for allowing to leave rows in file but set pic_cycles to 1 to ignore them look good. I have inserted binding.pry in several places and inspected the changes as well. The update clearly excludes rows with pcr_cycles set to 1 before comparing active requests and customer file wells, for validations. There are enough code comments. 
Type info in comments, params, returns would make them better.

PR #1716, opened 2024-05-29
DPL-924 Changes to binning config based on R&D feedback

### LINKED ISSUE #1397: DPL-924 10X PCR binning strategies change

KatyTaylor:
Closes #1397 

- Changes to binning config based on Abby feedback during UAT of DPL-924
- The config may change again, but this is closer to correct
- Associated change in Int Suite - https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/130

### DIFF
--- a/config/purposes/scrna_core_cdna_prep.wip.yml
+++ b/config/purposes/scrna_core_cdna_prep.wip.yml
@@ -164,7 +164,7 @@ LRC HT 5p Cherrypick:
 # NB. For the following binning step the min and max amounts for the bins are the total amount of cDNA
 # in the well i.e. QC concentration * (source_volume + diluent_volume)
 # Here we are using a fixed source volume and there is no dilution (diluent_volume = 0)
-# TODO: R&D to confirm the ranges for each PRC cycle
+# TODO: R&D to confirm the number of PCR cycles for each range
 LRC HT 5p GEX Dil:
   :asset_type: plate
   :stock_plate: false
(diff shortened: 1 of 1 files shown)


yoldas (approved):
Looks good.

I have checked if the bin endpoints meet; they do. The start is inclusive in the code, end is exclusive. In the config, the first bin does not have min value (set to zero), which is right and the last bin does not have max value (set to infinity), which is right. 
per_cycles follow the pattern of odd numbers from 15 to 5, which looks like the averages of the table given in the story. The source sample volume is changed from 20.0 to 10.0 in the config.

PR #1718, opened 2024-06-03
1392 dpl 919 transfer pbmc pools to 10x chip

yoldas:
Closes #

- Add the labware creator `StampedPlateReorderingColumnsToRows` for the purpose `LRC GEM-X 5p Chip`
- Add new methods to WellFilter and Plate for accessing and sorting wells
- Update purposes, pipelines, request types, and drivers for GEM-X chip. 

Note: Check the `asset_shape_id` and `size` fields of the purpose with the name LRC GEM-X 5p Chip in the `plate_purposes` table of the Sequencescape database. The `asset_shape_id` should be the id of the record in `asset_shapes` with name ChromiumX and the size should be 8.


app/models/labware_creators/stamped_plate_reordering_columns_to_rows.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module LabwareCreators
+  # This class is a subclass of StampedPlate but it reorders the wells from
+  # columns to rows on the child plate. It compresses the wells in the child
+  # plate; there will not be any gaps corresponding to the source wells that
+  # were filtered out.
```
KatyTaylor, line 7: The question Andrew was asking in Slack is relevant here too - worth checking with Abby whether the 'compressing' is desired. If we're not making a driver file (which I think we're not), then the hardcoded robot method won't be able to handle compressing to the left of the chip.

KatyTaylor (approved):
Approved, although I have left one question.

app/models/labware_creators/stamped_plate_reordering_columns_to_rows.rb
yoldas: I asked the question in Slack channel. we can't preserve the gaps of the entire source plate (96-well plate to 8-well chip). We need to assume only the first column of the source plate is used if they ask for it. This is OK for Pools plate because of the pooling algorithm but maybe not for Pools Input. They need to add using only first column to their process and there needs to be a validation to enforce it. I am keeping the compression for now. 

Stamped plate reordering is written as a generic way; reordering hash needs to be built by taking all first-column wells instead of filtered wells if necessary. So it will be a single column to a single row transfer.

PR #1735, opened 2024-06-12
Dpl 920 transfer from chip plate to gems plate

### LINKED ISSUE #1393: DPL-920 Transfer from 10X chip to GEMs plate
**User story**
As a user of the scRNA Core pipeline, I would like LIMS support for the transfer from the 10X chip to the GEMs plate.

Katy, Lesley, Abby

Lesley, Abby

**Acceptance criteria**
- [x] Input chip is "LRC GEM-X 5p Chip"
- [x] Output plate is "LRC GEM-X 5p GEMs"
- [x] It is a straight stamp (although the row on the chip is transposed to a column on the plate)
- [x] Bed verification:
    - [x] Input chip on bed 15
    - [x] Output plate on bed 5
- [x] Integration Suite test is updated to include this transfer.

**Related issues**
Probably best to NOT do this in parallel with https://github.com/sanger/limber/issues/1392, as some work to do with implementing the new chip might be duplicated.

andrewsparkes:
Closes #1393 

Adds bed verifications

Adds labware creator to stamp and compress to top left


app/models/labware_creators/stamped_plate_compressed.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    # Returns the well object at a specified well location on a given plate.
+    #
+    # @param plate [Plate] the plate object
+    # @param well_location [String] the well location on the plate
+    # @return [Well] the well object at the well location on the plate
+    #
+    def get_well_for_plate_location(plate, well_location)
+      plate.wells.detect { |well| well.location == well_location }
+    end
```
yoldas, line 34: Recently we have noticed that we are repeating this method. It was refactored into a Plate method. 
This method definition can be removed and `child_plate.well_at_location(destination_location)` can be called in the request_hash method instead.

yoldas (approved):
Looks good. I have suggested a refactoring.

app/models/labware_creators/stamped_plate_compressed.rb
andrewsparkes: Thank you, have changed to use that.

yoldas (approved):

PR #1736, opened 2024-06-13
Y24-123: Cell counting - QC labels - quick fix

### LINKED ISSUE #1729: Y24-nnn Cell counting - QC labels - quick fix
**User story**
As a user of the scRNA Core pipeline, I would like the label printing feature on the cell counting screens to print off 4 QC labels every time (currently they dynamically decide how many to print based on plate occupancy).

Katy, Abby

Abby

**Acceptance criteria**
- [ ] Set of labels printed for LRC PBMC Bank plate includes main plate label, and QC1-4 labels.
- [ ] Set of labels printed for LRC PBMC Pools plate includes main plate label, and QC1-4 labels.

**Dependencies**
This story is a quick fix and will be superseded by another story, which is harder to implement but provides more flexibility (haven't written up yet).

**Additional context**
This work is part of the change in cell counting strategy - from getting the LIMS to generate the input files for the Cellaca instrument, to the LIMS instead generating driver files for the Hamilton robot transfer into the QC plates. The LIMS cannot know how many QC labels will be needed, because Cellaca plates will be reused, and the LIMS won't know how many free wells there are on the Cellaca plate being used for each transfer.

StephenHulme:
Closes #1729 

- Always create 4 labels independent of plate occupancy
- Add additional test to confirm this

### DIFF
--- a/app/models/labels/plate_label_quad_qc.rb
+++ b/app/models/labels/plate_label_quad_qc.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Plate label class to print off 4 QC labels for the cellaca process
+class Labels::PlateLabelQuadQc < Labels::PlateLabelBase
+  NUMBER_OF_LABELS = 4
+
+  def attributes
+    super.merge(barcode: labware.barcode.human)
+  end
+
+  # NB. reverse order so printed in correct sequence
+  def qc_label_definitions
+    Array.new(NUMBER_OF_LABELS) { |index| qc_label(index + 1) }.reverse
+  end
+
+  private
+
+  def qc_label(number)
+    {
+      top_left: date_today,
+      bottom_left: "#{labware.barcode.human} QC#{number}",
+      top_right: workline_identifier,
+      barcode: [labware.barcode.human, "QC#{number}"].compact.join('-')
+    }
+  end
+end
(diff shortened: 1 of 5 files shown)


andrewsparkes (commented):
Changing this class changes the label printing for the Cardinal pipeline as well as for the new scRNA pipeline (search for use of this label template plate_cellaca_qc), are we Ok with that? They specifically asked for it not to waste labels in that pipeline (labels are surprisingly expensive).

If we are ok with that then it looks good.

KatyTaylor:
> Changing this class changes the label printing for the Cardinal pipeline as well as for the new scRNA pipeline (search for use of this label template plate_cellaca_qc), are we Ok with that? They specifically asked for it not to waste labels in that pipeline (labels are surprisingly expensive).
> 
> If we are ok with that then it looks good.

Am I right in thinking it would reasonably easy to change it so we're not breaking Cardinal? New plate label subclass, override that method, change purpose config to point to new class...?

If so I think we should do that - even though it looks like the Cardinal pipeline might become obsolete, we don't know that for sure yet so makes sense to not break it.

andrewsparkes (approved):
Looks good.

yoldas (approved):
Looks good.

KatyTaylor (approved):

PR #1739, opened 2024-06-17
1653 y24 037 cell count spot checking

yoldas:
Closes #

- 

### DIFF
--- a/app/models/utility/cell_count_spot_checking.rb
+++ b/app/models/utility/cell_count_spot_checking.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+module Utility
+  # This class is used to select wells from a plate for spot checking.
+  class CellCountSpotChecking
+    attr_reader :plate, :ancestor_tubes
+
+    # Initializes the object with a plate and its ancestor tubes. It is used
+    # to pick wells from the plate for spot checking.
+    #
+    # @param plate [Plate] the plate associated with this instance
+    # @param ancestor_tubes [Array<Tube>] the ancestor tubes of the plate
+    def initialize(plate, ancestor_tubes)
+      @plate = plate
+      @ancestor_tubes = ancestor_tubes
+    end
+
+    # Selects at most the specified number of wells from the plate. If the
+    # number of ancestor tubes is less than the count, or it the count is
+    # not specified it selects one well per ancestor tube.
+    #
+    # @param count [Integer] the number of wells to select
+    # @return [Array<Well>] an array of selected Well objects
+    def select_wells(count = nil)
+      count ||= ancestor_tubes.size
+      select_wells_until([count, ancestor_tubes.size].min)
+    end
+
+    # Selects wells until the specified count is reached or all bins of wells
+    # are exhausted. The method ensures the selected wells that are distributed
+    # across the plate.
+    #
+    # @param count [Integer] number of wells to select
+    # @return [Array<Well>] an array of selected Well objects
+    def select_wells_until(count)
+      bins = prepare_bins(count)
+      selected = {}
+      process_bins(count, bins, selected) while selected.size < count || bins.any?(&:any?)
+      selected.values
+    end
+
+    # Processes bins to select wells until the specified count is reached. It
+    # iterates over each bin, pops the first well, and puts the UUID of its
+    # first sample into the selected hash if it is not already there. it stops
+    # if the selected size reaches the count.
+    #
+    # @param count [Integer] the number of wells to select
+    # @param bins [Array<Array<Well>>] the bins of wells to select from
+    # @param selected [Hash{String => Well}] tracks wells already selected
+    # @return [void]
+    def process_bins(count, bins, selected)
+      bins.each do |bin|
+        next if bin.empty?
+        well = bin.shift
+        uuid = well.aliquots.first.sample.uuid
+        selected[uuid] = well unless selected.key?(uuid)
+        break if selected.size >= count
+      end
+    end
+
+    # Prepares bins of wells by based on the specified count which is the
+    # number of wells to select. Empty wells are rejected. This is used for
+    # picking wells that are distributed across the plate.
+    #
+    # @param count [Integer] the number of wells to select
+    # @return [Array<Array<Well>>] bins of wells to cycle selection
+    def prepare_bins(count)
+      wells = plate.wells_in_columns.reject(&:empty?)
+      bin_size = (wells.length / count.to_f).ceil
+      wells.each_slice(bin_size).to_a
+    end
+  end
+end
--- a/config/purposes/scrna_core_cell_extraction.yml
+++ b/config/purposes/scrna_core_cell_extraction.yml
@@ -70,22 +70,6 @@ LRC PBMC Bank:
       id: hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare
     - name: 'Download PBMC Bank Tubes Content Report'
       id: pbmc_bank_tubes_content_report
-    - name: 'Download Cellaca Input QC1'
-      id: cellaca_input_file
-      params:
-        page: 0
-    - name: 'Download Cellaca Input QC2'
-      id: cellaca_input_file
-      params:
-        page: 1
-    - name: 'Download Cellaca Input QC3'
-      id: cellaca_input_file
-      params:
-        page: 2
-    - name: 'Download Cellaca Input QC4'
-      id: cellaca_input_file
-      params:
-        page: 3
   :qc_thresholds:
     viability:
       units: '%'
--- a/config/purposes/scrna_core_cell_extraction.yml
+++ b/config/purposes/scrna_core_cell_extraction.yml
@@ -70,6 +70,10 @@ LRC PBMC Bank:
       id: hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare
     - name: 'Download PBMC Bank Tubes Content Report'
       id: pbmc_bank_tubes_content_report
+    - name: 'Download Hamilton LRC PBMC Bank plate to Celleca Routine Selection'
+      id: hamilton_lrc_pbmc_bank_to_celleca_routine_selection
+    - name: 'Download Hamilton LRC PBMC Bank plate to Celleca Occasional Selection'
+      id: hamilton_lrc_pbmc_bank_to_celleca_occasional_selection
   :qc_thresholds:
     viability:
       units: '%'
--- a/app/views/exports/hamilton_lrc_pbmc_bank_to_celleca_occasional_selection.erb
+++ b/app/views/exports/hamilton_lrc_pbmc_bank_to_celleca_occasional_selection.erb
@@ -1,0 +1,14 @@
+<%=
+  CSV.generate_line [
+    'Plate Barcode',
+    'Well Position',
+    'Vac Tube Barcode',
+    'Well Name',
+    'Sample Name'
+],
+row_sep: ''
+%>
+
+<% 
+
+%>
--- a/app/views/exports/hamilton_lrc_pbmc_bank_to_celleca_routine_selection.erb
+++ b/app/views/exports/hamilton_lrc_pbmc_bank_to_celleca_routine_selection.erb
@@ -1,0 +1,27 @@
+<%=
+  CSV.generate_line [
+    'Plate Barcode',
+    'Well Position',
+    'Vac Tube Barcode',
+    'Well Name',
+    'Sample Name'
+],
+row_sep: ''
+%>
+binding.pry
+CellCountSpotChecking.new(@plate, @ancestor_tubes)
+    .select_wells(@number_of_wells)
+    .each do |well|
+      sample = well.aliquots.first.sample
+      puts CSV.generate_line [
+        @plate.labware_barcode.human,
+        well.position,
+        @ancestor_tubes[sample.uuid].labware_barcode.human,
+        well.name,
+        sample.name
+      ],
+      row_sep: ''
+    end
+<% 
+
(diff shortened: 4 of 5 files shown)


app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,83 @@
...
+      # or a Hash with includes and excludes keys for states.
+      # @param default [Boolean] The default return value if no specific
+      # condition is met.
+      # @return [Boolean] Returns true if the feature can be enabled based on
+      # the states parameter, otherwise returns the default.
+      #
+      # :reek:BooleanParameter
+      # :reek:ManualDispatch
+      # :reek:TooManyStatements
+      def can_be_enabled?(states, default: true)
```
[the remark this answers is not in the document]
yoldas: Won't fix. Putting every line in a separate method will make it difficult to follow the algorithm.

app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,77 @@
...
```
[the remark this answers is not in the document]
yoldas: Won't fix. Putting every line in a separate method will make it difficult to follow the algorithm.

config/purposes/scrna_core_cell_extraction.yml
```diff
@@ -62,30 +62,25 @@ LRC PBMC Bank:
...
   :label_template: plate_quad_qc # creates QC4 barcodes
   :file_links:
     - name: 'Download Hamilton LRC Blood Bank plate to LRC PBMC Bank plate CSV'
       id: hamilton_lrc_blood_bank_to_lrc_pbmc_bank
     - name: 'Download Hamilton LRC PBMC Bank plate to LRC Bank Seq and Spare tubes CSV'
       id: hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare
+      # states can be specified as string, symbol, or array values to set the
+      # includes or hash with includes/excludes keys and string, symbol, or
+      # array values. In this particular case, we want to enable the download
+      # link only when the plate is in the passed and qc_complete states.
```
KatyTaylor, line 75:
```
states can be specified as string, symbol, or array values to set the
includes or hash with includes/excludes keys and string, symbol, or
array values.
```
This is generic documentation - could go in `docs/purposes_yaml_files.md` ?

```
In this particular case, we want to enable the download link only when the plate is in the passed and qc_complete states.
```
I think this is obvious from the code but leave it in if you like as I obviously have more Limber context than some future developers might!

config/purposes/scrna_core_cell_extraction.yml
```diff
@@ -62,30 +62,25 @@ LRC PBMC Bank:
...
-        page: 2
-    - name: 'Download Cellaca Input QC4'
-      id: cellaca_input_file
-      params:
-        page: 3
+      states: [passed, qc_complete]
+    - name: 'Download Cellaca LRC PBMC Bank Hamilton 4 Count CSV'
+      id: hamilton_lrc_pbmc_bank_to_cellaca_4_count
+    - name: 'Download Cellaca LRC PBMC Bank Hamilton ALL Count CSV'
+      id: hamilton_lrc_pbmc_bank_to_cellaca_all_count
```
KatyTaylor, line 83: Maybe Abby has reviewed the names already but I think it's unnecessary to include `LRC PBMC Bank` in the button names - because we already know we're on that particular plate page. So maybe just `Download Cellaca 4 Count CSV`

app/models/utility/cell_count_spot_checking.rb
```diff
@@ -0,0 +1,119 @@
...
+    attr_reader :plate, :ancestor_tubes
+
+    # Initializes the object with a plate and its ancestor tubes. It is used
+    # to pick wells from the plate for spot checking.
+    #
+    # @param plate [Plate] the plate associated with this instance
+    # @param ancestor_tubes [Array<Tube>] the ancestor tubes of the plate
+    def initialize(plate, ancestor_tubes)
+      @plate = plate
+      @ancestor_tubes = ancestor_tubes
```
KatyTaylor, line 15: You shouldn't have to retrieve the ancestor tubes, because the Vac tube barcode is saved in the `sample_metadata` `supplier_name`. (specified in docs so should be consistent https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/scRNA+Core+Pipeline+Documentation)

app/models/presenters/banking_plate_presenter.rb
KatyTaylor, line 1: Could this be a generic 'buttons_by_state_presenter'? Or even better, can we add the button / state functionality to the default presenter?

app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,83 @@
...
+      def parse_tokens(tokens)
+        case tokens
+        when String, Symbol
+          [tokens.to_s] # single
+        when Array
+          tokens.map(&:to_s) # multiple
+        else
+          [] # none
+        end
+      end
```
KatyTaylor, line 80: Just opinion - I find this method a bit tricky to read - perhaps a bit less abstraction might be easier here?

dasunpubudumal: I gave Copilot this function and asked to optimise it. This is what it came up with:

```ruby
def parse_tokens(tokens)
  Array(tokens).map(&:to_s).compact
end
```

Did the same in [GPT-4o](https://openai.com/index/gpt-4o-and-more-tools-to-chatgpt-free/). It gave me this: 

```ruby
def parse_tokens(tokens)
  Array(tokens).map(&:to_s)
rescue NoMethodError
  []
end
```

Both are wrong 😆

yoldas: It is a simple method to accept symbol, string, or array VALUE in cases:
states: VALUE
In this case, it will be assumed that only the state name inclusions are specified

includes: VALUE 
The inclusions are specified separately under states. The method will collect the state names for inclusions

excludes: VALUE
The exclusions are specified separately under states. The method will collect the state names for exclusions

If there is no such value, it also collects an empty array. 

The method was named as parse_tokens to mean that state name words are collected whatever form they are given. Because the method name sounds too abstract and there is not enough comment to explain the details, it was responded with a code-review comment.

I will change the name and add comments. The method is just so useful. I won't change or throw away.

app/models/presenters/banking_plate_presenter.rb
yoldas: I have added included the FeatureInStates concern to StandardPresenter, TubePresenter and TubeRackPresenter because they are where a state machine is available. I have updated their csv_file_links methods for filtering the download links.

app/models/utility/cell_count_spot_checking.rb
yoldas: As requested, I have updated several places now so it relies on sample_metadata supplier name for Vac tube barcodes, instead of ancestor_tubes via sample uuid access. Also, ancestor_tubes has been removed from export includes to stop retrieving them.

app/models/utility/cell_count_spot_checking.rb
```diff
@@ -0,0 +1,119 @@
...
+    # @param count [Integer] number of wells to select
+    # @return [Array<Well>] an array of selected Well objects
+    #
+    # :reek:FeatureEnvy
+    def select_wells_until(count)
+      bins = prepare_bins(count)
+      selected = {}
+      process_bins(count, bins, selected) while selected.size < count && bins.any?(&:any?)
+      selected.values.sort_by(&:coordinate)
+    end
```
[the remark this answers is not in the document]
yoldas: It uses sample_metadata supplier name for tracking the origin vac tubes now. Relevant line is selected[barcode] assignment in the process_bins(count, bins, selected) method on the hash reference passed by the select_wells_until(count) method.

app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,98 @@
...
+      # or a Hash with includes and excludes keys for states.
+      # @param default [Boolean] The default return value if no specific
+      # condition is met.
+      # @return [Boolean] Returns true if the feature can be enabled based on
+      # the states parameter, otherwise returns the default.
+      #
+      # :reek:BooleanParameter
+      # :reek:ManualDispatch
+      # :reek:TooManyStatements
+      def can_be_enabled?(states, default: true)
```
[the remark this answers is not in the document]
yoldas: Won't fix. Putting every line in a separate method will make it difficult to follow the algorithm.

app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,98 @@
...
+        current_state = state.to_s
+        return default if current_state.blank? # No current state.
+
+        includes, excludes = collect_includes_and_excludes(states)
+
+        # Exludes take precedence over includes.
+        return false if excludes.present? && excludes.include?(current_state)
+
+        # If includes is present, the current state must be in the list.
+        return includes.include?(current_state) if includes.present?
```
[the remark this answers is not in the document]
yoldas: Won't fix. They are carefully placed critical guard statements.

app/models/concerns/presenters/statemachine/feature_in_states.rb
```diff
@@ -0,0 +1,98 @@
...
+      def can_be_enabled?(states, default: true)
+        return default unless respond_to?(:state) # No state method.
+        return default if states.blank? # No states to check.
+
+        current_state = state.to_s
+        return default if current_state.blank? # No current state.
+
+        includes, excludes = collect_includes_and_excludes(states)
+
+        # Exludes take precedence over includes.
```
KatyTaylor, line 37:
```suggestion
        # Excludes take precedence over includes.
```

app/models/presenters/standard_presenter.rb
```diff
@@ -8,8 +8,31 @@ module Presenters
...
+              [
+                :limber_plate,
+                :export,
+                { id: link.id, limber_plate_id: human_barcode, format: :csv, **link.params || {} }
+              ]
+            ]
+          end
+      links << ['Download Worksheet CSV', { format: :csv }] if csv.present?
+      links
+    end
```
KatyTaylor, line 36: Did you consider putting it in the PlatePresenter? There are lots of other Presenters that inherit from PlatePresenter that could potentially benefit from this.

KatyTaylor (commented):
I started re-reviewing but wasn't sure if it's up to date, as the binning algorithm parts look the same, and I thought you said you re-wrote them? thanks

app/models/concerns/presenters/statemachine/feature_in_states.rb
yoldas: The binning algorithm is the same. What I rewrote is how the vac tubes are tracked. Previously it was using ancestor_tubes relation, now it is using sample_metadata supplier_name as you suggested.

app/models/presenters/standard_presenter.rb
yoldas: I have put into places where
- a state machine provided state is available
- file_links are customised

tube, plate and tube rack presenters where the concern is included do those two.

yoldas: I think, a better way of doing this is including the FeatureInStates concern in the top level presenter actually. I will add one liner for the file_links customisations.

yoldas: I have included the FeatureInStates concern in Presenter, removed csv_file_links override from StandardPresenter and updated csv_file_links of PlatePresenter with the can_be_enabled?(states) call on the links.

app/models/utility/cell_count_spot_checking.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+module Utility
+  # This class is used to select wells from a plate for spot checking.
+  # NB. It assumes that the barcodes of the ancestor tubes are stored in the
+  # sample metadata supplier_name field.
+  class CellCountSpotChecking
+    attr_reader :plate, :ancestor_tubes
```
KatyTaylor, line 8: Is :ancestor_tubes still used?

app/models/utility/cell_count_spot_checking.rb
```diff
@@ -0,0 +1,144 @@
...
+    end
+
+    # Selects at most the specified number of wells from the plate. If the
+    # number of ancestor tubes is less than the count, or if the count is
+    # not specified, it selects one well per ancestor tube.
+    #
+    # @param count [Integer] the number of wells to select
+    # @return [Array<Well>] an array of selected Well objects
+    def select_wells(count = nil)
+      count ||= ancestor_tubes_size
```
KatyTaylor, line 34: I think that `||=` here has no benefit over `=`, because you're not saving it outside the scope of this method?

Normally it would be saved into an instance variable, and the method would be called more than once.

KatyTaylor (approved):
Left a couple of comments but approving as they are small.

app/models/utility/cell_count_spot_checking.rb
yoldas: It is used twice within the same scope. The first use is setting a default for count (all tubes if not given), the second is for honouring an explicit count if given by taking the min of count or ancestor_tubes_size. So, there is a performance benefit in caching it. That is why it is done. One could create another variable and store it separately as well but this shortens the code.

yoldas: I might have misunderstood it but still it is necessary. It is setting a default for the count if not given.

PR #1752, opened 2024-06-28
DPL-1031 - File name improvements

### LINKED ISSUE #1521: DPL-1031 File name improvements
**User story**
As a user of the scRNA Core Cell Extraction pipeline, I would like a number of improvements to be made to the robot driver file names, to make them easier to work with in the lab and reduce mistakes and manual workarounds.

Abby, Andrew, Katy

Abby

**Acceptance criteria**
Enhancements to the file naming config:
- [ ] Add an option to explicitly set the file name in `exports.yml`, so that the names can be more user friendly.
- [ ] Add an option to add the labware barcode at the start of the filename (the `include_barcode` option currently adds it to the end).
- [ ] Add an option to add the parent labware's barcode

Adjustments to existing file names using above enhancements:
- [ ] 'hamilton_lrc_pbmc_bank_to_lrc_bank_seq_and_spare' file needs to have a shorter name, so that once the labware barcode is added it is under 60 characters (Hamilton robot rejects longer names, so they have to manually adjust it currently).
- [ ] Give hamilton_lrc_blood_bank_to_lrc_pbmc_bank a shorter, more user friendly name too, to be consistent within the pipeline.
- [ ] Both of the above file names should begin with the LRC PBMC Bank plate barcode

**Additional context**
Excerpt from original request in Slack:
> We can manually change filenames in the meantime (i.e. when we do the training in January) so it doesn't need to be fixed before deployment today [13/12/23] necessarily but we will need to alter the naming at some point.

BenTopping:
Closes #1521 

- Extends exports.yml documentation
- Adds ability to specify independent filename for exports
- Adds ability to prepend barcodes in exports filename
- Adds ability to include parent labware barcode in exports filename

There are no independent unit tests, mainly because the exports_controller tests it by proxy and it is has a different style of testing. But I have tested manually with a few variations and it is working.

### DIFF
--- a/app/controllers/concerns/exports_filename_behaviour.rb
+++ b/app/controllers/concerns/exports_filename_behaviour.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Helper methods for the Exports controllers
+module ExportsFilenameBehaviour    
+    # Builds and returns the filename and file extension for the export
+    def set_filename(labware, page)
+      # The filename falls back to the csv template attribute if no filename is provided.
+      filename = export.filename&.fetch('name') || export.csv
+      filename = build_filename(filename, labware, page)
+      file_extension = export.file_extension || 'csv'
+      response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
+    end
+
+    private
+
+    def build_filename(filename, labware, page)
+        # Append or prepend the given barcodes to the filename if specified in the export configuration.
+        filename = handle_filename_barcode(filename, labware, export.filename&.fetch('labware_barcode', nil))
+        filename = handle_filename_barcode(filename, labware.parents&.first, 
+                                           export.filename&.fetch('parent_labware_barcode', nil))
+
+        # Append the page number to the filename if specified in the export configuration.
+        filename += "_#{page + 1}" if export.filename&.fetch('include_page', false)
+        filename
+    end
+
+    def handle_filename_barcode(filename, labware, options)
+        return filename if options.blank? || labware.blank?
+
+        barcode = labware.barcode.human
+        filename = "#{barcode}_#{filename}" if options['prepend']
+        filename = "#{filename}_#{barcode}" if options['append']
+        filename
+    end
+end
--- a/app/controllers/concerns/exports_filename_behaviour.rb
+++ b/app/controllers/concerns/exports_filename_behaviour.rb
@@ -1,35 +1,35 @@
 # frozen_string_literal: true
 
 # Helper methods for the Exports controllers
-module ExportsFilenameBehaviour    
-    # Builds and returns the filename and file extension for the export
-    def set_filename(labware, page)
-      # The filename falls back to the csv template attribute if no filename is provided.
-      filename = export.filename&.fetch('name') || export.csv
-      filename = build_filename(filename, labware, page)
-      file_extension = export.file_extension || 'csv'
-      response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
-    end
+module ExportsFilenameBehaviour
+  # Builds and returns the filename and file extension for the export
+  def set_filename(labware, page)
+    # The filename falls back to the csv template attribute if no filename is provided.
+    filename = export.filename&.fetch('name', nil) || export.csv
+    filename = build_filename(filename, labware, page)
+    file_extension = export.file_extension || 'csv'
+    response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
+  end
 
-    private
+  private
 
-    def build_filename(filename, labware, page)
-        # Append or prepend the given barcodes to the filename if specified in the export configuration.
-        filename = handle_filename_barcode(filename, labware, export.filename&.fetch('labware_barcode', nil))
-        filename = handle_filename_barcode(filename, labware.parents&.first, 
-                                           export.filename&.fetch('parent_labware_barcode', nil))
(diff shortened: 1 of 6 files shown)


docs/exports_yaml_files.md
```diff
@@ -121,6 +122,36 @@ Example in `app/views/exports/duplex_seq_pcr_xp_concentrations_for_custom_poolin
...
+  # Includes the labware's human barcode in the file
+  labware_barcode:
+    # Includes the name of the labware in the file
+    append: true
+    # Add the name to the end of the file
+    prepend: false
+  # Includes the parent labware's human barcode
+  parent_labware_barcode:
+    append: true
+    prepend: true
```
KatyTaylor, line 144: Are there any defaults here or, if you include `labware_barcode`, do you have to specify `true`/`false` for `append` AND `prepend` ?

BenTopping: Neither are required or have defaults. Ill update the comments to reflect that.

app/controllers/exports_controller.rb
```diff
@@ -72,10 +72,30 @@ def include_parameters
...
-    filename += "_#{@labware.human_barcode}" if export.filename['include_barcode']
+    # The filename falls back to the csv template attribute if no filename is provided.
+    filename = export.filename['name'] || export.csv
+    filename = build_filename(filename)
+    file_extension = export.file_extension || 'csv'
+    response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
+  end
+
+  def build_filename(filename)
+    # Append or prepend the give barcodes to the filename if specified in the export configuration.
```
KatyTaylor, line 83:
```suggestion
    # Append or prepend the given barcodes to the filename if specified in the export configuration.
```

app/controllers/exports_controller.rb
```diff
@@ -72,10 +72,30 @@ def include_parameters
...
+    # Append the page number to the filename if specified in the export configuration.
     filename += "_#{@page + 1}" if export.filename['include_page']
-    response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.csv\""
+    filename
+  end
+
+  def handle_filename_barcode(filename, labware, options)
+    return filename if options.blank? || labware.blank?
+
+    barcode = labware.barcode.human
```
KatyTaylor, line 95: Interested to see `labware.barcode.human` here, as @StephenHulme found recently that the `human` attribute wasn't present - see discussion here https://github.com/sanger/limber/pull/1746#discussion_r1657294204

Perhaps it's because Stephen was working with tubes and you're working with plates - maybe the API returns different things. Wonder if it would be worth doing something like 
```
 barcode = labware.barcode.human || labware.barcode.machine
```
in case human is blank? Although in that case, not sure if `labware.barcode.human` would return `nil` or blow up.

app/controllers/tubes/tubes_exports_controller.rb
```diff
@@ -55,10 +55,29 @@
...
+    # The filename falls back to the csv template attribute if no filename is provided.
+    filename = export.filename['name'] || export.csv
+    filename = build_filename(filename)
     file_extension = export.file_extension || 'csv'
-    filename += "_#{@labware.human_barcode}" if export.filename['include_barcode']
-    filename += "_#{@page + 1}" if export.filename['include_page']
     response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
   end
+
+  def build_filename(filename)
```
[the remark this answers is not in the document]
KatyTaylor: Yep, some duplication here. Did you consider refactoring?

app/controllers/exports_controller.rb
BenTopping: Good point, the reason I did this was because labware.human_barcode didnt exist on some of my tests but labware.barcode.human did. I haven't run into a nil case for it yet but happy to add some handling for it. Would a machine barcode be more useful than just showing nil? They can manually edit it if it there are any issues.

app/controllers/tubes/tubes_exports_controller.rb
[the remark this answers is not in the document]
BenTopping: Yep just moved it to a concern which prevents the duplication

app/controllers/exports_controller.rb
andrewsparkes: I guess from the users point of view the aim is to match the export file with the physical plate using it's barcode label, so unless what's in the filename matches what's displayed on the plate label (human readable) it is not useful. Important for driver files where they are loading a plate on the robot and using the file to drive transfers for it.
And they (mostly) don't transfer to/from tubes on robots.
Odd there is no human attribute on tube labware barcodes. Maybe we should fix that.

andrewsparkes (approved):
Looks good to me, useful enhancements, thx!

app/controllers/concerns/exports_filename_behaviour.rb
```diff
@@ -0,0 +1,35 @@
...
+    response.headers['Content-Disposition'] = "attachment; filename=\"#{filename}.#{file_extension}\""
+  end
+
+  private
+
+  def build_filename(filename, labware, page)
+    # Append or prepend the given barcodes to the filename if specified in the export configuration.
+    filename = handle_filename_barcode(filename, labware, export.filename&.fetch('labware_barcode', nil))
+    filename =
+      handle_filename_barcode(filename, labware.parents&.first, export.filename&.fetch('parent_labware_barcode', nil))
```
yoldas, line 20: I need to add the labware barcode to the export filename. Is this adding the labware's **parent** barcode instead?

PR #1756, opened 2024-07-01
Test fix cellaca plate label date test

BenTopping:
This change is causing CI issues currently.
See: https://github.com/sanger/limber/actions/runs/9744623445/job/26890652637

- Updates plate_label_cellaca_qc_spec to match label date with a date regex keeping it inline with other specs

### DIFF
--- a/spec/models/labels/plate_label_cellaca_qc_spec.rb
+++ b/spec/models/labels/plate_label_cellaca_qc_spec.rb
@@ -18,6 +18,7 @@ RSpec.describe Labels::PlateLabelCellacaQc, type: :model do
   describe '#qc_label_definitions' do
     subject(:qc_label_definitions) { label.qc_label_definitions }
     let(:label) { described_class.new(labware) }
+    let(:date_format) { /\A\s?\d{1,2}-[A-Z]{3}-\d{4}\z/ } # e.g., ' 4 JUL 2023' or '24 JUL 2023'
 
     context 'when creating the label of a full plate' do
       let(:labware) { create :v2_plate, pool_sizes: [96] }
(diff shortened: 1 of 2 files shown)


spec/models/labels/plate_label_cellaca_qc_spec.rb
```diff
@@ -18,14 +18,15 @@
   describe '#qc_label_definitions' do
     subject(:qc_label_definitions) { label.qc_label_definitions }
     let(:label) { described_class.new(labware) }
+    let(:date_format) { /\A\s?\d{1,2}-[A-Z]{3}-\d{4}\z/ } # e.g., ' 4 JUL 2023' or '24 JUL 2023'
```
sdjmchattie, line 21: I accept that this regex matches the intended formats, but it also matches, for example `' 12-JUL-2023'` when you didn't intend a space before the two digits.  I don't know if this is actually important, but a regex that only allows the space for a single digit date is `/\A(?:\s\d|\d\d)-[A-Z]{3}-\d{4}\z/`

Your examples don't actually match your regex which expects dashes and not spaces between the segments.

sdjmchattie (approved):
Approved, but you might want to modify your examples and consider changing the regex to better match your expectations.

BenTopping:
@StephenHulme 
My theory is this check only fails from the 1st -9th as its the leading zero thats causing the problems. From a quick look this test hasn't been in long so it could be it has never run from the 1st-9th before.
I think this is because the test date match is wrong, where it should be using the same format as [here](https://github.com/sanger/limber/blob/develop/app/models/labels/base.rb#L35). I ripped the regex from another label test but looking at some of the other tests perhaps using this date time structure would be the better fix ...

spec/models/labels/plate_label_cellaca_qc_spec.rb
BenTopping: Sure, I actually ripped this regex from another test, I think some of the other tests check the exact equals using the same datetime string from https://github.com/sanger/limber/blob/develop/app/models/labels/base.rb#L35. That might be the nicer way of doing it.

sdjmchattie (approved):

PR #1770, opened 2024-07-08
Y24 037 uat update 3 cell count spot checking

yoldas:
Closes #

- Remove the distribution algorithm because first wells or second wells for vac tubes are returned now
- Add download buttons for FIRST and SECOND counts
- Add methods for first replicates and second replicates


app/models/utility/cell_count_spot_checking.rb
```diff
@@ -137,6 +29,24 @@ def first_replicates
         .values
     end
 
+    # Returns the second replicate wells in the plate. It groups the filtered
+    # wells by the ancestor tube barcode and returns the second well in each
+    # group.
+    def second_replicates
```
andrewsparkes, line 35: What if there is no second replicate? Or is it fixed here and there is always a second replicate?
For the Seq/Spare tubes the number of replicates depended on ancestor volume available. Low volume, fewer replicates.

andrewsparkes: This code would skip if no replicate it looks like. That ok?

config/purposes/scrna_core_cell_extraction.yml
```diff
@@ -72,10 +72,10 @@ LRC PBMC Bank:
     - name: 'Download PBMC Bank Tubes Content Report'
       id: pbmc_bank_tubes_content_report
       states: [passed, qc_complete]
-    - name: 'Download Cellaca Max 4 Count CSV'
-      id: hamilton_lrc_pbmc_bank_to_cellaca_max_4_count
-    - name: 'Download Cellaca ALL Count CSV'
-      id: hamilton_lrc_pbmc_bank_to_cellaca_all_count
+    - name: 'Download Cellaca FIRST Count CSV'
```
andrewsparkes, line 75: I assume they proposed these names? Don't seem very descriptive but I guess they will have it written in their SOP what file they need when.

spec/views/exports/hamilton_lrc_pbmc_bank_to_cellaca_common.csv.erb_spec.rb
```diff
@@ -31,65 +31,44 @@
...
+  context 'with second replicates' do
+    let(:template) { 'exports/hamilton_lrc_pbmc_bank_to_cellaca_second_count' }
     it 'renders the expected content' do
       render(template: template)
       content = CSV.parse(rendered)
-      expect(content).to eq(expected_content(count))
+      selected_wells = Utility::CellCountSpotChecking.new(plate).second_replicates
+      expect(content).to eq(expected_content(selected_wells))
     end
   end
```
andrewsparkes, line 58: maybe add a test including one where there is no second replicate? (if this is appropriate)

andrewsparkes (commented):
Looks good (a lot simpler).
Just clarify that there will always be a second replicate, or if not what you should do.

app/models/utility/cell_count_spot_checking.rb
yoldas: Yes, that is OK. I will update the comment to clarify.

From the post that starts with "After our meeting today," at https://sanger-global.slack.com/archives/C05F21NKATD/p1720104757004039?thread_ts=1719846560.109819&cid=C05F21NKATD

"It would be the same story as before but to pick a replicate from **all LRC Blood Vacs** available rather than just 4. **If a Blood Vac only has 1 replicate, no well should be counted**.

config/purposes/scrna_core_cell_extraction.yml
yoldas: Yes, they wanted Hamilton to Cellaca download buttons look different from other buttons on the page and shorter. They also wanted the words FIRST and SECOND in caps. The words refer to "replicates" as they described them.

spec/views/exports/hamilton_lrc_pbmc_bank_to_cellaca_common.csv.erb_spec.rb
yoldas: Thanks. I will add a test for this case.

app/models/utility/cell_count_spot_checking.rb
```diff
@@ -51,7 +52,7 @@ def second_replicates
     #
     # @return [Array<Well>] the wells in the plate that are not empty
     def filtered_wells
-      @filtered_wells ||= plate.wells_in_columns.reject(&:empty?)
+      @filtered_wells ||= plate.wells.reject(&:empty?).sort_by(&:coordinate)
```
andrewsparkes, line 55: does that way of sorting work ok? still in columns and not A11 not before A2?
there's also `wells_with_aliquots` which does something similar

PR #1780, opened 2024-07-11
Y24 041 uat update 4 cell count spot checking

yoldas:
Closes #

- Filter out failed wells, which shifts the wells to the failed well positions in the downloads. 

### DIFF
--- a/app/models/utility/cell_count_spot_checking.rb
+++ b/app/models/utility/cell_count_spot_checking.rb
@@ -48,11 +48,14 @@ module Utility
 
     private
 
-    # Returns the wells in the plate that are not empty in column-major order.
+    # Returns the wells in the plate in column-major order. Empty and failed
+    # wells are filtered out. The filtering will make sure that in case a well
+    # is failed, the well following that will be used for that in the download
+    # for spot checking.
     #
     # @return [Array<Well>] the wells in the plate that are not empty
     def filtered_wells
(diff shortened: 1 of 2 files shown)


app/models/utility/cell_count_spot_checking.rb
```diff
@@ -48,11 +48,14 @@ def second_replicates
...
-    # Returns the wells in the plate that are not empty in column-major order.
+    # Returns the wells in the plate in column-major order. Empty and failed
+    # wells are filtered out. The filtering will make sure that in case a well
+    # is failed, the well following that will be used for that in the download
+    # for spot checking.
     #
     # @return [Array<Well>] the wells in the plate that are not empty
     def filtered_wells
-      @filtered_wells ||= plate.wells.reject(&:empty?).sort_by(&:coordinate)
+      @filtered_wells ||= plate.wells.reject { |well| well.empty? || well.failed? }.sort_by(&:coordinate)
```
andrewsparkes, line 58: are 'cancelled' wells possible? just wondering...

yoldas: That is a very good point. It was never mentioned in the criteria so far. I will ask and add it in a next iteration.

yoldas: Adding this comment here just in case: _"I don't think there's a GUI function to cancel a well, only fail. You can cancel the whole plate but then you wouldn't need the file."_

PR #1792, opened 2024-07-17
BUG - removed unneeded input_plate flag from config files

andrewsparkes:
Update to resource highlighted that input_plate flag being included in purpose config files was pointless as cannot be set through the API. Removed as unneeded. Those labwares that are inputs should be configured as such in the Sequencescape configuration in any case.

### DIFF
--- a/docs/purposes_yaml_files.md
+++ b/docs/purposes_yaml_files.md
@@ -41,7 +41,6 @@ LB Cherrypick:
   :asset_type: plate
   :stock_plate: true
   :cherrypickable_target: true
-  :input_plate: true
   :presenter_class: Presenters::StockPlatePresenter
 LB Shear:
   :asset_type: plate
@@ -111,7 +110,6 @@ Example plate:
   :asset_type: plate
   :stock_plate: false
   :cherrypickable_target: false
-  :input_plate: false
   :size: 96
   :presenter_class: Presenters::StandardPresenter
   :state_changer_class: StateChangers::DefaultStateChanger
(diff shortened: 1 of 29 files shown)


docs/purposes_yaml_files.md
```diff
@@ -143,20 +141,6 @@ in Sequencescape. Usually only true for the first plate in the pipeline.
...
-Boolean, indicates that the plate has the input_plate flag set
-in Sequencescape. Usually only true for the first plate in the pipeline. Also
-used to determine if the plate shows in the 'New Input Plates' inbox, and to
-determine which barcode gets shown on downstream ancestors
-
-```yaml
-:input_plate: false
-```
-
-Default: `false`
```
sdjmchattie, line 158: Is this default implicit?  I don't see any changes to the code that uses these `.yml` config files.

andrewsparkes: the purpose yml files themselves are used by a rake task `config:generate`

KatyTaylor (approved):
So this was causing the 404 error when running config:generate?

You said 'resource updated' - do we know what caused the issue exactly?

yoldas:
Have you executed integration suite on this? Particularly the Bioscan test. We had some reason for doing this:
```
LILYS-96 Stock:
  :stock_plate: true
  :input_plate: true

LBSN-96 Lysate:
  :stock_plate: true
  :input_plate: false
```

andrewsparkes:
Decided not to do this change at this point, and revert the SS change that made input_plate read only.

yoldas:
Explicit specification of createable fields is introduced on the resource, which missed input_plate. I think, input_plate is a settable attribute. The resource class is supposed to use the attribute writer on the underlying model.

PR #1797, opened 2024-07-19
Fix automatic state change on LRC GEM-X 5p cDNA PCR XP plate

KatyTaylor:
2 fixes based on UAT testing:

1. Fix automatic state change on LRC GEM-X 5p cDNA PCR XP plate
- It wasn't completing the requests when coming down the SeqOps cDNA Prep route
- It only worked for the Faculty input route
- This is because the only request type specified in the 'work_completion_request_type' config was 'limber_scrna_core_cdna_prep_input'
- Had to enhance the AutomaticStateChange functionality to allow multiple request types

2. De-duplicate requests in the requestsForWell function
- in the aggregation pipeline for scRNA Core, the plates that the submission is done on may contain multiple aliquots
- if this is the case, each aliquot in the well references the same request
- this method was pulling the request from each of the aliquots - presumably with the assumption that each aliquot would have a different request
- in our case, it created duplicates in the array
- there could also theoretically by duplicates if the same request was referenced on the aliquot and in requests_as_source (although I think this shouldn't happen)
- therefore I added de-duplication to the end of the function chain, based on request id

Closes #

- 

### DIFF
--- a/app/javascript/shared/wellHelpers.js
+++ b/app/javascript/shared/wellHelpers.js
@@ -33,8 +33,23 @@ const quadrantTargetFor = function (plateIndex, wellName, rowOffset, colOffset) 
   return wellCoordinateToName([destinationColumn, destinationRow])
 }
 
+// Given a well, return the requests that are relevant to it.
 const requestsForWell = function (well) {
-  return [...well.requests_as_source, ...well.aliquots.map((aliquot) => aliquot.request)].filter((request) => request)
+  const arr = [
+    // If the well is in the plate that the submission was done on,
+    // the requests originate in these wells and can be retrieved using requests_as_source.
+    ...well.requests_as_source,
+    // If the well in on a plate downstream of the one the submission was done on,
+    // the relevant request ids are stored on the aliquots.
+    ...well.aliquots.map((aliquot) => aliquot.request)
+  ].filter((request) => request)
+
+  // Turn the array into a Map and back again, to remove duplicate requests.
+  // Duplicates could arise if:
+  // a) there are multiple aliquots in a well that reference the same request, or
+  // b) the same request is present in both requests_as_source and aliquots
+  const mp = new Map(arr.map((request) => [request.id, request]));
+  return Array.from(mp.values());
 }
 
 const rowNumToLetter = function (value) {
--- a/app/javascript/shared/wellHelpers.js
+++ b/app/javascript/shared/wellHelpers.js
@@ -48,7 +48,7 @@ const requestsForWell = function (well) {
   // Duplicates could arise if:
   // a) there are multiple aliquots in a well that reference the same request, or
   // b) the same request is present in both requests_as_source and aliquots
-  const mp = new Map(arr.map((request) => [request.id, request]))
+  const mp = new Map(arr.map((request) => [request.uuid, request]))
   return Array.from(mp.values())
 }
 
--- a/app/javascript/shared/wellHelpers.js
+++ b/app/javascript/shared/wellHelpers.js
@@ -41,15 +41,15 @@ const requestsForWell = function (well) {
     ...well.requests_as_source,
     // If the well in on a plate downstream of the one the submission was done on,
     // the relevant request ids are stored on the aliquots.
-    ...well.aliquots.map((aliquot) => aliquot.request)
+    ...well.aliquots.map((aliquot) => aliquot.request),
   ].filter((request) => request)
 
   // Turn the array into a Map and back again, to remove duplicate requests.
   // Duplicates could arise if:
   // a) there are multiple aliquots in a well that reference the same request, or
   // b) the same request is present in both requests_as_source and aliquots
-  const mp = new Map(arr.map((request) => [request.id, request]));
-  return Array.from(mp.values());
+  const mp = new Map(arr.map((request) => [request.id, request]))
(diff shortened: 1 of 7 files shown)


app/javascript/shared/wellHelpers.js
```diff
@@ -33,8 +33,23 @@ const quadrantTargetFor = function (plateIndex, wellName, rowOffset, colOffset)
   return wellCoordinateToName([destinationColumn, destinationRow])
 }
 
+// Given a well, return the requests that are relevant to it.
 const requestsForWell = function (well) {
-  return [...well.requests_as_source, ...well.aliquots.map((aliquot) => aliquot.request)].filter((request) => request)
+  const arr = [
```
andrewsparkes, line 38: Feels like to me that surely there has to be a more succinct way of doing what's happening in this method(!)
Ask CoPilot? It's good at that.

andrewsparkes: But don't necessarily trust the first solution it comes up with...

KatyTaylor:
```
This is what it came up with, which isn't any more succinct, except for removing a lot of the comments :(

If the comments are too much, I can re-write them?

Interesting that it changes `.filter((request) => request)` to `.filter(Boolean); // Filter out falsy values`
I wasn't sure what that was doing, so I left it alone...

function getUniqueRequestsFromWell(well) {
  // Combine requests_as_source and requests from aliquots into one array
  const combinedRequests = [
    ...well.requests_as_source,
    ...well.aliquots.map(aliquot => aliquot.request),
  ].filter(Boolean); // Filter out falsy values

  // Use a Set to remove duplicates based on the request's uuid
  const uniqueRequests = Array.from(new Set(combinedRequests.map(req => req.uuid)))
    .map(uuid => combinedRequests.find(req => req.uuid === uuid));

  return uniqueRequests;
}
```

andrewsparkes: Nah comments are good.
agree for filter(Boolean) is not clear what that is doing

app/javascript/shared/wellHelpers.spec.js
```diff
@@ -15,4 +15,29 @@ describe('wellHelpers', () => {
...
+      uuid: '3',
+      position: { name: 'A1' },
+      aliquots: [
+        {
+          uuid: '4',
+          request: request1,
+        },
+        {
+          uuid: '5',
+          request: null,
```
andrewsparkes, line 33: not quite following why this is null
is it possible for one aliquot in a well to have a request and the other not?
also, is it possible for request1 to be both in requests_as_source (which it will be if the submission is made on this labware) AND in the aliquot (which is set only on children of the submitted well)?
maybe add some comments describing the scenario?

andrewsparkes: the real scenario was more that multiple aliquots in the well has the same aliquot.request yes?

spec/models/state_changers_spec.rb
```diff
@@ -125,6 +125,24 @@
           expect(work_completion_creation).to_not have_been_made
         end
       end
+
+      # The ability to have multiple request types in the config was added for scRNA Core pipeline.
+      # The expectation was that any one plate would only have one of the request types on it,
+      # so I haven't tested a plate with a mix of request types.
+      context 'when one of the multiple config request types matches the in progress submissions' do
```
andrewsparkes, line 132: It occurs to me that aggregation is like cherrypicking. So they may return to the same source and pick a different subset of wells (or even repeat pick some). So multiple requests_as_source in a well is possible. Only one should be active at a time though.
I think this is fine for work completion, as I think the filtering for active requests is happening on the SS side.
But for the de-duplication code further up, do we need to consider filtering requests_as_source for active requests?

andrewsparkes: what is the `requests_for_well` method used for again?

app/javascript/shared/wellHelpers.spec.js
KatyTaylor: You're right, it's not representing the real scenario, I was just trying to cover everything in one test (bad practice 😅 ). I'll split them out.

I don't think it's expected for the same request to be in both requests_as_source and on the aliquot... but I wouldn't be surprised if that happens in the future (or in some edge case now), so I thought I should account for it.

yoldas: requestsForWell filters falsy values out and the test is describing the requestsForWell function. Having a null request covers that case.

spec/models/state_changers_spec.rb
KatyTaylor: `requestsForWell` is used in `requestsFromPlates` in `requestHelpers.js`

That's then used in `MultiStamp.vue`, where it then calls `requestIsActive` on them.

So `requestsForWell` simply grabs the requests, and they're filtered later.

yoldas: requests_as_source implies transfer requests on receptacle I think.

yoldas (commented):
Enables multiple work completion request types. 
Extracts unique requests from well. 
Looks good. Waiting until finalising discussions.

KatyTaylor:
Merging to unblock @andrewsparkes , but will come back to improve the js tests

PR #1805, opened 2024-07-23
Y24-190-2: Use SS API v2 for barcode printers

### LINKED ISSUE #1803: Y24-190-2: Use SS API v2 endpoints for Chromium pipelines
This is a sub-story for updating all parts of Limber involved in Chromium pipelines in the integration suite so that they use Sequencescape API V2 endpoints.

## References

Parent: https://github.com/sanger/limber/issues/1772
Sibling: https://github.com/sanger/sequencescape/issues/4223

sdjmchattie:
Closes #1803 

- Use the new Barcode Printer API on Sequencescape v2.
- A huge number of updates to tests which were all expecting v1 endpoints to be hit.


spec/factories/barcode_printer_factories.rb
```diff
@@ -1,63 +1,22 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  # Build an API v1 barcode printer. You probably want to use: plate_barcode_printer or
-  # tube_barcode_printer instead
-  factory :barcode_printer, class: Sequencescape::BarcodePrinter, traits: [:api_object] do
-    json_root { 'barcode_printer' }
-
-    active { true } # Whether Sequencescape reports the printer as active or not
```
BenTopping, line 9: Do we want to keep active?

spec/features/charge_and_pass_libraries_spec.rb
```diff
@@ -21,7 +21,7 @@
   background do
     # We look up the user
     stub_swipecard_search(user_swipecard, user)
```
BenTopping, line 23: This spec also has a 'default_tube_printer' defined. Can that be removed since its not used?

BenTopping (approved):
I haven't pulled down and tested and I am not to familiar with limber barcode printing so probably worth another set of eyes.

spec/factories/barcode_printer_factories.rb
sdjmchattie: I don't think it's actually referred to anywhere inside Limber.  It certainly didn't break any unit tests.  I can't accurately test these changes as it stands because I still have loads of other v1 endpoints to unblock first.  Plus I can't actually print without the hardware 😆 It may be a hang over from some previous functionality.

app/helpers/barcode_labels_helper.rb
```diff
@@ -24,7 +24,7 @@ def barcode_printing_form(
   end
 
   def printers_of_type(printer_types)
-    @printers.select { |printer| printer_types.include?(printer.type.name) } # rubocop:todo Rails/HelperInstanceVariable
+    @printers.select { |printer| printer_types.include?(printer.barcode_type) } # rubocop:todo Rails/HelperInstanceVariable
```
andrewsparkes, line 27: checking that isn't meant to be barcode_type**.name**

andrewsparkes (approved):

app/helpers/barcode_labels_helper.rb
sdjmchattie: Thanks for checking, but no it's OK.  The name is implicitly what the barcode_type refers to in the API.

image

PR #1828, opened 2024-08-02
Y24-177 - scRNA - simplified aggregation submission template name and added plate scan validation

### LINKED ISSUE #1767: Y24-177 - Modify Aggregation submission
**Background**
We decided in a meeting with Abby & Liz H to make the Library Prep submission 'generic': the template name is:

`Limber-Htp - scRNA Core Library Prep`

rather than 

`Limber-Htp - scRNA Core GEM-X 5p Library Prep`

This mirrors the equivalent submission template for the Bespoke Chromium pipelines, which is:

`Limber-Bespoke - Chromium`

The downside is that there could be user error when making a submission, in that they could include a GEM-X 5p purpose, but use a GEM-X 3p library type, for instance (when other variants of the pipeline come in).

The upside is that we won't have proliferation of submission templates when other variants come in.

**This story**
I suggest we do the same for the aggregation step - make it more generic - for the same reasons discussed above, and to make it consistent with the library prep submission.

Katy, Andrew, Abby, Liz H

Abby (R&D), Liz H (SSR)

**Acceptance criteria**
- [x] Aggregation submission template is called `Limber-Htp - scRNA Core Aggregation` (I removed the word 'Chromium' as well as it seems redundant with 'scRNA Core')
- [x] Request type is changed accordingly
- [x] [Documentation](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/scRNA+Core+Pipeline+Documentation) is updated to include this submission, including submission template name, appropriate plate purposes to use and any other information you think is useful.

andrewsparkes:
Closes #1767 

- 


spec/models/labware_creators/ten_stamp_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(subject.acceptable_purposes).to eq(%w[Purpose1 Purpose2])
+    end
+  end
+
+  context 'when purpose_config does not have acceptable_purposes' do
+    let!(:purpose_config) do
+      create :aggregation_purpose_with_args_config,
+             name: child_purpose_name,
+             uuid: child_purpose_uuid,
+             acceptable_purposes: {}
```
yoldas, line 49: It is setting acceptable_purposes to an empty hash. Is this intended?

app/javascript/multi-stamp/components/MultiStamp.vue
```diff
@@ -257,7 +264,7 @@ export default {
       return [
         checkSize(12, 8),
         checkDuplicates(currPlates),
-        // checkExcess(this.excessTransfers)
```
KatyTaylor, line 260: Not sure what the story is here...

`checkExcess` seems to be defined in `plateScanValidators` and not used - so could probably be deleted along with its test?

It's a bit of a rabbit hole though as I wonder whether `this.excessTransfers` and related code should also be removed.

app/javascript/shared/components/plateScanValidators.js
```diff
@@ -375,6 +375,28 @@ const checkPlateWithSameReadyLibrarySubmissions = (cached_submission_ids) => {
...
 }
 
+// Returns a validator that ensures that the scanned plate does not have an unacceptable plate purpose.
+// plate: The current plate being scanned
+const checkForUnacceptablePlatePurpose = (acceptable_purposes) => {
+  return (plate) => {
+    if (!acceptable_purposes || acceptable_purposes.length == 0) {
+      // return valid if no acceptable purposes are provided
+      return validScanMessage()
+    } else if (acceptable_purposes.includes(plate.purpose.name)) {
```
KatyTaylor, line 385: Wondering whether we should guard against the plate purpose name not being available...

app/javascript/shared/components/plateScanValidators.js
```diff
@@ -375,6 +375,28 @@ const checkPlateWithSameReadyLibrarySubmissions = (cached_submission_ids) => {
   }
 }
 
+// Returns a validator that ensures that the scanned plate does not have an unacceptable plate purpose.
+// plate: The current plate being scanned
```
KatyTaylor, line 379: document the type and expected contents of the `acceptable_purposes` parameter?

app/javascript/multi-stamp/components/MultiStamp.vue
```diff
@@ -127,6 +127,10 @@ export default {
 
     // Default volume to define in the UI for the volume control
     defaultVolume: { type: String, required: false, default: null },
+
+    // Acceptable plate purpose names that can be used as source plates e.g. "['PurposeA', 'PurposeB']"
+    // See computed method for conversion to array
+    acceptablePurposes: { type: String, required: false, default: '[]' },
```
KatyTaylor, line 133: Do they have to be passed in as strings?

KatyTaylor: Is this `MultiStamp` component only used in situations where plates are scanned in? Or can it also be used for scanning tubes in? Think the former, because we've also got a component called `MultiStampTubes`?

Just checking because this new code assumes we're scanning plates, I think.

KatyTaylor (commented):
I've just left a few questions and comments.

spec/models/labware_creators/ten_stamp_spec.rb
andrewsparkes: No will change to array, thx.

andrewsparkes: Actually to nil. This test is for a config that does not have aceptable_purposes at all, so should get defaulted to empty array by the code.

app/javascript/multi-stamp/components/MultiStamp.vue
andrewsparkes: Reading the comment above the validator code, this check is intended to flag up when you are combining plates and there are more samples in the sources than will fit in the destination. e.g. plate 1 has 50, plate 2 has 50, it flags up that 4 in plate 2 will not fit in the destination (assuming 96).
Instead it now seems to just be comparing `validTransfers` to the number of destination plate wells to do that check in `this.excessTransfers`.
So I think the validator can be removed but not the excessTransfers.

app/javascript/shared/components/plateScanValidators.js
andrewsparkes: Done

app/javascript/multi-stamp/components/MultiStamp.vue
andrewsparkes: Yeah they are passed as string interpolations in the erb that passes the variable values to the Vue JS component.
See `app/views/plate_creation/multi_stamp.html.erb`

I believe it's just used for plates, as it has many references to plate in the template and methods. The tube version also calls a separate file for its scanValidations.

app/javascript/shared/components/Plate.vue
```diff
@@ -21,7 +21,7 @@
         <td v-for="column in columns" :key="column">
           <lb-well
             v-bind="wellAt(row, column)"
-            :tooltip_label="tooltip_label(row, column)"
+            :tooltip_labL="tooltip_label(row, column)"
```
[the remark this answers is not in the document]
andrewsparkes: changed, thx

app/javascript/shared/components/plateScanValidators.js
andrewsparkes: added, thx

KatyTaylor (approved):

Issue #1818, opened 2024-07-30
Y24-190-3: Use SS API v2 endpoints for Labware Creators


sdjmchattie:
This is a sub-story for updating Labware Creators so that they use Sequencescape API V2 endpoints.

## References

Parent: https://github.com/sanger/limber/issues/1772
Sibling: https://github.com/sanger/sequencescape/issues/4244

PR #1835, opened 2024-08-06
Y24-190-3: Use SS v2 for TagLayoutTemplates

sdjmchattie:
Closes #1818 

- Switch to using SS v2 for TagLayoutTemplate
- There is one remaining instance of v1 used for TagLayoutTemplate which I have been unable to work around yet.  I plan to fix this by also implementing TagLayout in SS v2 and then I can push the creation through the correct endpoint.


app/models/labware_creators/tagged_plate.rb
```diff
@@ -79,6 +79,10 @@ def convert_tag_plate_to_new_purpose
 
     def create_labware!
       create_plate! do |plate_uuid|
+        # TODO: {Y24-190} Work out a way to call the `create!` method on TagLayoutTemplate model in Sequencescape
+        #       using the V2 API. I think either we need to misuse the PATCH method with some kind of
+        #       attributes telling Sequencescape to run the `create!` method, or we need to create a new
+        #       endpoint associated with a TagLayoutTemplate that will run the `create!` method.
```
KatyTaylor, line 85: I thought you were planning to redesign it so that you could call `create!` on `TagLayout` instead, and that pulled info from the `TagLayoutTemplate`?

app/sequencescape/sequencescape/api/v2/tag_layout_template.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  # Performs the coercion of this instance so that it behaves appropriately given the direction
+  # and walking algorithm information.
+  def coerce
+    extend("limber/tag_layout_template/in_#{direction.gsub(/\s+/, '_')}s".camelize.constantize)
+    extend("limber/tag_layout_template/walk_#{walking_by.gsub(/\s+/, '_')}".camelize.constantize)
+  rescue NameError => e
+    Rails.logger.warn("Unrecognised layout options: #{e.message}")
+    extend Unsupported
```
KatyTaylor, line 19: So this method and the one below are taken from the existing `app > models > limber > tag_layout_template.rb` file... is that file needed any more?

spec/factories/tag_factories.rb
```diff
@@ -12,13 +12,23 @@ def generate_oligo
...
+    tag_group { create :v2_tag_group, v2_tags: [instance] }
   end
 
-  factory :v2_tag_group, class: Sequencescape::Api::V2::Tag do
+  factory :v2_tag_group, class: Sequencescape::Api::V2::TagGroup do
     skip_create
 
+    transient { v2_tags { [] } }
     sequence(:name) { |index| "TagGroup#{index}" }
+    tags { v2_tags.map { |t| { index: t.map_id, oligo: t.oligo } } }
```
KatyTaylor, line 25: So `v2_tags` get passed into this factory as instances of `Sequencescape::Api::V2::Tag`, but this data structure seems to change them into a simple hash? Is that how it appears in real API results too?

KatyTaylor (approved):
Looks good I think, just a couple of questions.

README.md
```diff
@@ -200,12 +200,68 @@ There are a few tools available to assist with writing specs:
...
+#### Request stubbing for the Sequencescape v1 API
 
 Request stubs are provided by webmock. Two helper methods will assist with the majority of mocking requests to the api, `stub_api_get` and `stub_api_post`. See `spec/support/api_url_helper.rb` for details.
 
 **Note**: Due to the way the api functions, the factories don't yet support nested associations.
 
+#### Request stubbing for the Sequencescape v2 API
+
+The V2 API uses `JsonApiClient::Resource` sub-classes to represent the records in memory.
+Generally these are quite dynamic so you don't need to explicitly specify every property the API will respond with.
```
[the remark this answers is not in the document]
sdjmchattie: Magic.  Basically the resources mostly look like this:

```
# frozen_string_literal: true

class Sequencescape::Api::V2::Project < Sequencescape::Api::V2::Base
end
```

but even so, that's just a template for the client gem to know that it should load a resource from that endpoint and the attributes from the API will be available to you when you use this class.  They are dynamically assigned.

The exception is for relationships.  If the API supports a resource having associated other resources, you do need to specify `has_one`, `has_many` etc in order for it to make those accessible via the client gem.

You can also add custom logic to these, like I did for TagLayoutTemplate, so you can change the named algorithms into implementations of that algorithm on the Limber side.

app/models/labware_creators/tagged_plate.rb
sdjmchattie: Yes, but we don't have `TagLayout` in V2 yet.  I plan to do that in a soon pull request.  This comment was written before we came across that solution.  I'm not sure it's worth updating since it's coming out in the next week.

app/sequencescape/sequencescape/api/v2/tag_layout_template.rb
sdjmchattie: I'm leaving removal of V1 until the end.  It's still unclear how tightly tied they are into the system and I do still need it until we remove the TODO items.

spec/factories/tag_factories.rb
sdjmchattie: Yes.  I wouldn't have done it this way as it's not how V1 works, but the V2 endpoint was pre-existing with this and if I change it, who knows what else might break.

One of the biggest (and possibly least obvious) issues with APIs is that they're even harder to change than other aspects of systems.  You're setting up a contract with people outside your team's scope for how they can interface with your application.  Making decisions about API implementations should never be taken lightly.  Deploying a bad decision has to be maintained until the next version of the API.

PR #1847, opened 2024-08-12
Y24-190-2: Use ss v2 for labware metadata

### LINKED ISSUE #1803: Y24-190-2: Use SS API v2 endpoints for Chromium pipelines

sdjmchattie:
Closes #1803 

- Change LabwareMetadata class to use V2 endpoints only.
- Fix a spectacular number of tests that this breaks.


spec/models/labware_creators/quadrant_split_plate_spec.rb
```diff
@@ -207,7 +180,24 @@
         )
       end
 
+      before do
+        allow(SearchHelper).to receive(:merger_plate_names).and_return(stock_purpose_name)
+
+        stub_v2_user(user)
+
+        stub_v2_labware(stock_plate_v2)
```
BenTopping, line 188: Is there a nice way of handling this where we can give it a list of plates instead of invoking it for each plate.

app/models/concerns/labware_creators/support_v2_source_plate.rb
```diff
@@ -0,0 +1,21 @@
...
+  # parent is using SS v1 API
+  # so this method is used to access the plate via SS v2 API
+  def source_plate
+    if parent.is_a? Limber::Plate
+      @source_plate ||= Sequencescape::Api::V2::Plate.find_by(uuid: parent.uuid) if @source_plate&.uuid != parent.uuid
+    else
+      @source_plate = nil
+    end
+
+    @source_plate
```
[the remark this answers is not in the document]
andrewsparkes: overriding the getter for the instance variable to fetch the plate using v2?

app/models/labware_creators/pooled_tubes_base.rb
```diff
@@ -93,10 +96,10 @@ def metadata_stock_barcode
     end
 
     def parent_metadata
-      if parent.is_a? Limber::Plate
```
andrewsparkes, line 96: There is the concept of a plate meant to be used in Limber and one only meant to be used in Sequencescape. You get a warning in the GUI if you scan a non-Limber plate in Limber. 
Is this Limber::Plate check to do with that? And why do you replace it here but then use it in the support_v2_plate_source_plate code above?

andrewsparkes (approved):

app/models/concerns/labware_creators/support_v2_source_plate.rb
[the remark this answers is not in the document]
sdjmchattie: It's a pattern called memoization where you use the `@` variable to store a cached value to avoid having to perform the lookup repeatedly for the parent UUID.  Rubocop requires that you use the same name for the `@` variable as you do for the method that populates it.  See a blog article on memoization here: https://blog.appsignal.com/2022/12/20/a-guide-to-memoization-in-ruby.html

When you call `self.source_plate` from elsewhere in the class (and publically in this case), the method will be preferred over the `@` variable.

app/models/labware_creators/pooled_tubes_base.rb
sdjmchattie: Maybe the method `source_plate` is badly named.  I took that from code James Glover wrote for cardinal pools plate labware creator to convert a `parent` into a V2 plate.  Not all the labware creators deal with plates I believe since one of the tests fails if I don't check that we have a plate here.  In which case it falls back to passing the parent's barcode to the LabwareMetadata.  I guess when it's a tube, as it is in one of the tests, it gives the barcode of the tube to `LabwareMetadata` and the labware gets looked up that way.  I don't know if there are any real life instances where the labware would not be able to be looked up by its barcode.  I assumed so because of the dual functionality constructor that can take the labware itself or the barcode.  But perhaps that's legacy functionality and we could make our lives a lot easier by only using the barcodes all the time?

spec/models/labware_creators/quadrant_split_plate_spec.rb
sdjmchattie: We could add more stub methods, but there are already a lot of them.  We could use some sort of nasty reflection method where we tell it to stub resources and pass the type as a string, constantizing it to get a reference to the individual stub method and looping inside the stubs module.  Or we could set this up something like:

```
[child_plate_a, child_plate_b, child_plate_c, child_plate_d, child_plate_e].each { |labware| stub_v2_labware(labware) }
```

But as these are tests, do not need to conform to DRY approaches, and I want the setup to be very clear to anyone who needs to understand what the setup and outcomes are, I don't think we should do any of that.

app/models/concerns/labware_creators/support_v2_source_plate.rb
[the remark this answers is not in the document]
BenTopping: Ah ok cool, thanks I didn't realise you could do it this way. I think I've used a similar way before something like: 
```
def source_plate
   @ source_plate ||= <method_call>
end
```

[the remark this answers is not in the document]
sdjmchattie: I think that's what I'm doing here as well.  It's just complicated by the fact that we have to check the type first before converting it and an `if` check to ensure we do over-write our memoized version if, somehow, the parent has changed.

app/models/labware_creators/pooled_tubes_base.rb
sdjmchattie: I think I've change this now to be cleaner.

PR #1866, opened 2024-08-21
Add extra QC barcode labels on  LRC PBMC Pools plates

sabrine33:


yoldas (approved):
This changes the label template for LRC PBMC Pools plate purpose. The change makes  the label template class `Labels::PlateLabelQuadQc` which has a method called `qc_label_definitions`. The presenter for the plate creates an instance of the class. The label printing form checks this `label` instance and adds hidden fields for each QC label as well.

yoldas:
Please add a brief description or highlights for the PR.

PR #1885, opened 2024-09-03
Updating tube rack form submission for PBMC plates to make contingency upload optional

### LINKED ISSUE #1485: DPL-1014 Change transfer to tube rack to support sequencing-only tubes
**User story**
As a user of the scRNA Core Cell Extraction pipeline, I would like the transfer from the LRC PBMC Bank plate to the LRC Bank Seq and LRC Bank Spare tubes to support the scenario where there is one destination sequencing rack, and NOT support the scenario where there is one destination contingency rack.

Abby, Lesley, Katy, Andrew

Abby

**Acceptance criteria**
In the transfer from LRC PBMC Bank plate to the LRC Bank Seq and Spare tubes:
- [ ] Tube rack scan upload page accepts just one scan file, in the 'sequencing' box
- [ ] Page rejects just one scan file, in the 'contingency' box
- [ ] Page accepts 2 scan files, one in each box

**Additional context**
Original text from Abby:

> Currently we restrict the user to only be able to upload either all contingency or a mix of both tube types. Since there have been some changes to the line (namely cell counting removal) and due to the mechanism for assigning tube purposes, we do not think there is a scenario where the user would be in a position to only upload contingency tubes. As we discussed though on Monday, there could be a niche case where the user may only have sequencing tubes. In light of this, would it be possible to please alter the tube rack scan upload to accommodate sequencing only and prevent contingency only?

dasunpubudumal:
Making contingency plate optional

Closes #1485 

- Making contingency plate optional and sequencing plate mandatory


app/models/labware_creators/plate_split_to_tube_racks.rb
```diff
@@ -72,6 +72,14 @@ class PlateSplitToTubeRacks < Base
     PARENT_PLATE_INCLUDES =
       'wells.aliquots,wells.aliquots.sample,wells.downstream_tubes,wells.downstream_tubes.custom_metadatum_collection'
 
+    def validate_file_presence
+      if contingency_file.blank? && sequencing_file.blank?
```
andrewsparkes, line 76: Can this whole if else statement not be simplified to a check for sequencing_file.blank?
Error 'Sequencing file must be present'.

andrewsparkes (approved):
I'm pretty sure it was this way round to start with, then they changed their minds and I had to change it round, and now they're changing it back again.

Anyway looks Ok!

.ruby-gemset
```diff
@@ -1 +1 @@
-limber
+-global
```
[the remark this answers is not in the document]
dasunpubudumal: I'm using RubyMine IDE and it did something weird automatically to this file (it has its own Ruby runtime and does weird stuff like this. It's good for navigating the codebase and managing version control, but not much else). I think it was written for Java development (IntelliJ) and Jetbrains got greedy and wrote IDEs having the same backend without providing much support for ecosystems like Ruby (and Rails) 😄 

I've reverted this file as it was.

app/models/labware_creators/plate_split_to_tube_racks.rb
dasunpubudumal: Yeah make sense. Thanks.

PR #1897, opened 2024-09-05
Y24-190-3: Use bait_library_layout API v2 endpoint

sdjmchattie:
Closes #1818 

- Move over to SS API v2 for creating and previewing BaitLibraryLayouts.
- Fix tests broken by this change.


app/models/labware_creators/baited_plate.rb
```diff
@@ -25,12 +25,13 @@ def plate
     end
 
     def bait_library_layout_preview
-      @bait_library_layout_preview ||= api.bait_library_layout.preview!(plate: parent_uuid, user: user_uuid).layout
+      @bait_library_layout_preview ||=
+        Sequencescape::Api::V2::BaitLibraryLayout.preview(plate_uuid: parent_uuid, user_uuid: user_uuid).first.layout
```
BenTopping, line 29: Should preview be returning a list (having to do .first) or is that a quirk of JSON API resources?

BenTopping (approved):

app/models/labware_creators/baited_plate.rb
sdjmchattie: I think it's a quirk of the client library here.  I tried to stop it from happening, but it seems quite intent.  I decided it wasn't worth spending any more time on and just accept you're going to get a list.

PR #1901, opened 2024-09-06
Y24-190-3 Use API v2 TagLayout

sdjmchattie:
Closes #1818 

- Use new endpoint for TagLayout.
- Use the custom method for TagLayoutTemplate to create a TagLayout.

### DIFF
--- a/.pryrc
+++ b/.pryrc
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+if defined?(PryByebug)
+  Pry.commands.alias_command 's', 'step'
+  Pry.commands.alias_command 'n', 'next'
+  Pry.commands.alias_command 'f', 'finish'
+  Pry.commands.alias_command 'c', 'continue'
+end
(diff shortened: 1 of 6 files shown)


spec/models/labware_creators/custom_tagged_plate_spec.rb
```diff
@@ -232,17 +233,18 @@ def expect_state_change_creation
           let(:tag_plate_state) { 'exhausted' }
 
           it 'creates a tagged plate' do
+            # This one will be VERY different
```
BenTopping, line 236: Intentional comment?

.pryrc
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+if defined?(PryByebug)
```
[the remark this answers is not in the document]
sdjmchattie: Hah.  There is a lot of debugging to do at times and I found this as a way of shortening the commands you use the most to single letters.

spec/models/labware_creators/custom_tagged_plate_spec.rb
sdjmchattie: Well it came from removed line 241 below and I'm making the same expectation, just via API V2, so hopefully it's intentional.

BenTopping: Yeah just not sure what the difference is that the comment is suggesting.

sdjmchattie: I don't know.  I didn't want to try to clean up existing comments and I assumed it was made for a reason.  It's very easy to put in comments that aren't useful in the future 😄

PR #1916, opened 2024-09-13
Y24-190-3: Use SpecificTubeCreation on API v2

sdjmchattie:
Closes #1818 

- Use API v2 endpoints for SpecificTubeCreation creation.
- Update tests to confirm behaviour is correct.


app/sequencescape/sequencescape/api/v2/specific_tube_creation.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Represents a specific tube creation in Limber via the Sequencescape API
+class Sequencescape::Api::V2::SpecificTubeCreation < Sequencescape::Api::V2::Base
+  has_many :children, class_name: 'Sequencescape::Api::V2::Tube'
+  has_many :parents, class_name: 'Sequencescape::Api::V2::Asset'
```
andrewsparkes, line 6: Labware rather than Asset?

app/sequencescape/sequencescape/api/v2/specific_tube_creation.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Represents a specific tube creation in Limber via the Sequencescape API
+class Sequencescape::Api::V2::SpecificTubeCreation < Sequencescape::Api::V2::Base
```
andrewsparkes, line 4: I don't think I really follow with JSON API why these relationships are defined at all for this endpoint, as the payload is sending some uuids rather than data objects for tubes, parents etc. I guess because it returns the tubes? Maybe you can help me understand how this all fits together.

andrewsparkes (commented):
Looks good but would like some discussion, thx

andrewsparkes (approved):

app/sequencescape/sequencescape/api/v2/specific_tube_creation.rb
sdjmchattie: I will need to change the resource on the Sequencescape side, here: https://github.com/sanger/sequencescape/blob/develop-Y24-190/app/resources/api/v2/specific_tube_creation_resource.rb#L88

To be honest I was going to say that the model in Sequencescape has these as Assets and that's why I had to use Asset for the model type.  However, I'm not seeing that now.  I was trying to keep things as consistent with v1 as possible during this phase because it's better to get all of v2 in than it is to try to resolve other issues in the backlog like Assets vs Labware.

I will look at this though and see if it's a simple fix.  If not, I'm going to leave it as we need to get this long list of endpoints done.

sdjmchattie: OK it's looking like I can just switch this over.  Not sure why it didn't add up for me last time I looked as I did consider this.  I haven't tested it with the integration suite, but I plan to do that all somewhat soon.

app/sequencescape/sequencescape/api/v2/specific_tube_creation.rb
```diff
@@ -0,0 +1,8 @@
...
+  has_many :children, class_name: 'Sequencescape::Api::V2::Tube'
```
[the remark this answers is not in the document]
sdjmchattie: This process creates tubes only, not plates.

PR #1951, opened 2024-09-23
Y24-313 scRNA pooling calculation changes

### LINKED ISSUE #1909: Y24-313 - PBMC donor pooling - use requested number of samples per pool
**User story**
As a user of the scRNA Core pipeline, I would like the automatic allocation of samples to pools to behave differently. 

Currently, a config table is used that maps number of samples to number of pools (see `interim_scrna_core_donor_pooling.yml` and [comment here](https://github.com/sanger/limber/issues/1390#issuecomment-1961202428)). Instead, we would like to specify a number of samples per pool, which could vary from pool to pool - this information will come in at the point of submission (implemented in https://github.com/sanger/sequencescape/issues/4337).

This is a first step in supporting a more flexible pooling strategy, where the customer can specify the number of cells per pool and the number of cells per chip well. This story is to support a pre-MVP so that something can be in place for when R&D do a test run.

Abby, Katy

Abby

**Acceptance criteria**
- [x] Instead of using `interim_scrna_core_donor_pooling.yml` to decide the number and size of the pools, labware creator `donor_pooling_plate.rb` uses the requested number of samples per pool.
- [x] 'Requested samples per pool' is retrieved from the request metadata on the cDNA Prep requests. For now, we can assume it will be the same value for all requests (for one run of the labware creator)
- [x] Can assume that the number of samples will be exactly divisible by the 'number of samples per pool'.
- [x] Integration Suite test will have to insert the 'Number of samples per pool' value against the requests to carry on working.

**Dependencies**
The number of samples per pool will come in as part of the submission:
- https://github.com/sanger/sequencescape/issues/4337

It's not blocking as, for this story, we can just assume it's stored against the request metadata.

**Additional context**
This is for the 'pre-MVP' version, to be used for R&D's end to end test. This will have to be re-jigged for MVP to allow a variable number of samples per pool, and potentially to add logic to achieve an even distribution of concentrations if possible.

Integration suite test PR is [here](https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/170)

andrewsparkes:
Closes #1909


app/models/labware_creators/donor_pooling_plate.rb
```diff
@@ -100,6 +93,31 @@ def source_wells_for_pooling
...
+    # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+    def fetch_number_of_samples_per_pool_from_request
+      source_wells = source_wells_for_pooling
+      return if source_wells.blank?
+
+      number_of_samples_per_pool =
+        source_wells.first&.aliquots&.first&.request&.request_metadata&.number_of_samples_per_pool || nil
+
+      if number_of_samples_per_pool.nil?
+        raise StandardError, 'Error: request_metadata.number_of_samples_per_pool is nil'
```
yoldas, line 113: Note that if the option is not provided in submission or added to the request metadata using the rake task, this will cause a 500 error while creating the PBMC Pools plate. So, the number_of_samples_per_pool option is not optional.

app/models/labware_creators/donor_pooling_plate.rb
```diff
@@ -218,7 +238,7 @@ def tag_depth_hash
     def build_pools
       groups = split_single_group_by_study_and_project(source_wells_for_pooling)
       groups = split_groups_by_unique_donor_ids(groups)
-      distribute_groups_across_pools(groups, number_of_pools)
+      distribute_groups_across_pools(groups, calculated_number_of_pools)
```
yoldas, line 241: Note that `distribute_groups_across_pools` uses a halving algorithm in distribution at the moment, which causes uneven distribution in small number of pools. I think It needs a weighted sharing of pools to each group based on size first and then either a round robin or a fair sharing of remainder after integer division for distribution in them.

app/sequencescape/sequencescape/api/v2/request_metadata.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Represents a request metadatum in Limber via the Sequencescape API
+class Sequencescape::Api::V2::RequestMetadata < Sequencescape::Api::V2::Base
```
yoldas, line 4: metadatum in the comment is confusing I think. RequestMetadata is set to be singular for this resource (inflections, uncountable). I would avoid using both forms in the same context.

yoldas (approved):
Looks good. I have added a few notes.

app/models/labware_creators/donor_pooling_plate.rb
andrewsparkes: Which is correct, but maybe the message should be customer friendly like saying it should have been set on submission? Not sure how they'd fix that other than support issue.

app/sequencescape/sequencescape/api/v2/request_metadata.rb
andrewsparkes: I think I changed the class name and forgot the comment. Will fix.

Issue #1914, opened 2024-09-13
Y24-319 - Pooling driver file tweaks


KatyTaylor:
**User story**
As a user of the scRNA Core pipeline, I would like some tweaks to the driver file that controls the creation of the LRC PBMC Pools plate, so that:
- More appropriate values for this pipeline are used rather than the defaults that were copied from Cardinal
- Resuspension volume is added, since we are allowing different target numbers of cells on different chip wells.

**Who are the primary contacts for this story**
Abby, Katy

**Who is the nominated tester for UAT**
Abby

**Acceptance criteria**
Driver file `hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools` is changed as follows:
- [ ] Default required number of cells is 30,000 (currently 5,000)
- [ ] Maximum sample volume is 60ul (currently 50)
- [ ] Cell count used in calculation is 'total cells' not 'live cells'
- [ ] Resuspension volume column (for each pool / target well) is calculated as:
    - (Number of samples in pool x 30,000 x 0.95238) / 2,400
    - Where:
        - 30,000 is the required number of cells
        - 0.95238 is accounting for wastage of material when transferring between labware
        - 2,400 is the desired chip loading concentration

**Additional context**
This is for the 'pre-MVP' version, to be used for R&D's end to end test. For the MVP version, we will probably store some of the constants in a more central place, to be used in other calculations elsewhere.

The 'required number of cells' is currently configurable by study - if specified on the study, this overrides the default value. This is no longer required for the project (it will be a blanket value of 30,000) - however, there doesn't seem any harm in leaving it in in case it's wanted in future.

PR #1973, opened 2024-10-02
Y24-319-1 - Additional pooling driver file adjustments.

BenTopping:
Additional tweaks on top of https://github.com/sanger/limber/issues/1914
Requirements from slack:
- "we want to cap the minimum vol to take at 5ul"
- "we don't need this (vol) or the resuspension vol to 2 decimal places, 1 is fine."

- Updates `hamilton_lrc_pbmc_defrost_pbs_to_lrc_pbmc_pools` to reflect the above requirements.


andrewsparkes (approved):
Looks good.
When we revisit the pooling calculations later we can revisit how to put the numbers used in the scRNA calculations somewhere central.

PR #2028, opened 2024-10-25
Y24-099 plate split to racks

### LINKED ISSUE #1696: Y24-099 Tube Rack Banking Sequencescape API
**User story**
As PSD, we would like to update Sequencescape v2 API to provide Limber support for TubeRack model similar to Plates using the versioned tube rack model proposed in Reuse of TubeRacks document.

Abdullah, Katy, Andrew

TBD

**Acceptance criteria**
- [x] Check existing labware creator (in  DPL-811) for Sequencescape API calls and replace them for tube racks.
- [x] Check TubeRacks and associated objects are retrieved by using JSONAPI includes, i.e. Tubes and RackedTube.
- [x] Add resource methods to TubeRacks and RackedTubes so that they are similar to Plate and Wells.
- [x] Add TubeRack barcode into new TubeRack poly_metadata in order to enable history tracking of physical tube rack

**Dependencies**
- #1482

References
- #1314
- #1713

**Additional context**
The versioned tube rack model is researched in DPL-1011 Re-use of tube racks and the model is summarised in the Google Document[ DPL-1011 Re-use of tube racks](https://docs.google.com/document/d/1rvPXeCk5rBEf72VkchHCD2FRmWANboke2pln_Lt_j5U/edit?usp=sharing)

This story is part of re-implementation of [DPL-811 Transfer PBMC isolations into FluidX tube racks](https://github.com/sanger/limber/issues/1314) (Banking) using the versioned tube rack model.

Story relationships are shown in the Lucidchart diagram [TubeRack Stories](https://lucid.app/lucidchart/d88bdcfd-9628-4fec-8b5a-453f680ee06b/edit?viewport_loc=-1051%2C-316%2C2770%2C1486%2C0_0&invitationId=inv_13d74286-1cc2-4c2f-aaf0-30887764601c) .

It is not possible to release the implementation of this story in isolation.

andrewsparkes:
Closes #1696 

Adds specific tube rack and tube creation for scRNA


app/frontend/stylesheets/limber/screen.scss
```diff
@@ -280,6 +282,7 @@ nav.robots-list {
   .work-completion-button,
   .create-plate-button,
   .create-tube-button,
+  .create-tube_rack-button,
```
KatyTaylor, line 285: Seems like some of this stuff could just be generic, like `.create-labware-button` or `params[:limber_labware_id]` in the above file. Let's not add refactoring to the scope of the story though!

app/helpers/page_helper.rb
```diff
@@ -53,6 +53,7 @@ def jumbotron(jumbotron_id = nil, options = {}, &)
   # eg. state_badge('pending')
   # <span class="state-badge-pending">Pending</span>
   def state_badge(state)
+    return if state.blank?
```
KatyTaylor, line 56: What happens if this returns `nil`? The state badge is not shown?

app/models/concerns/presenters/creation_behaviour.rb
```diff
@@ -15,6 +15,10 @@ def compatible_tube_purposes
     construct_buttons(purposes_of_type('tube'))
   end
 
+  def compatible_tube_rack_purposes
+    construct_buttons(purposes_of_type('tube_rack'))
+  end
+
```
KatyTaylor, line 21: Feels like this wants to be refactored to:
```
def compatible_labware_purposes(labware_type)
    construct_buttons(purposes_of_type(labware_type))
end
```

But again, maybe we can save up refactoring ideas for another story to try to draw a line under this story! Will keep commenting them if I notice and we can decide what to do.

app/models/state_changers.rb
```diff
@@ -58,6 +58,23 @@ def self.lookup_for(purpose_uuid)
...
 
+  # The tube rack state changer is used by TubeRacks.
+  # It contains tubes.
+  class TubeRackStateChanger < DefaultStateChanger
+    # TubeRacks contain tubes
+    def contents_for(_target_state)
+      return nil unless FILTER_FAILS_ON.include?(target_state)
+
+      # determine list of tubes requiring the state change
+      tubes_locations_filtered = labware.tubes.reject { |t| t.state == 'failed' }.map(&:racked_tube.coordinate)
```
KatyTaylor, line 69: Not sure about the `&:racked_tube`.

A tube can have many `racked_tubes`, can't it, because it can be in many tube racks over time. You could do:

```
labware.racked_tubes.reject { |rt| rt.tube.state == 'failed' }.map(&:coordinate)
```
Would that work?

app/sequencescape/sequencescape/api/v2.rb
```diff
@@ -22,6 +22,13 @@ module Sequencescape::Api::V2
     }
   ].freeze
 
+  # NB. a receptale can have many aliquots, and aliquot.request is an array (for some reason)
```
KatyTaylor, line 25:
```suggestion
  # NB. a receptacle can have many aliquots, and aliquot.request is an array (for some reason)
```

app/sequencescape/sequencescape/api/v2/aliquot.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class Sequencescape::Api::V2::Aliquot < Sequencescape::Api::V2::Base # rubocop:todo Style/Documentation
-  # requires shallow path otherwise get a resource not found issue
-  belongs_to :request, shallow_path: true
+  # requires shallow path otherwise get a resource not found issue TODO: where/when do we get this?
+  has_one :request, shallow_path: true
```
KatyTaylor, line 5: What was the driver for this change? Feels a bit dangerous!

app/sequencescape/sequencescape/api/v2/tube_rack.rb
```diff
@@ -1,25 +1,42 @@
...
+  # to allow us to treat old and new api the same
+  #
+  # @return [ActiveModel::Name] The resource behaves like a Limber::TubeRack
+  #
+  def model_name
+    ::ActiveModel::Name.new(Limber::TubeRack, false)
+  end
+
+  has_many :racked_tubes, class_name: 'Sequencescape::Api::V2::RackedTube'
+  has_many :parents, class_name: 'Sequencescape::Api::V2::Asset' # Having issues with polymorphism, temporary class
```
KatyTaylor, line 28: Feels like I only understand that comment because I've read the original - I'd leave it out.

app/sequencescape/sequencescape/api/v2/tube_rack.rb
```diff
@@ -1,25 +1,42 @@
...
 
   property :name
   property :size
   property :number_or_rows
   property :number_of_columns
+  property :tuberack_barcode
```
KatyTaylor, line 34: I would just go for `barcode`... but I guess you're following a convention from tube and plate classes?

config/purposes/scrna_core_cell_extraction.yml
```diff
@@ -86,34 +86,52 @@ LRC PBMC Bank:
...
       child_seq_tube_purpose_name: LRC Bank Seq
       child_seq_tube_name_prefix: SEQ
+      child_seq_tube_rack_purpose_name: TR LRC Bank Seq
       child_spare_tube_purpose_name: LRC Bank Spare
       child_spare_tube_name_prefix: SPR
-  :ancestor_stock_tube_purpose_name: LRC Blood Vac
-  :presenter_class: Presenters::SimpleTubePresenter
-  :state_changer_class: StateChangers::AutomaticTubeStateChanger
+      child_spare_tube_rack_purpose_name: TR LRC Bank Spare
+      child_tube_rack_metadata_key: 'tube_rack_barcode'
```
KatyTaylor, line 104: We want this key to be the same EVERYWHERE, right? It's part of Abdullah's design doc?
So it doesn't feel quite right here against a specific purpose.

config/routes.rb
```diff
@@ -33,6 +35,7 @@
   resources :limber_plates, controller: :plates do
     resources :children, controller: :plate_creation
     resources :tubes, controller: :tube_creation
+    resources :tube_racks, controller: :tube_rack_creation
```
KatyTaylor, line 38: Don't quite understand what these routes are for

KatyTaylor (commented):
Haven't looked at the big one yet - the labware creator - but have left some comments on the smaller changes.

app/frontend/stylesheets/limber/screen.scss
andrewsparkes: Agree, assuming the buttons don't have to have different text or some other difference.

app/models/concerns/presenters/creation_behaviour.rb
andrewsparkes: Yeah. I think this is building the 'other actions' grey buttons. That could do with some better ordering/structuring in any case.

app/sequencescape/sequencescape/api/v2/aliquot.rb
andrewsparkes: Why does that feel dangerous to you? Stuart and I just found out it does affect the requests you get through the aliquot. With has_one it just gives us id, with belongs_to we get a full object. Why?

config/purposes/scrna_core_cell_extraction.yml
andrewsparkes: How do I reference something like a TubeRack class constant from within a yml file?

config/routes.rb
andrewsparkes: Nope nor me. I think it's to do with asset creation, parent plate here with possible child creation routes?

app/sequencescape/sequencescape/api/v2/tube_rack.rb
andrewsparkes: removed, thanks

app/sequencescape/sequencescape/api/v2/tube_rack.rb
andrewsparkes: Doesn't appear to be needed so removed it. Class uses common HasBarcode concern.

app/frontend/stylesheets/limber/screen.scss
andrewsparkes: Plus the suggested action Add button now looks like 'Add <purpose> tube_rack' with an underscore which looks odd (works for tube and plate as they are single word classes).

app/sequencescape/sequencescape/api/v2.rb
andrewsparkes: done

config/purposes/scrna_core_cell_extraction.yml
andrewsparkes: if it will ALWAYS be the same key then there is no need for Limber to know what it is or include it in the purpose config. instead of passing it through for each tube rack in the attributes, I can pull it out of a tuberack constants initializer file like I made in the other story on the SS side.

app/models/state_changers.rb
KatyTaylor: Agreed

KatyTaylor: Discussed whether we should be filtering out 'cancelled' tubes here too? Have this - https://github.com/sanger/limber/pull/2034/files

Also need to test failing and cancelling at the tube rack level. 

Think about whether to create a new story for any of these things.

PR #2034, opened 2024-10-29
Y24-190-3: Use API V2 for Tube queries everywhere

sdjmchattie:

- Switch all API v1 usage for Tubes to use the v2 endpoint.
- Update tests to match expectations for API calls.


app/sequencescape/sequencescape/api/v2/tube.rb
```diff
@@ -51,6 +51,11 @@ def self.find_all(params)
 
   delegate :requests_as_source, to: :receptacle
 
+  # Mocked out for the time being
```
sdjmchattie, line 54: This comment was carried here from the V1 Tube model.  I don't know how long "the time being" is, but I was trying to avoid making changes to how this works compared to V1.

spec/models/labware_creators/multi_stamp_tubes_using_tube_rack_scan_spec.rb
```diff
@@ -193,8 +193,7 @@
 
     it 'creates a plate!' do
       # barcode from multi_stamp_tubes_using_tube_rack_scan/tube_rack_scan_valid.csv
-      subject.labware.barcode.machine = 'AB10000001'
-      subject.labware.barcode.ean13 = nil
+      subject.labware.labware_barcode = { 'human_barcode' => 'AB10000001', 'machine_barcode' => 'AB10000001' }
```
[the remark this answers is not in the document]
sdjmchattie: I'm just keeping it the way [this person](https://github.com/sanger/limber/commit/7add6bef3bb022cda94997cdb988a5499441d6a6#diff-0d6231df332e7091e5ac8f72a6b5034acc322781a3659077b6c99df86a2ce22fR235) set it up originally ;)

app/sequencescape/sequencescape/api/v2/tube.rb
```diff
@@ -51,6 +51,11 @@ def self.find_all(params)
...
+  def in_progress_submission_uuids(*)
```
harrietc52, line 55: If the method for `v1` previously didn’t accept any params, would the `(*)` change the definition of the method? (just incase it is being used somewhere)

spec/factories/tube_factories.rb
```diff
@@ -116,10 +118,25 @@
       parents { [] }
       purpose { create :v2_purpose, name: purpose_name, uuid: purpose_uuid }
 
+      siblings_count { 0 }
+      sibling_default_state { 'passed' }
+      siblings do
```
harrietc52, line 123: Ah so is one `sibling_tubes` created by default, then if there is to be any more, `siblings_count` will be specified to be > 0>

spec/factories/tube_factories.rb
```diff
@@ -116,10 +118,25 @@
...
+            uuid: "sibling-tube-#{i}"
+          }
+        end
+      end
+
       # The CustomMetadatumCollection will be cached as a relationship in the after(:build) block.
       custom_metadatum_collection { nil }
     end
 
+    to_create { |instance, _evaluator| instance.mark_as_persisted! }
```
harrietc52, line 138: Would you mind adding a comment about this `mark_as_persisted!` magic!

spec/features/pool_tubes_at_end_of_pipeline_spec.rb
```diff
@@ -12,11 +12,15 @@
...
-    create(:v2_tube, uuid: tube_uuid, state: 'passed', barcode_number: 1, purpose_name: 'Example Purpose')
-  end
-  let(:example_tube) do
-    json(:tube_with_siblings, uuid: tube_uuid, siblings_count: 1, state: 'passed', barcode_number: 1)
+  let(:tube) do
+    create(
+      :v2_tube,
+      barcode_number: 1,
+      purpose_name: 'Example Purpose',
+      siblings_count: 1,
```
harrietc52, line 20: Maybe this is slightly confusing, (only if my above comment is correct), that is you put `siblings_count: 1`, it is actually creating 2, one by default and one extra?

spec/features/charge_and_pass_libraries_spec.rb
```diff
@@ -27,9 +27,9 @@
   context 'plate with no submissions to be made' do
     before { create :purpose_config, uuid: 'example-purpose-uuid' }
 
-    let(:labware_barcode) { example_plate_v2.labware_barcode.machine }
+    let(:plate_barcode) { plate.labware_barcode.machine }
```
harrietc52, line 30: I thought it was `.machine_barcode`, but maybe its `.machine` 😃

harrietc52 (approved):
Looks good, some minor thoughts

spec/factories/tube_factories.rb
sdjmchattie: Not the way I would have designed it, but it's mimicking the way it's done for the v1 factory.  If I don't mimic that, I need to change the tests to match and at that point, who knows if the behaviour is unchanged?  For what it's worth, no the first sibling isn't an actual sibling.  It's a reference to the tube itself.  So it counts itself as a sibling of itself.  The sibling count is actually the number of **_other_** tubes it is related to.

spec/features/pool_tubes_at_end_of_pipeline_spec.rb
sdjmchattie: I think it's important to maintain the same functionality as the v1 factories, even if we don't agree with it.  I don't necessarily understand why it was written that way, but we're trying to prove the new code changes do not have any different behaviours under v2, so we have to swallow weird mocking choices.

spec/features/charge_and_pass_libraries_spec.rb
sdjmchattie: I think it is `machine_barcode` on Sequencescape but then it has become `machine` by the time it's gone through various layers into Limber.

```
# Wraps labware barcodes to assist with rendering and to provide a cleaner interface
class LabwareBarcode
  attr_reader :human, :machine, :ean13

  def initialize(human:, machine:, ean13: nil)
    @human = human
    @machine = machine
    @ean13 = ean13
  end
```

PR #2062, opened 2024-11-08
Y24 335 updating flowcel types and library types

dasunpubudumal:
Closes #

- Adding `limber_hic` request type.
- Fixing the issue where it only prints the label of the last tube of pool stock tubes.

The corresponding sequencescape changes are https://github.com/sanger/sequencescape/pull/4497.


config/pipelines/bespoke_pcr.yml
```diff
@@ -2,7 +2,9 @@
 Bespoke PCR:
   pipeline_group: Bespoke PCR
   filters:
-    request_type_key: limber_pcr_bespoke
+    request_type_key:
+      - limber_pcr_bespoke
+      - limber_hic # Have to move this into a the new pipeline config when HiC pipeline is implemented.
```
KatyTaylor, line 7: Could put the HiC pipeline story reference here in the comment.

KatyTaylor (approved):

config/pipelines/hic.wip.yml
dasunpubudumal, line 1: Had to take this out of `Bespoke PCR` to make `limber_hic` WIP-able.

spec/controllers/print_jobs_controller_spec.rb
```diff
@@ -16,10 +16,7 @@
           label_templates_by_service: 'tube_label_template_1d',
           labels: [{ 'label' => { 'barcode' => '12345', 'test_attr' => 'test' } }],
           labels_sprint: {
-            sprint: {
-              'extra_right_text' => 'some x right text',
-              'extra_left_text' => 'some x left text'
-            }
+            sprint: [{ 'extra_right_text' => 'some x right text', 'extra_left_text' => 'some x left text' }]
           },
```
yoldas, line 20: Why is this changed to an array? 
As far as I know,  labels_sprint object provides all attributes in a a single object. The labels object on the other hand has separate objects for main label and lid label for a tube in an array for example.

dasunpubudumal: Screenshot 2024-11-20 at 10 31 01 AM

When given on a single object, if you have multiple tube labels to be printed, the previous tubes will be overwritten by the last tube; therefore printing only the last tube label.

yoldas: Thanks. So it fixes the issue for Squix printing. That will be still a problem if Toshiba printing is used then. Because the array there is not for multiple labels; it is for sections of a single label.

PR #2069, opened 2024-11-15
scRNA - Pooling calculation stories Epic

andrewsparkes:
#### Includes
[Y24-375 - Constants refactor](https://github.com/sanger/limber/tree/y24-375-file-downloads-improvements-for-mvp)
Y24-384 scRNA - adjust pooling for full allowance
Y24-376 - Changes to donor pooling algorithm
Fix poly metadatum by key for chip loading driver


spec/models/presenters/donor_pooling_plate_presenter_spec.rb
```diff
@@ -69,63 +69,39 @@
...
-      end
-
-      it 'warns the user about one study' do
-        expect(subject).not_to be_valid # There are warnings in the errors collection.
-        study_names = [study_to_a1].map(&:name).join(', ')
-        formatted_string = format(warning_template, default_cell_count, study_names)
-        expect(subject.errors[:required_number_of_cells]).to include(formatted_string)
-      end
-    end
-
```
yoldas, line 118: The removed lines were the testing the error formatting for required number of cells of one or more studies as well. I see the replacements contain validity checks on the presenter but no checking of error text.

Because the error text is not checked anymore, we do not need the `warning_template` block, which returns  `Validators::RequiredNumberOfCellsValidator::STUDIES_WITHOUT_REQUIRED_NUMBER_OF_CELLS_PER_SAMPLE_PER_POOL` .

yoldas (approved):
Great refactoring: All constants and options are at a single location.

I have one minor comment about a few lines of dead test code. Although it can be used in a test to check formatted error text, it is not absolutely necessary and the validator code lines are visited in valid? calls in the replacement tests anyway.

yoldas (approved):
Looks good.

KatyTaylor (approved):

PR #2078, opened 2024-11-20
Y24-384 scRNA - adjust pooling for full allowance

### LINKED ISSUE #1992: Y24-384 - Adjust requested pooling strategy to enable two full runs
**User story**
As a user of the scRNA Core pipeline, I would like the LIMS to check whether there is enough material for two full runs on the chip, just prior to pooling (into the LRC PBMC Pools plate), in case some samples have been failed and therefore numbers changed since the point of submission - so that the parameters can be adjusted if necessary to allow for two full runs.

Katy, Abby

Abby

**Acceptance criteria**
When creating LRC PBMC Pools plate, after calculating the number and size of pools (covered in https://github.com/sanger/limber/issues/1985)...

- [x] For any pools with 5-8 samples, calculate whether there is enough material for two full runs on the chip - is 'final resuspension volume' >= 'full allowance'? (see additional context for the calculations)
- [x] If not, the 'number of cells per chip well' to be used going forward (for each pool) is not taken from the value requested in the submission, but instead selected from the table below:
image (3)
- [x] The selected 'number of cells per chip well' is saved for each pool for future reference by [driver files](https://github.com/sanger/limber/issues/1983) (and for audit / troubleshooting purposes) - this is necessary even if we are using the requested value from the submission. Think this would make sense to save against each well on the destination plate? Maybe as polymetadata?
EDIT: Andrew: As part of story [Y24-375](https://github.com/sanger/limber/issues/1983) I have added an expectation for that information to be stored in the pool well as poly-metadata. The key is stored in a config file `config/initializers/scrna_config.rb` and is set to this (key and value):  `number_of_cells_per_chip_well_key: 'scrna_core_pbmc_donor_pooling_number_of_cells_per_chip_well'` and is accessed with `Rails.application.config.scrna_config[:number_of_cells_per_chip_well_key]`. NB. Values in poly-metadata are stored as strings even when numbers.

**Dependencies**
- https://github.com/sanger/limber/issues/1985

This story uses the same calculations as https://github.com/sanger/sequencescape/issues/4460 - consider whether to share code.

**Additional Context**
_Common constants / variables:_
    * "Chip loading concentration" = 2,400

_Calculating "final resuspension volume":_
* "Total cells in 300ul" = ("Number of samples in the pool" * 30,000) * 0.95
    * "Number of samples in the pool" is the actual number of samples in this pool - can vary between pools - count them
    * 30,000 is the fixed number of cells taken for each sample into the pool
    * 0.95 is due to wastage when transferring between labware
* "Final resuspension volume" = "Total cells in 300ul" / "Chip loading concentration" 

_Calculating "full allowance":_
* "Chip loading volume" = "Number of cells per chip well" / "Chip loading concentration" 
    * "Number of cells per chip well" comes from the submission / request 

* "Full allowance" = ( "Chip loading volume" * 2) + 25
    * 2 is because this is for 2 runs
    * 25 is 2 lots of 10ul for cell counting, and 5ul for wastage when transferring between labware

andrewsparkes:
Closes #1992 

Adds the calculations to check for full allowance, adjusting number of cells per chip well if needed and storing that value on the destination pool wells for use in creating the driver file for loading the chip plate later.

### DIFF
--- a/spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
+++ b/spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
@@ -7,7 +7,7 @@ class TestPoolingClass
 end
 
 RSpec.describe LabwareCreators::DonorPoolingCalculator do
-  let(:test_pooling_class) { TestPoolingClass.new }
+  let(:instance_of_test_pooling_class) { TestPoolingClass.new }
   let(:pool) { [source_well1] }
   let(:source_well1) { create :v2_well, aliquots: [aliquot1] }
   let(:aliquot1) { create :v2_aliquot, request: request1 }
@@ -21,7 +21,7 @@ RSpec.describe LabwareCreators::DonorPoolingCalculator do
 
       it 'returns nil' do
         expect do
-          test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)
+          instance_of_test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)
         end.to raise_error StandardError,
                     'No request found for source well at A1, cannot fetch ' \
                       'cells per chip well metadata for full allowance calculations'
@@ -30,7 +30,9 @@ RSpec.describe LabwareCreators::DonorPoolingCalculator do
 
     context 'when there is a single source well with request_metadata' do
       it 'returns the number of cells per chip well' do
-        expect(test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)).to eq(cells_per_chip_well)
+        expect(instance_of_test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)).to eq(
+          cells_per_chip_well
+        )
       end
     end
 
@@ -42,7 +44,9 @@ RSpec.describe LabwareCreators::DonorPoolingCalculator do
       let(:pool) { [source_well1, source_well2] }
 
       it 'returns the number of cells per chip well from the first source well' do
-        expect(test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)).to eq(cells_per_chip_well)
+        expect(instance_of_test_pooling_class.send(:number_of_cells_per_chip_well_from_request, pool)).to eq(
+          cells_per_chip_well
+        )
       end
     end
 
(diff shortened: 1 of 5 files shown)


spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+  end
+
+  describe '#calculate_chip_loading_volume' do
+    context 'when the number of cells per chip well is present' do
+      let(:expected_volume) do
+        cells_per_chip_well / Rails.application.config.scrna_config[:desired_chip_loading_concentration]
+      end
+
+      it 'calculates the chip loading volume' do
+        expect(test_pooling_class.send(:calculate_chip_loading_volume, pool, cells_per_chip_well)).to eq(
```
yoldas, line 81: calculate_chip_loading_volume accepts one argument; remove pool; keep cells_per_chip_well.

spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+  end
+
+  describe '#calculate_total_cells_in_300ul' do
+    let(:count_of_samples_in_pool) { 10 }
+    let(:num_cells_per_sample) { Rails.application.config.scrna_config[:required_number_of_cells_per_sample_in_pool] }
+    let(:wastage_factor) { Rails.application.config.scrna_config[:wastage_factor] }
+    let(:expected_volume) { (count_of_samples_in_pool * num_cells_per_sample) * wastage_factor }
+
+    it 'calculates the value correctly' do
+      expect(test_pooling_class.send(:calculate_total_cells_in_300ul, pool, count_of_samples_in_pool)).to eq(
```
yoldas, line 68: calculate_total_cells_in_300ul accepts one argument; remove pool; keep count_of_samples_in_pool

spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
```diff
@@ -0,0 +1,187 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+class TestPoolingClass
+  include LabwareCreators::DonorPoolingCalculator
+end
+
+RSpec.describe LabwareCreators::DonorPoolingCalculator do
+  let(:test_pooling_class) { TestPoolingClass.new }
```
yoldas, line 10: test_pooling_class returns an instance of TestPoolingClass but the name is misleading; test_pooling_instance ?

yoldas (commented):
There are two test failures because of wrong number of algorithms. I marked where they are. I have a minor comment about a naming. I am also writing the following note but it is not asking for any change.

Note: float comparisons are done using equality in tests as they are computed the same way and compared. From different sources, different algorithms, or one side hard-coded, it would need to use the be_within matcher. https://rspec.info/features/3-12/rspec-expectations/built-in-matchers/be-within/

spec/models/concerns/labware_creators/donor_pooling_calculator_spec.rb
andrewsparkes: changed

yoldas (approved):

PR #2091, opened 2024-11-27
Y24-190: Use v2 tag layout creation with a template

sdjmchattie:

- Use the new endpoint properties to create tag layouts with a template.
- Update tests to match.


spec/features/creating_a_tag_plate_spec.rb
```diff
@@ -177,7 +187,7 @@
       let(:template_factory) { :v2_dual_index_tag_layout_template }
 
       context 'when nothing has been done' do
-        #let(:submission_pools) { json(:dual_submission_pool_collection) }
+        let(:enforce_uniqueness) { false }
```
sdjmchattie, line 190: I'm not sure why I needed to add this.  Is it the correct expectation @andrewsparkes or are the mocks letting us down and so this should have been submitted as `true`?

PR #2099, opened 2024-12-02
API v2 Milestone 3

sdjmchattie:

All has been reviewed before, but this is a last once over before going into `develop`.

- Update usage of v2 endpoints on Sequencescape:
  - `bulk_transfers`
  - `plates`
  - `plate_conversions`
  - `plate_creations`
  - `qc_files`
  - `submission_pools`
  - `transfer_request_collections`
  - `transfers`
  - `tubes`
  - `tube_from_tube_creations`
- Update tests accordingly.


andrewsparkes (approved):

yoldas (approved):

app/controllers/qc_files_controller.rb
```diff
@@ -11,17 +11,20 @@ class QcFilesController < ApplicationController
...
   def index
-    respond_to { |format| format.json { render json: { 'qc_files' => asset.qc_files } } }
+    respond_to { |format| format.json { render json: { 'qc_files' => asset.qc_files.map { |q| qc_file_to_json(q) } } } }
   end
 
   def show
-    response = api.qc_file.find(params[:id]).retrieve
-    filename = /filename="([^"]*)"/.match(response['Content-Disposition'])[1] || 'unnamed_file'
-    send_data(response.body, filename: filename, type: 'sequencescape/qc_file')
+    qc_file = Sequencescape::Api::V2::QcFile.find(uuid: params[:id]).first
```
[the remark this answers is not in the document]
sdjmchattie: We could (in a separate piece of work) create an alias for the `Sequencescape::Api::V2` module.  Maybe something like `SSAPIv2` so that these would become `SSAPIv2::QcFile` etc.  I don't think we should name the real module differently, but an alias would be a reasonable approach.

app/sequencescape/sequencescape/api/v2.rb
```diff
@@ -7,6 +7,7 @@ module Sequencescape::Api::V2
   # transfer_request_as_target.source_asset is added to the includes to make
   # the source receptacles available in presenters.
   PLATE_PRESENTER_INCLUDES = [
+    :ancestors,
```
sdjmchattie, line 10: @andrewsparkes and @StephenHulme I added this to include the ancestors when we get a plate for a presenter.  This didn't fix it on its own though, so see the other comment I'm about to make.

app/sequencescape/sequencescape/api/v2/plate.rb
```diff
@@ -103,14 +105,16 @@ def size
...
+    ancestors.where(purpose_name: purpose_names)
   end
 
   def stock_plate
     return self if stock_plate?
 
-    stock_plates.order(id: :asc).last
+    # Note that it's only when we call last that we get an actual object to cache.
+    # If we cache the query to the API, then it will still be made every time we call stock_plate.
+    @stock_plate ||= fetch_stock_plate_ancestors.order(id: :asc).last
```
sdjmchattie, line 117: @andrewsparkes and @StephenHulme This is the fix for the repeated requests for stock plates.  I've added a comment to make it clearer to future maintainers, but basically we were caching the query object back in the `stock_plates` method and not the stock plate itself.  The query object is cheap to create and hasn't made a request to the API at the time we were caching it.  Instead we need to cache the response which is only obtained after you call `last` on this line.  This gives us the speed up we need AFAIK.

spec/sequencescape/api/v2/plate_spec.rb
```diff
@@ -35,27 +35,19 @@
...
-    context 'when it is not a stock plate' do
-      let(:stock_plate_names) { ['Stock platey plate stock'] }
-      let(:stock_plates) { create_list :v2_plate, 2 }
-      before do
-        allow(plate).to receive(:ancestors).and_return(ancestors_scope)
-        allow(SearchHelper).to receive(:stock_plate_names).and_return(stock_plate_names)
-        allow(ancestors_scope).to receive(:where).with(purpose_name: stock_plate_names).and_return(stock_plates)
-      end
 
-      it 'returns the stock plates because that must be cool' do
```
sdjmchattie, line 55: I think someone was having a bad day when they chose these test names.

PR #2120, opened 2024-12-10
adding a new plate purpose for multiplexed sequencing submissions, to avoid FC

sabrine33:


dasunpubudumal (commented):
Is there a integration suite update for this?

sabrine33:
> Is there a integration suite update for this?

good point, I will try to add some

config/pipelines/high_throughput_custom_pool.yml
```diff
@@ -4,5 +4,6 @@ Custom Pool:
     request_type_key: limber_lcmb
     state: passed
   relationships:
+    Library Stock Plate: LB Custom Pool #new plate purpose for multiplexed sequencing submissions, to avoid FC
```
KatyTaylor, line 7: I would put the comment as something more like 
```
These are supplied as ready-made libraries and enter the pipeline at this point.
```

I think labelling it as 'new' will become outdated, and FC needs explaining more.

config/purposes/wgs_isc_top.yml
```diff
@@ -43,6 +43,11 @@ LB Lib PCR-XP:
       id: concentrations_nm
     - name: Download Concentration (ng/ul) CSV
       id: concentrations_ngul
+Library Stock Plate: #new plate purpose for multiplexed sequencing submissions, to avoid FC
+  :asset_type: plate
+  :stock_plate: true
+  :input_plate: true
+  :presenter_class: Presenters::StockPlatePresenter
```
KatyTaylor, line 50: I would put this in its own file. As it's shared between several pipelines, it doesn't make sense in a file for a particular pipeline.

The file that it's in has no significance, technically, as they are all concatenated together. So just for readability.

KatyTaylor:
Hey, how did you decide which pipeline files to add this into? Thanks

KatyTaylor:
> I basically checked the ones used by LB-Lib-PCR-XP, as from the story they are doing a FC to create a LB-Lib-PCR-XP and then perform the multiplexing - it might be more generic way of doing it through

OK yep. Could be worth telling the customer which pipelines you've added it too, and checking that it's relevant to all of them. 

Wondering particularly about ISC because it comes in half way and it's not just doing multiplexing - there's bait capture hybridization (or something) happening.

sabrine33:
> > I basically checked the ones used by LB-Lib-PCR-XP, as from the story they are doing a FC to create a LB-Lib-PCR-XP and then perform the multiplexing - it might be more generic way of doing it through
> 
> OK yep. Could be worth telling the customer which pipelines you've added it too, and checking that it's relevant to all of them.
> 
> Wondering particularly about ISC because it comes in half way and it's not just doing multiplexing - there's bait capture hybridization (or something) happening.

Thank you, I will do, do we have a slack channel for this ? I know Liz and Siobhan are the key persons to ask..

config/pipelines/high_throughput_wgs.yml
```diff
@@ -19,5 +19,6 @@ WGS MX: # Bottom of the pipeline (Pooling and normalization)
     request_type_key:
       - limber_multiplexing
   relationships:
+    Library Stock Plate: LB Lib Pool #These are supplied as ready-made libraries and enter the pipeline at this point.
```
dasunpubudumal, line 22: If the MX stage follows the lib prep stage, how is this new plate be linked to the lib prep stage?

app/models/labware_creators/pooled_tubes_by_submission.rb
```diff
@@ -28,7 +28,7 @@ def all_wells_in_pool_passed?(pool_info)
     end
 
     def pick?(location)
-      well_locations.fetch(location).passed?
+      !(well_locations.fetch(location).failed? || well_locations.fetch(location).empty?)
```
KatyTaylor, line 31: Needed because when the multiplexing pipeline starts with a stock plate, the wells are of state 'unknown'

Issue #2131, opened 2024-12-17
Y24-521 - [BUG] Uploaded files not displayed in File tab


neilsycamore:
When files are uploaded for a plate in Limber they no longer appear in the Files tab.

Seena commented (in Slack): There was a centralise component initialisation story deployed two weeks back which changes the way components are getting loaded (https://github.com/sanger/limber/issues/2033). I think there is a chance that it might have caused this.

Relates to [FS17363](https://sanger.freshservice.com/a/tickets/17363) which contains further details

PR #2132, opened 2024-12-17
fix(file-list): Fix for uploaded files not displayed in file tab

seenanair:
Closes https://github.com/sanger/limber/issues/2131

This includes the fix for bug - Uploaded files not displayed in File tab

### DIFF
--- a/app/frontend/javascript/file-list/components/FileList.vue
+++ b/app/frontend/javascript/file-list/components/FileList.vue
@@ -37,7 +37,7 @@ export default {
         self.loading = false
       }
       xhr.send()
-    }
+    },
   },
 }
 </script>
--- a/app/frontend/javascript/file-list/components/FileList.vue
+++ b/app/frontend/javascript/file-list/components/FileList.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="list-group list-group-flush" @click="handleClick">
+  <div class="list-group list-group-flush">
     <div v-if="loading" class="spinner-dark">Updating...</div>
     <a v-for="qc_file in qc_files" :key="qc_file.uuid" class="list-group-item" :href="'/qc_files/' + qc_file.uuid">
       {{ qc_file.filename }} - {{ qc_file.created }}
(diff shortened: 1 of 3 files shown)


app/frontend/javascript/file-list/components/FileList.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="list-group list-group-flush" @click="handleClick">
```
BenTopping, line 2: Did you find out why this is not being executed when the list is clicked?

seenanair: I believe the way it was handling the click before was by using another element's click event, which iwraps this element. The previous code is as follows which was in the index file of this component
`  if (document.getElementById('files-list')) {
    /* The files-list element isn't on all pages. So only initialize our
     * Vue app if we actually find it */
    let app = new Vue({
      el: '#files-list',
      render: (h) => h(FileList),
    })
    document.getElementById('files-tab-link').addEventListener('click', function () {
      app.$children[0].fetchData()
    })
  }
})`
I think it’s better to avoid that dependency here. Also, in the scenario where we are in the 'Files' tab and upload a new file, the page currently reloads, which causes the focused tab to switch to another one—likely the default tab. Therefore mount will be called anyway which avoids the need to do handle click. Hope it makes sense

seenanair: But, I think we are reloading the the page, the components will be re-mounted any way, right?

seenanair: Te component is loaded this way and the click was originally linked to the 'link' element with id 'files-tab-link'
 <li class="nav-item">
      <a id='files-tab-link' class="nav-link" href="#files_tab" data-toggle="tab" role="tab" data-plate-view="files-view">Files</a>
    </li>

seenanair: Yes, console.log confirms it

seenanair: @BenTopping, @StephenHulme - It would be great if one of you could also test whether the file upload is correctly displayed once it is in UAT, please.

BenTopping (approved):

seenanair:
> Would it benefit from a unit test? Otherwise looks good 👍

Added unit tests 79b5753f88ce147721bc40d7341880e655787314

PR #2156, opened 2025-01-13
Y24-102 tube rack banking bed verification

### LINKED ISSUE #1699: Y24-102 Tube Rack Banking Bed verification
**User story**
As PSD, we would like to support bed verification using the versioned tube rack model proposed in Reuse of TubeRacks document. LRC Bank Seq and LRC Bank Spare tube racks are destination labware in this bed verification.

Abdullah, Katy, Andrew

Once [Y24-088](https://github.com/sanger/limber/issues/1713) and its dependencies are complete, we can ask Abby if she will UAT it. Could send some screenshots for earlier feedback.

**Acceptance criteria**
- [x] [DPL-829](https://github.com/sanger/limber/issues/1333) acceptance criteria using the TubeRack model, retiring the TubeRackWrapper implementation.
- [x] Update the integration suite tests if necessary.

It is not possible to release the implementation of this story in isolation. There are epic branches in use for SS, Limber and Int Suite to branch from and to.

**Dependencies**
Story relationships are shown in the Lucidchart diagram [TubeRack Stories](https://lucid.app/lucidchart/d88bdcfd-9628-4fec-8b5a-453f680ee06b/edit?viewport_loc=-1051%2C-316%2C2770%2C1486%2C0_0&invitationId=inv_13d74286-1cc2-4c2f-aaf0-30887764601c) .

**Additional context**
The existing bed verification for scRNA Core uses a wrapper to represent Tube Racks in bed verification stage. See [DPL-829](https://github.com/sanger/limber/issues/1333). It queries the parent's downstream tubes to find out which tubes are on the tube rack. It transforms the given robot configuration to support optional tube racks. See `app/models/robots/plate_to_tube_racks_robot.rb`

The versioned tube rack model is researched in [DPL-1011](https://github.com/sanger/limber/issues/1482) and the model is summarised in the Google Document[ DPL-1011 Re-use of tube racks](https://docs.google.com/document/d/1rvPXeCk5rBEf72VkchHCD2FRmWANboke2pln_Lt_j5U/edit?usp=sharing)

This story is part of re-implementation of [DPL-811 Transfer PBMC isolations into FluidX tube racks](https://github.com/sanger/limber/issues/1314) (Banking) using the versioned tube rack model.

andrewsparkes:
Closes #1699


app/models/robots/bed/plate_to_tube_racks_bed.rb
```diff
@@ -4,19 +4,19 @@ module Robots::Bed
...
   class PlateToTubeRacksBed < Robots::Bed::Base
     # Updates the metadata of the labware with the robot barcode.
     # This method is called inside the robot controller's start action for
-    # tube-rack wrappers and it sets the created_with_robot metadata field.
+    # Tube Racks and it sets the created_with_robot metadata field.
     #
     # @param robot_barcode [String] the robot barcode
     # @return [void]
     #
     def labware_created_with_robot(robot_barcode)
```
yoldas, line 17: The robot controller has a method with the same name. The controller checks if the bed has a custom method to set the robot barcode in the labware metadata; otherwise it uses its own. The reason it is implemented that way was because TubeRackWrapper was not a real labware. The bed was setting it on the labware's tubes instead. It is used for checking if the labware should be returned to the same robot for the next bed verification if configured with verify_robot. I noticed now that in the wrapper implementation, custom_metadata_collection call on a wrapper should have delegated the to last tube there to be able to use this original robot condition.

Because we are removing the wrapper now, and the tube rack is a real labware, I think we should remove the `labware_created_with_robot` custom method from the bed and rely on the controller method to write the metadata on the tube rack. Otherwise, if configured with verify_robot setting in the robot config, it will fail because the lookup will happen in the custom metadata collection of the tube rack but the robot barcode was written on that of tubes.

yoldas (changes requested):
We have a discussion point I think.

app/models/robots/bed/plate_to_tube_racks_bed.rb
andrewsparkes: Thank you.
I have written the metadata to both tube rack and tubes now, so it's in both places. I think it's a handy bit of extra information to have on a tube in case it gets split from it's rack.
Added a unit test as well.

yoldas: Keeping the custom method to write the custom metadata on tubes as well as the tube rack is a very good idea. Thanks.

yoldas:
There are some comments need updating (I have searched the text 'wrapper' in `./app/models/robots`):
- top-level comments of `app/models/robots/plate_to_tube_racks_robot.rb` 
-  description of the `find_bed_labware` method
- description of the `add_tube_racks_to_labware_store` method
- description of the `labware_store` method
- `return` value description of the `child_labware` method

yoldas (approved):
Real bed verification with tube racks! Well documented and tested. Looks great.

PR #2166, opened 2025-01-17
Y25-013 [PR] bug show informative error messages from pooling algorithm

yoldas:
Closes #

Add validation to check if pools can be built and convert exceptions to error messages

### DIFF
--- a/app/models/concerns/labware_creators/donor_pooling_validator.rb
+++ b/app/models/concerns/labware_creators/donor_pooling_validator.rb
@@ -10,6 +10,7 @@ module LabwareCreators::DonorPoolingValidator
     validate :source_plates_must_exist
     validate :wells_with_aliquots_must_have_donor_id
     validate :wells_with_aliquots_must_have_cell_count
+    validate :validate_pools_can_be_built
   end
 
   SOURCE_BARCODES_MUST_BE_ENTERED = 'At least one source plate must be scanned.'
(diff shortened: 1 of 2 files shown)


app/models/concerns/labware_creators/donor_pooling_validator.rb
```diff
@@ -170,4 +171,25 @@ def invalid_wells_hash(invalid_wells)
...
+  # creation controller will then add the error messages to the flash.
+  #
+  # The advantages of this approach are:
+  # - If the pools cannot be built, the create_labware! method will not be
+  # called at all and the exceptions will be converted to error messages.
+  # - If the pools can be built, the create_labware! method will use the cached
+  # result to generate the transfers.
+  #
+  # @return [void]
+  def validate_pools_can_be_built
```
BenTopping, line 187: Had a look in the code base and its quite hard to follow the calls so apologies if this is a silly question but can we just validate pools directly with `validate :pools`

yoldas: The pools method calls build_pools and caches it (||=). The labware creator uses the result to create the child  labware and transfer requests. The problem with that is pools calculations raise exceptions, which produces an error page which does not mean anything to the user. 

The idea of the story and the implementation in this PR is that we can call the pools method inside a validator and catch if there are any exceptions. If there are exceptions, add them to the errors collection of the labware creator. By doing that we make the object invalid, i.e. valid? call on it will return false because the error collection is not empty.

The creation controller holds an instance of the labware creator and checks what its save method returns. The save method of the creator calls valid? and then create_labware. **If we do not have this new validator**, the valid? call will return true and exceptions will be raised later in create_labware. There will be even a redundant child labware creation as well as the bad-looking something-went-wrong exception page.

However, now **with this new validator**, as we transformed the exception into error messages, the valid? call will fail before proceeding with the actual labware and transfers creation.

Why can't we just call do `validate :pools` ?
No, because
1. the pools method is not a validation method. Its purpose is calculating the pools.
2. the pools method wraps build_pools method for caching the result of the computations if successful.
3. pools method can raise exceptions, which is exactly what we are trying to intercept
4. `validate :validate_pools_can_be_built` declaration is necessary to have a dedicated validator for this case because it has a special logic.
5. The `validate_pools_can_be_built` validator needs try-pools-catch-exceptions logic to be done in the validator.

yoldas: I might add, of course, if save returns false on the labware creator, the creation controller puts the errors into flash and shows in the page. If save returns true, that is the happy path, child creation with transfers based on pools.

BenTopping (approved):

PR #2176, opened 2025-01-20
Y24-038 - Changed "Custom Pool Norm" to "Custom Norm" in purposes

### LINKED ISSUE #1656: Y24-038 Change printed plate name for LCMB Custom plates
As team lead for RNA (Liz E) it is very hard for the lab team, when making pools, to determine the correct label to put on the tube as the whole name doesn’t fit on the tube.
LCMB custom pool and LCMB custom Pool Norm.  both of these samples have the same human readable name on the label ‘Custom Pool’ making it difficult for us to know which label is which when we print them.
pool_label_LCMB

Please can we remove the word custom Or Make one Custom Pool and one Custom Norm.

Raised via [RT799627](https://rt.sanger.ac.uk/Ticket/Display.html?id=799627)

SHIV5T3R:
Closes #1656 

- Changed the purpose name for `Custom Pool Norm` to `Custom Norm` so it can fit on labels.
- `Custom Pool` has not been changed.

### DIFF
--- a/config/pipelines/high_throughput_lcmb.yml
+++ b/config/pipelines/high_throughput_lcmb.yml
@@ -15,4 +15,4 @@ LCMB: # Top of the pipeline (Library Prep)
     LCMB End Prep: LCMB Lib PCR
     LCMB Lib PCR: LCMB Lib PCR-XP
     LCMB Lib PCR-XP: LCMB Custom Pool
-    LCMB Custom Pool: LCMB Custom Pool Norm
+    LCMB Custom Pool: LCMB Custom Norm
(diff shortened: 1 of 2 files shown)


dasunpubudumal (commented):
Looks good, but might have to be tested on-site with our printers?

Maybe try to deploy via branch and test it.

andrewsparkes (approved):
An issue with this: 
When you deploy Limber it will create a new 'LCMB Custom Norm' purpose. It will not rename the existing one.
So when a user looks at Tubes made with the existing name in the LIMS, they will still have purpose 'LCMB Custom Pool Norm'. Only new ones created after the change will have the new name.
You would have to do a data migration to change those legacy tubes to be your new purpose.
And then potentially another to delete the old 'LCMB Custom Pool Norm' purpose.

Alternatively, before this deployment, do a data migration or direct database update to change the existing purpose name. Then when you deploy this it will see the name matches and not create a new one.

PR #2196, opened 2025-01-30
Limber dependencies - update rails to 7 2 2 1

andrewsparkes:

Update Limber to Rails 7.2.2.1
- removed secrets and replaced with credentials in deployment project
- updated framework defaults to 7.2 and removed old defaults files
- removed unneeded requires statements in specs
- attempted to fix some autoload issues in lib directory
- updated factory_bot to 6.5 to fix error:
```
NameError:
  uninitialized constant #<Class:ActiveSupport::Delegation>::Inflector
```


app/controllers/search_controller.rb
```diff
@@ -5,8 +5,6 @@ class SearchController < ApplicationController
   class InputError < StandardError
   end
 
-  before_action :check_for_login!, only: [:my_plates] # rubocop:todo Rails/LexicallyScopedActionFilter
```
yoldas, line 8: Does this remove login requirement for my plates?

andrewsparkes: With that line in it couldn't identify what 'my_plates' was referring to.
There is no reference to 'my_plates' in the code base that I can see, maybe some rails magic that no longer works?

bin/setup
```diff
@@ -2,10 +2,11 @@
 require "fileutils"
 
 # path to your application root.
-APP_ROOT = File.expand_path('..', __dir__)
+APP_ROOT = File.expand_path("..", __dir__)
+APP_NAME = "limber"
 
 def system!(*args)
-  system(*args) || abort("\n== Command #{args} failed ==")
+  system(*args, exception: true) || abort("\n== Command #{args} failed ==")
```
yoldas, line 9: I think the second part of the shortcut-or is redundant now because of the `exception: true` argument.
```
limber(dev)> system('ls nowhere')
ls: nowhere: No such file or directory
=> false
limber(dev)> system('ls nowhere', exception: true)
ls: nowhere: No such file or directory
(limber):2:in `system': Command failed with exit 1: ls (RuntimeError)
	from (limber):2:in `<main>'
```

andrewsparkes: Maybe I should have left it as it was. I think the rails update suggested this change.

andrewsparkes: reverted

yoldas: Thanks. Yes, if we want to preserve the error message and printing the command and args, we should keep it. Otherwise, it just prints a stack trace, which may not be helpful. Note: the wrapping method is used by bin/setup and bin/update scripts only for the development environment.

PR #2212, opened 2025-02-07
Y25-025 - Maintain customer-requested number of repeats

### LINKED ISSUE #2164: Y25-025 - Maintain customer-requested number of repeats (Limber)
**User story**

As a user of the scRNA Core pipeline, I would like the 'number of cells per chip well' to be adjusted at point of pooling, if there have been well failures, so that the number of repeats and cell counts that were possible at submission are still achievable.

This is so that we can still deliver the service that was requested as much as possible - we believe it's more important that the number of repeats and cell counts is maintained, rather than number of cells per chip well. Also so that we don't have to go back to the customer at this point, because we cannot pause the pipeline.

Katy, Abdullah, Abby

Abby

**Acceptance criteria**

When creating LRC PBMC Pools plate (Limber), after calculating the number and size of pools, for any pools with 5-8 samples:
- [ ] Check what the original "allowance band" met at point of submission was (request metadata - it should be the same for all requests in the pool) 
- [ ] Calculate whether there is still enough material to meet the same allowance band (will only have changed if there have been well failures). (Technically, no need to check if the original band was "1 pool attempt, 1 count" - see additional context.)
- [ ] If not, the 'number of cells per chip well' to be used going forward (for each pool) is not taken from the value requested in the submission, but instead selected from the table below:
Image
- [ ] As is done currently, the selected 'number of cells per chip well' is saved into polymetadata for each pool - this is necessary even if we are using the requested value from the submission.

**Dependencies**

This story is a change to / enhancement of https://github.com/sanger/limber/issues/1992

(Soft) blocked by sanger/sequencescape#4652 

**Additional context**

> It is only possible for someone to request '1 pool attempt, 1 count' if they only want 5 samples/pool (>5 samples/pool and the max cells/chip well of 90K always results in at least '1 pool attempt, 2 counts' or more).
If 1 sample then failed the whole run would become no longer viable so we don't have to adjust any values for this, the fail would be captured by the <5 samples/pool flag.
>
> If there was not enough at submission that would be caught by our restrictions to no. cells/chip well and samples/well size. If there was not enough at pooling recheck that would only be due to not enough samples/well which would fail as mentioned.

BenTopping:
Closes #2164 

- Changes donor_pooling_calculator allowance calculations.
- Updates scrna config allowance_table to handle multiple allowance bands


BenTopping:
Initial code is ready for review. There may be changes after Sequencescape side is ready to integrate correctly.

app/models/concerns/labware_creators/donor_pooling_calculator.rb
```diff
@@ -261,18 +288,29 @@ def calculate_chip_loading_volume(num_cells_per_chip_well)
...
+        scrna_config[:wastage_volume]
+    when '2 pool attempts, 1 count'
+      (chip_loading_volume * scrna_config[:desired_number_of_runs]) + scrna_config[:volume_taken_for_cell_counting] +
+        scrna_config[:wastage_volume]
+    when '1 pool attempt, 2 counts'
+      chip_loading_volume + (scrna_config[:desired_number_of_runs] * scrna_config[:volume_taken_for_cell_counting]) +
+        scrna_config[:wastage_volume]
+    when '1 pool attempt, 1 count'
+      chip_loading_volume + scrna_config[:volume_taken_for_cell_counting] + scrna_config[:wastage_volume]
+    end
```
KatyTaylor, line 313: I've asked the same of Sabrine in her story - can we get rid of the `desired_number_of_runs` constant please, and just hardcode `2` instead? It's misleading, as it isn't necessarily the desired number of runs - the customer might have only supplied enough for 1 run, and that's OK.

app/models/concerns/labware_creators/donor_pooling_calculator.rb
```diff
@@ -261,18 +288,29 @@ def calculate_chip_loading_volume(num_cells_per_chip_well)
...
 
-  # This method calculates the full allowance volume for a given chip loading volume.
+  # This method calculates the allowance volume for a given chip loading volume and allowance band.
   # It uses configuration values from the scrna_config for the desired
   # number of runs, the volume taken for cell counting, and the wastage volume.
   #
   # @param chip_loading_volume [Float] the chip loading volume
-  # @return [Float] the full allowance volume
-  def calculate_full_allowance(chip_loading_volume)
+  # @return [Float] the allowance volume
```
KatyTaylor, line 296:
```suggestion
  # @return [Float] the volume of material required to do the number of runs and cell counts specified in the allowance band
```

app/models/concerns/labware_creators/donor_pooling_calculator.rb
```diff
@@ -226,14 +253,14 @@ def adjust_number_of_cells_per_chip_well(count_of_samples_in_pool, number_of_cel
     # calculate chip loading volume
     chip_loading_volume = calculate_chip_loading_volume(number_of_cells_per_chip_well)
 
-    # calculate full allowance
-    full_allowance = calculate_full_allowance(chip_loading_volume)
+    # calculate allowance
+    allowance = calculate_allowance(chip_loading_volume, allowance_band)
```
KatyTaylor, line 257:
```suggestion
    volume_needed = calculate_allowance(chip_loading_volume, allowance_band)
```

Just aware there's a lot of jargon - maybe we can rename some of these variables to make it more intuitive to understand?

app/models/labware_creators/donor_pooling_plate.rb
```diff
@@ -144,7 +144,7 @@ def transfer_material_from_parent!(dest_uuid)
         transfer_requests_attributes: transfer_request_attributes(dest_plate),
         user_uuid: user_uuid
       )
-      determine_if_pools_have_full_allowance(dest_plate)
+      determine_if_pools_have_allowance_band(dest_plate)
```
KatyTaylor, line 147:
```suggestion
      determine_if_pools_have_enough_material_for_allowance_band(dest_plate)
```

app/models/labware_creators/donor_pooling_plate.rb
```diff
@@ -226,21 +226,21 @@ def build_pools
       built_pools
     end
 
-    # This method determines if the pools have full allowance.
+    # This method determines if the pools have the allowance band.
```
KatyTaylor, line 229:
```suggestion
    # Given some wells may have been failed, this method determines if we still have enough material to meet the number of runs/cell counts requested in the submission.
```

app/models/concerns/labware_creators/donor_pooling_calculator.rb
```diff
@@ -226,14 +253,14 @@ def adjust_number_of_cells_per_chip_well(count_of_samples_in_pool, number_of_cel
...
 
     # do not adjust existing value if we have enough volume
-    return number_of_cells_per_chip_well if final_suspension_volume >= full_allowance
+    return number_of_cells_per_chip_well if final_suspension_volume >= allowance
 
     # we need to adjust the number of cells per chip well according to the number of samples
-    Rails.application.config.scrna_config[:full_allowance_table][count_of_samples_in_pool]
+    Rails.application.config.scrna_config[:allowance_table][allowance_band][count_of_samples_in_pool]
```
KatyTaylor, line 263: I think we should put something in to check this hasn't returned `nil` - I guess it should error out if it does. It shouldn't hit those values because, for instance, if you have 8 samples, you should always have enough material for '2 pool attempts, 1 count'.

KatyTaylor (changes requested):
Looks great! I can't see anything that won't work as intended. I've left several comments but it's mostly naming-related. And one error handling thing. Thanks.

app/models/concerns/labware_creators/donor_pooling_calculator.rb
BenTopping: 1. Yeah happy to make that changes, feels more consistent.
2. Good suggestion, an enum makes sense.

app/models/concerns/labware_creators/donor_pooling_calculator.rb
BenTopping: Yes I wasn't sure if this needed specific handling. I suspect it will error later down the line but a descriptive error would be useful here.

BenTopping: Added error handling in 
https://github.com/sanger/limber/pull/2212/commits/2279f54606c333c75bd71c2511f50b453ebeac57

KatyTaylor (approved):
Thanks for making the changes, looks good to me. Obviously merging / releasing is dependent on Sabrine's, and they need to be tested together both:
a) manually, and
b) in the Integration Suite test (maybe... can't remember if we test this feature or not currently)

app/models/concerns/labware_creators/donor_pooling_calculator.rb
```diff
@@ -202,20 +212,44 @@ def number_of_cells_per_chip_well_from_request(pool)
...
+
+    # fetch request metadata for number of cells per chip well from first aliquot in the source well
+    # (it should be the same value for all the aliquots in all the source wells in the pool)
+    allowance_band = source_well&.aliquots&.first&.request&.request_metadata&.allowance_band
+
+    if allowance_band.blank?
+      raise StandardError,
+            "No request found for source well at #{source_well.location}, cannot fetch allowance band " \
+              'well metadata for allowance band calculations'
+    end
```
sabrine33, line 240: This is breaking the integration tests—probably the check is not needed here. 
May be we can add the check before running `adjust_number_of_cells_per_chip_well`.
Or may be we need to update the integ suite input data, so it is always present

PR #2285, opened 2025-03-26
Y25-155 - PBMC Bank tubes content report CSV to use Tube Racks

### LINKED ISSUE #2232: Y25-155 - PBMC Bank tubes content report CSV to use Tube Racks
**User story**

Update the [PBMC Bank tubes content report CSV](https://github.com/sanger/limber/blob/develop/app/views/exports/pbmc_bank_tubes_content_report.csv.erb)  to use Tube Racks. You can model this off the changes done in https://github.com/sanger/limber/issues/2070. This is a follow on from https://github.com/sanger/limber/issues/1439 to support the new tube rack changes.

Andrew S, Katy T, Ben T, Abdullah Y

**Acceptance criteria**

- [x] Update pbmc_bank_tubes_content_report.csv.erb export to use tube racks instead of tube metadata. See issue linked above for similar changes.

**Dependencies**
- This work should be done off the tube rack epic branch. https://github.com/sanger/limber/pull/2134

BenTopping:
Closes #2232

- Updates PBMC Bank tubes content report to use tube racks.

### DIFF
--- a/app/views/exports/pbmc_bank_tubes_content_report.csv.erb
+++ b/app/views/exports/pbmc_bank_tubes_content_report.csv.erb
@@ -21,26 +21,6 @@
 <%
   rows_array = []
 
-  # Function for finding a racked_tube given a tube and set of tube racks
-  def find_racked_tube_in_tube_racks(tube_racks, tube_id)
-    tube_racks.each do |tube_rack|
-      racked_tube = tube_rack.racked_tubes.find { |rt| rt.tube.id == tube_id }
-      return racked_tube if racked_tube
-    end
-    nil
-  end
-
-  # Fetch the tube racks from the plate
-  tube_racks = @plate.children.select { |child| child.type == 'tube_racks' }.map do |tr|
-    Sequencescape::Api::V2.tube_rack_with_custom_includes(
-      'racked_tubes.tube',
-      nil,
-      barcode: tr.labware_barcode.machine
-    )
-  end
-
-  return if tube_racks.nil?
-
   @plate.wells_in_columns.each do |source_well|
     # skip if well empty (not all wells have samples in them)
     next if source_well.empty?
@@ -52,11 +32,6 @@
     next if source_well.downstream_tubes.empty?
 
     child_tube = source_well.downstream_tubes.last
-    racked_tube = find_racked_tube_in_tube_racks(tube_racks, child_tube.id)
-
-    # If this is nil, then the tube is not in the tube racks, likely something has gone wrong
-    next if racked_tube.nil?
-    
     child_tube_purpose = child_tube.purpose.name
 
     sample_uuid = source_well.aliquots.first.sample.uuid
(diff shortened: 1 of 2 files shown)


app/views/exports/pbmc_bank_tubes_content_report.csv.erb
```diff
@@ -30,25 +51,13 @@
     # skip if no downstream tubes for this well (not set up children in Limber yet)
     next if source_well.downstream_tubes.empty?
 
-    child_tube_v1 = source_well.downstream_tubes.last
```
KatyTaylor, line 33: Was this naming actually not accurate any more? Presumably `@plate.wells_in_columns` used to use the v1 API but now (since Stuart) uses the v2 API?

app/views/exports/pbmc_bank_tubes_content_report.csv.erb
```diff
@@ -20,6 +20,27 @@
...
+  tube_racks = @plate.children.select { |child| child.type == 'tube_racks' }.map do |tr|
+    Sequencescape::Api::V2.tube_rack_with_custom_includes(
+      'racked_tubes.tube',
+      nil,
+      barcode: tr.labware_barcode.machine
+    )
+  end
+
+  return if tube_racks.nil?
+
```
KatyTaylor, line 43: Just wondering - we _could_ get rid of most of this stuff. We needed to query the racked tubes in that other file you did because the file included the tube rack barcode. This file doesn't actually include it in any of the columns. 

So the only reason to query it here is for this line: `next if racked_tube.nil?`

Not really sure if that line is needed - I don't see how it would be possible to get in the situation where you've got a `downstream_tube` that isn't part if the tube rack. Unless it's from before we make the change to tube racks... in which case, it would be a bonus if this file still works.

Would be much simpler to read without all this stuff - what do you think?

spec/views/exports/pbmc_bank_tubes_content_report.csv.erb_spec.rb
```diff
@@ -308,166 +301,27 @@
...
-      let(:dest_tube6) do
+      let(:spr_tube_rack) do
         create(
-          :v2_tube_with_metadata,
-          uuid: dest_tube6_uuid,
-          barcode_prefix: 'FX',
-          barcode_number: 9,
-          aliquots: [dest_aliquot6],
-          name: 'SPR:NT2P:D1',
-          custom_metadatum_collection: useless_custom_metadata
```
KatyTaylor, line 470: Nice - lots of code deletion 😅

KatyTaylor (commented):
Looks fine I think - just a suggestion to consider

app/views/exports/pbmc_bank_tubes_content_report.csv.erb
BenTopping: Yes this naming was incorrect, its now of type `Sequencescape::Api::V2::Tube`

app/views/exports/pbmc_bank_tubes_content_report.csv.erb
BenTopping: Yes I agree, good spot. Makes it much simpler 🙌

KatyTaylor (approved):
Nice thanks

PR #2299, opened 2025-04-02
Y25-202 - scRNA Library Prep Chromium BCR and TCR pipeline config

### LINKED ISSUE #2265: Duplication of and amendment to existing VDJ (TCR & BCR) pipelines
**User story**

As a user of the scRNA Core cDNA Prep pipeline, I would like new functionality to be added to enable users to process 5' GEM-X TCR and/or BCR samples.

Samples due to use this pipeline in May (likely end of). Ideally would have these changes in place in April to allow time for training out new process

Katy, Abby

Abby

**Acceptance criteria**

- [ ] users to submit LRC GEM-X 5p Cherrypick plates for 5' GEM-X TCR or 5' GEM-X BCR library prep
- [ ] a different workflow name for the csv file generated for the LRC GEM-X 5p BCR/TCR Enrich1 2XSPRI to LRC GEM-X 5p BCR/TCR Enrich2 2XSPRI plate step
- [ ] a set 10ul sample volume and 0ul dilution volume for the LRC GEM-X 5p BCR/TCR Enrich2 2XSPRI to LRC GEM-X 5p BCR/TCR Dil 2 csv
- [ ] binning of the samples for the LRC GEM-X 5p BCR/TCR Enrich2 2XSPRI to LRC GEM-X 5p BCR/TCR Dil 2 csv to match the table below based on uploaded quants for the former plate
- [ ] the LRC GEM-X 5p BCR/TCR  Pool Norm tubes to be combined with LRC GEM-X 5p GE Pool Norm and LRC GEM-X 5p CITE Pool Norm (separate story in to create this) tubes, generating a LB Lib Pool Norm tube

**Additional context**

Image

**Deployment notes**

Add any details that should be noted when deploying this feature. For example, is a database migration needed, a specifc rake task or other script that needs to be run, a dependency on another application or service that needs to be released first etc?
These notes are intended to ensure the developers releasing the feature know what needs to be done.

BenTopping:
Closes #2265, #2280 

- Adds `scRNA Core Chromium GEM-X 5p BCR` and `scRNA Core Chromium GEM-X 5p BCR MX` pipelines
  - Adds pipeline purposes and bed verifications.
- Adds `scRNA Core Chromium GEM-X 5p TCR` and `scRNA Core Chromium GEM-X 5p TCR MX` pipelines
  - Adds pipeline purposes and bed verifications.
- Adds custom workflows for BCR and TCR pipelines.
 

Sequencescape PR: https://github.com/sanger/sequencescape/pull/4814
Integration suite PR: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/236


app/models/labware_creators/well_filter_allowing_partials.rb
```diff
@@ -15,7 +15,7 @@ def filter_by_state(requests)
   end
 
   def filter_requests(requests, well)
-    return nil if well.requests_as_source.empty?
```
BenTopping, line 18: This has been looked at before by @andrewsparkes: https://github.com/sanger/limber/commit/e4d9e2f8f3de87e4d9dfa60bf79c8385087cd059. I don't think the existing behaviour is correct because this checks that the parent well contains the request (source of request is parent well). This means this can only be valid if the plate is the first plate in the pipeline (submitted on) and I don't think that is correct? I think it just happens that the other pipelines that use this meet that condition.

app/models/labware_creators/partial_stamped_plate.rb
```diff
@@ -25,7 +25,7 @@ def dilutions_calculator
     # The well filter will be used to identify the parent wells to be taken forward.
     # Filters on request type, library type and state.
     def well_filter
-      @well_filter ||= WellFilterAllowingPartials.new(creator: self, request_state: 'pending')
```
BenTopping, line 28: This pipeline's request are 'started' not pending. I think `started` is safe to use here as is it is also an `ongoing` request state.

config/purposes/scrna_core_library_prep_chromium_bcr.yml
```diff
@@ -0,0 +1,83 @@
+# This configuration file is for the Bespoke Chromium Automated pipeline for scRNA core library prep VDJ option BCR
+# library type requests.
```
KatyTaylor, line 2: I would take out the 'Bespoke Chromium Automated pipeline' bit from the comment - I know what you're getting at, because this is cloned from that pipeline - but I think we just call this one 'the BCR branch of the scRNA Core library prep pipeline' or something, to avoid confusion.

config/robots.rb
```diff
@@ -4253,4 +4253,198 @@
       from 'BGE Cap Lib PCR', bed(1)
       to 'BGE Cap Lib PCR XP', bed(9)
     end
+
+    # scRNA pipeline
```
KatyTaylor, line 4257:
```suggestion
    # scRNA Core pipeline
```

Pedantic sorry, but there is another pipeline called scRNA :/

config/robots.rb
```diff
@@ -4253,4 +4253,198 @@
...
+          purpose: 'LRC GEM-X 5p BCR Post Lig 1XSPRI',
+          states: ['pending'],
+          label: 'Bed 3',
+          target_state: 'passed',
+          parent: bed(13).barcode
+        }
+      }
+    )
+
+    # scRNA pipeline
```
KatyTaylor, line 4354:
```suggestion
    # scRNA Core pipeline
```

KatyTaylor (commented):
Still need to think about that requests one

app/models/labware_creators/partial_stamped_plate.rb
BenTopping: Yeah that was my thinking to. I think this is more a prevention mechanism for passed requests than anything else.

config/pipelines/high_throughput_scrna_core_library_prep_chromium_bcr.yml
```diff
@@ -0,0 +1,23 @@
+---
+scRNA Core Chromium GEM-X 5p BCR:
+  pipeline_group: scRNA Core Library Prep
```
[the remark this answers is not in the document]
BenTopping: Sure. `scRNA Core BCR Library Prep` and `scRNA Core TCR Library Prep` sound ok?

config/purposes/scrna_core_library_prep_chromium_bcr.yml
```diff
@@ -0,0 +1,83 @@
...
+        colour: 2
+      - min: 150.0
+        max: 500.0
+        pcr_cycles: 7
+        colour: 3
+      - min: 500.0
+        max: 1_000.0
+        pcr_cycles: 6
+        colour: 4
+      - min: 1_000.0
```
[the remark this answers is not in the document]
BenTopping: Yes this has been confirmed with Abby. Anything above 1000 is 5 pcr cycles

config/exports/exports.yml
```diff
@@ -250,3 +250,33 @@ hamilton_gem_x_5p_chip_loading:
...
+  plate_includes: wells.transfer_requests_as_target.source_asset
+  workflow: 10X GEM-X 5' VDJ Post Target Enrichment 2 Double SPRI
+hamilton_lrc_gem_x_5p_bcr_enrich2_2xspri_to_lrc_gem_x_5p_bcr_dil_2:
+  csv: hamilton_fixed_volume_dilutions
+  plate_includes: wells.transfer_requests_as_target.source_asset
+  workflow: Sample Dilution
+hamilton_lrc_gem_x_5p_bcr_dil_2_to_lrc_gem_x_5p_bcr_post_lig_1xspri:
+  csv: hamilton_plate_stamp
+  plate_includes: wells.transfer_requests_as_target.source_asset
+  workflow: 10X GEM-X 5' VDJ Post Ligation SPRI
```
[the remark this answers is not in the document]
BenTopping: Yes this was confirmed with Abby.

app/models/labware_creators/concentration_binned_plate_stamp.rb
```diff
@@ -0,0 +1,86 @@
...
+  # |B1| conc=1.2   x10=12  (bin 1)  |D1|E1|  |
+  # +--+--+--~  +                    +--+--+--~
+  # |C1| conc=67.2  x10=672 (bin 3)  |  |G1|  |
+  # +--+--+--~                       +--+--+--~
+  # |D1| conc=2.1   x10=21  (bin 1)  |  |  |  |
+  # +--+--+--~                       +--+--+--~
+  # |E1| conc=33.7  x10=337 (bin 2)  |  |  |  |
+  # +--+--+--~                       +--+--+--~
+  # |G1| conc=25.9  x10=259 (bin 2)  |  |  |  |
+  class ConcentrationBinnedPlateStamp < StampedPlate
```
andrewsparkes, line 57: I don't like the naming here. 
'Binning' means rearranging the samples on transfer into the child plate into new well locations depending on some parameter (by concentration in the original).
'Stamp' means source well = destination well, i.e. A1 to A1, B1 to B1 etc. They are not compatible terms.

andrewsparkes (commented):
Can we discuss this story? You can't have a binned stamp, so confused what you are trying to do.
Does the original not work for you? it just looks for wells with the right request type (partial submissions for the original pipeline). But if all the wells have the right request won't it just take them all? Assuming that's what you are after.

app/models/labware_creators/well_filter_allowing_partials.rb
andrewsparkes: Yeah we can discuss.
The labware creator was built for a specific plate at a branch point on the cherrypick plate in a specific pipeline, and works for that. It's not been designed to be generic.
I cannot remember why I had to change that line to get it to work though!
active_requests seems to use requests_as_source about 3 or 4 levels of method calls in. it's horribly complicated though.

andrewsparkes: for chromium they would create the cherrypick, then do partial submissions on the wells in the cherrypick depending on which of 3 library prep types they wanted to do, hence partials. plus the same well could be submitted for more than one library prep. plus they might come back and repeat.
the well filter is having to filter on request type, library type and state to identify the right wells.

andrewsparkes: the requests were not in the the aliquots on the cherrypick plate, because as you spotted that is the submission plate, they are in requests_as_source only.

BenTopping: Yeah Katy and I discussed this yesterday and thought it best to create a new labware creator instead of modifying this because at the step we are interested in we aren't using partials so it doesnt make sense to use a partial stamp creator.

app/models/labware_creators/concentration_binned_plate_stamp.rb
BenTopping: Ah ok I wasn't aware thats what binning meant. The only difference in this labware creator is that it inherits from StampedPlate instead of PartialStampedPlate.

BenTopping:
> Can we discuss this story? You can't have a binned stamp, so confused what you are trying to do. Does the original not work for you? it just looks for wells with the right request type (partial submissions for the original pipeline). But if all the wells have the right request won't it just take them all? Assuming that's what you are after.

Yes sure. The only difference in this new labware creator compared to the existing one (ConcentrationBinnedPlate) is that it inherits from StampedPlate instead of PartialStampedPlate. This is because partial stamped plate uses a well filter that checks the parent plate is a submitted plate (start of the pipeline / has requests_as_source present). Which this pipeline does not match because we are doing this concentration binning half way down the pipeline. So in essence its the same but  we are just taking the whole plate forward.

andrewsparkes (approved):
Looks good to me

.rubocop_todo.yml
```diff
@@ -79,6 +79,7 @@ RSpec/ContextWording:
     - 'spec/models/labware_creators/common_file_handling/csv_file_for_tube_rack_spec.rb'
     - 'spec/models/labware_creators/common_file_handling/csv_file_for_tube_rack_with_rack_barcode_spec.rb'
     - 'spec/models/labware_creators/concentration_binned_plate_spec.rb'
+    - 'spec/models/labware_creators/concentration_binned_full_plate_spec.rb'
```
andrewsparkes, line 82: strictly we should try to reduce these todo's not add more. although adding disables, which tends to be what i do, is not that much better.

PR #2306, opened 2025-04-04
Y25-241

### LINKED ISSUE #2301: Y25-241 - Fix UAT v2 API Searches controller bug
**Describe the Housekeeping**

A client error happens when from Limbers Search page clicks “All Ongoing Plates” or “My Plates”. The cause of this seems to be: the GET request contains “options” which are the many Plate Purposes to filter the plates by. Locally, when reducing this options list of Plate Purposes to ~5 it works, but on UAT there are 385 Plate Purposes, Training: 358, and Prod: 358. We can test deploying to Training to see if it works. If it does, we need to find another fix, as the API v1 used to handle this amount of options, but the v2 API doesn’t. 

```
A JsonApiClient::Errors::ClientError occurred in search#ongoing_plates:

  JsonApiClient::Errors::ClientError
  app/sequencescape/sequencescape/api/v2/plate.rb:45:in `find_all'
```

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**

harrietc52:
Closes #2301 

-   Default plate purposes to an empty array, because there are hundreds of plate purposes in the database, and the GET query was exceeding the maximum number of bytes. Instead, the page can load, then the user can then select a purpose from the list, which will be paginated.
- Intergration Suite (on Training) passed [here](https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/pipelines/375373)  

### DIFF
--- a/app/models/ongoing_plate.rb
+++ b/app/models/ongoing_plate.rb
@@ -21,7 +21,7 @@ class OngoingPlate < AssetSearchForm
   # @note
   # This is currently only being used by the SearchController, as other usages pass in `purposes`,
   #   and therefore don't use the default_purposes method.
-  #   We are returning an empty array here, becuase there are hundreads of plate purposes in the database,
+  #   We are returning an empty array here, because there are hundreds of plate purposes in the database,
   #   and the GET query was exceeding the maximum number of bytes.
   #   Instead, the page can load, then the user can then select a purpose from the list, which will be paginated.
   def default_purposes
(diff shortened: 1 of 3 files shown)


BenTopping (approved):
Approving as tested locally and it works. 
Is there a long-term story to restore the original behaviour? I am not sure how much the users use it, but it is better UX to automatically filter the plates.

harrietc52:
> Approving as tested locally and it works. Is there a long-term story to restore the original behaviour? I am not sure how much the users use it, but it is better UX to automatically filter the plates.

Good point, yes https://github.com/sanger/limber/issues/2310

PR #2324, opened 2025-04-15
Y25-196 new cite-seq pipeline

seenanair:
 
This pull request includes several changes to the configuration files for high throughput scRNA core cDNA and library prep pipelines related to CITE-Seq pipeline. The changes primarily focus on adding new relationships, defining new pipeline groups, and configuring new robots for specific tasks. 
Also, this PR introduces some important enhancements:

- Adds CompositeWellFilteredCustomTaggedPlateCreator
This is an enhanced version of CustomTaggedPlateCreator that supports filtering of requests using provided filters when multiple requests are present. It also allows partial submissions.

- Adds WellFilterComposite
This new filter combines the functionality of both WellFilterAllowingPartials and WellFilter. It can filter requests based on their state when multiple request types are present, and also correctly handles scenarios where only a single request type is in the pending state.

 Below are the most important changes:

### 1. Well Filtering Enhancements:
* Added `CompositeWellFilteredCustomTaggedPlateCreator` class to extend the functionality of `CustomTaggedPlate` by introducing composite well filtering capabilities. This ensures only relevant requests in wells are processed. (`app/models/labware_creators/composite_well_filtered_custom_tagged_plate_creator.rb`)
* Introduced `WellFilterComposite` class to conditionally apply partial filtering logic based on well states, enabling more granular request filtering. (`app/models/labware_creators/well_filter_composite.rb`)
* Updated `active_requests` method in `HasRequests` module to ensure request uniqueness by ID. (`app/sequencescape/sequencescape/api/v2/shared/has_requests.rb`)

### 2. Submission on LRC GEM-X 5p CITE plate:
- request_options:
        library_type: 'Chromium single cell GEM-X 5p v3 cell'
        request_type_key: limber_scrna_core_cdna_pcr_cite

### 3. Submission on LRC GEM-X 5p CITE SUP XP:
- request_options:
        library_type: 'Chromium single cell GEM-X 5p v3 cell'
        request_type_key: limber_scrna_core_library_prep

### 4. Pipeline Configuration Updates:

* Added new relationships in `scRNA Core cDNA Prep GEM-X 5p GEMs Input` to include `LRC GEM-X 5p cDNA PCR XP` and introduced a new pipeline group `scRNA Core cDNA Prep GEM-X 5p GEMs Input CITE` in `config/pipelines/high_throughput_scrna_core_cdna_prep.yml`.
* Added new pipeline groups `scRNA Core Chromium GEM-X 5p cDNA PCR CITE` and `scRNA Core Chromium GEM-X 5p CITE MX` with corresponding relationships in `config/pipelines/high_throughput_scrna_core_library_prep.yml`.

### 5. Purpose Configuration Updates:

* Defined new submission options and presenter class for `LRC GEM-X 5p cDNA PCR CITE` in `config/purposes/scrna_core_cdna_prep.yml`.
* Added new asset types and configurations for various stages such as `LRC GEM-X 5p CITE SUP`, `LRC GEM-X 5p CITE Ligation`, and `LRC GEM-X 5p CITE Lib PCR-XP` in `config/purposes/scrna_core_library_prep.yml`.

### 6. Robot Configuration Updates:

* Updated robot configuration to handle `LRC GEM-X 5p cDNA PCR (or CITE)` to `LRC GEM-X 5p cDNA PCR XP` in `config/robots.rb`.
* Added new robot configurations for tasks such as `LRC GEM-X 5p cDNA PCR CITE` to `LRC GEM-X 5p CITE SUP`, `LRC GEM-X 5p CITE SUP` to `LRC GEM-X 5p CITE SUP XP`, and `LRC GEM-X 5p CITE Ligation` to `LRC GEM-X 5p CITE Lib PCR-XP` in `config/robots.rb`.

### 7. UI changes:
* Updated `CustomTaggedPlate.vue` to include `filters` as a required prop and pass it to the backend for tag group filtering. 

#### Background
In the case of LRC GEM-X 5p CITE SUP XP, manual submission is required using the request type:
limber_scrna_core_library_prep.

The limber_scrna_core_cdna_pcr_cite requests are passed down to LRC GEM-X 5p CITE SUP XP from its upstream plate: LRC GEM-X 5p cDNA PCR CITE.
While CITE SUP XP uses StateChangers::AutomaticPlateStateChanger to close these limber_scrna_core_cdna_pcr_cite requests, they are still passed further to LRC GEM-X 5p CITE Ligation as part of the standard well transfer process.

Separately, there is a LRC GEM-X 5p CITE SUP Input entry point which passes down only limber_scrna_core_library_prep requests to Ligation.

#### Problem

The `CustomTaggedPlateCreator`, used for LRC GEM-X 5p CITE Ligation, was incorrectly using only the irrelevant `limber_scrna_core_cdna_pcr_cite` requests from the parent to create outer_requests, instead of using the correct `limber_scrna_core_library_prep` requests.

There are some existing well filters that can be used in CustomTaggedPlateCreator, but they are specific to certain scenarios:
 - WellFilterAllowingPartials only supports contexts with multiple request types in different states.
 - WellFilter supports contexts where only one request type is present.

However, in LRC GEM-X 5p CITE Ligation, both scenarios can occur: it may receive a mix of request types and states, or it may receive only a single type (e.g., from a CITE SUP Input).
Currently, there is no existing filter that can flexibly handle both situations.

#### Solution
Introduce a new well filter called WellFilterComposite:
  - Overrides the well_transfers method.
  - Dynamically chooses between:
       WellFilterAllowingPartials (when multiple request types exist with mixed states).
       Basic WellFilter (when only a single request type is present and pending).

Introduce a new labware creator called CompositeWellFilteredCustomTaggedPlateCreator:
   - Extends CustomTaggedPlateCreator.
   - Adds support for context-aware filtering of requests using a new WellFilterComposite.

This new labware creator with new hybrid/composite filter ensures that:
Only relevant requests (limber_scrna_core_library_prep) are considered for outer_requests.
Filters out irrelevant request types such as limber_scrna_core_cdna_pcr_cite
Both submission paths (via CITE SUP XP and CITE SUP Input) are correctly supported.

[Pipeline diagram ](https://lucid.app/lucidchart/2909654d-8daa-43b4-90d3-5c5da051a88d/edit?invitationId=inv_cc95f797-54b2-4b65-880e-455d29807a4a&page=0_0#)
[Integration-suite-PR](https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/239)


config/pipelines/high_throughput_scrna_core_cdna_prep.yml
```diff
@@ -23,6 +23,11 @@ scRNA Core cDNA Prep GEM-X 5p GEMs Input:
     request_type_key: limber_scrna_core_cdna_prep_gem_x_5p_gems_input
   relationships:
     LRC GEM-X 5p GEMs Input: LRC GEM-X 5p cDNA PCR
+    LRC GEM-X 5p cDNA PCR: LRC GEM-X 5p cDNA PCR XP
+scRNA Core cDNA Prep GEM-X 5p GEMs Input CITE:
+  pipeline_group: scRNA Core cDNA Prep
+  filters:
+    request_type_key: limber_scrna_core_cdna_prep_gem_x_5p_gems_input
```
andrewsparkes, line 30: should this be the _cite request type? or is it the same for this single step?

config/purposes/scrna_core_library_prep.yml
```diff
@@ -110,3 +110,36 @@ LRC GEM-X 5p GE Pool Norm:
   :type: IlluminaHtp::InitialStockTubePurpose
   :presenter_class: Presenters::SimpleTubePresenter
   :default_printer_type: :tube
+LRC GEM-X 5p CITE SUP:
+  :asset_type: plate
+  :stock_plate: false
```
andrewsparkes, line 115: think defaults are false for stock and input, but don't mind being explicit

config/pipelines/high_throughput_scrna_core_cdna_prep.yml
andrewsparkes: think it's fine

andrewsparkes (approved):
Looks good.

config/purposes/scrna_core_cdna_prep.yml
```diff
@@ -144,3 +144,11 @@ LRC GEM-X 5p cDNA PCR CITE:
...
   :stock_plate: false
   :input_plate: false
+  :presenter_class: Presenters::PermissiveSubmissionPlatePresenter
+  :submission_options:
+    CITE-Seq Branch - Automated Submission:
+      template_name: 'Limber-Htp - scRNA Core cDNA PCR CITE'
+      allowed_extra_barcodes: false
+      request_options:
+        library_type: 'Chromium single cell GEM-X 5p v3 cell surface protein'
+        request_type_key: limber_scrna_core_cdna_pcr_cite
```
yoldas, line 154: Please specify the library type in the Sequencescape request type definition. Sequencescape:
`limber_scrna_core_cdna_pcr_cite` in config/default_records/request_types/014_limber_scrna_core_cdna_prep_request_types.yml

See config/default_records/request_types/015_limber_scrna_core_library_prep_request_types.yml as an example.

yoldas (approved):
Looks great!

config/pipelines/high_throughput_scrna_core_cdna_prep.yml
seenanair: Yes, it is the same for this step

config/robots.rb
```diff
@@ -3715,12 +3715,12 @@
     # Hamilton bed verification
     # LRC GEM-X 5p cDNA PCR (or CITE) to LRC GEM-X 5p cDNA PCR XP
```
KatyTaylor, line 3716: Above comment still says 'or CITE'

config/robots.rb
```diff
@@ -3757,6 +3761,29 @@
...
+    #     },
+    #     bed(9).barcode => {
+    #       purpose: 'LRC GEM-X 5p CITE SUP',
+    #       states: ['pending'],
+    #       label: 'Bed 9',
+    #       parent: bed(13).barcode,
+    #       target_state: 'passed'
+    #     }
+    #   }
+    # )
```
KatyTaylor, line 3785: Why is this commented out one here?

app/models/labware_creators/well_filter.rb
```diff
@@ -34,7 +34,7 @@ def filtered_wells
   def filter_requests(requests, well)
     return extract_submission(well) if well.requests_as_source.empty?
 
-    filtered_requests_by_rt = filter_by_request_type(requests)
+    filtered_requests_by_rt = filter_by_request_type(requests.uniq(&:id))
```
yoldas, line 37: This is a workaround for duplication of the same request when there are multiple aliquots in the well.

seenanair: Thanks for comments Katy. 
I had a look as well. I think active_requests is a better place to be put in which ensures that duplicates from both requests_as_source and requests_in_progress are removed. Also, this allows deduplication in all methods that use active_requests.

yoldas (approved):
Looks good.

config/purposes/scrna_core_library_prep.yml
```diff
@@ -116,8 +116,9 @@ LRC GEM-X 5p CITE SUP:
   :input_plate: false
 LRC GEM-X 5p CITE SUP XP:
   :asset_type: plate
-  :stock_plate: false
-  :input_plate: false
+  :stock_plate: true # Uploaded via sample manifest.
```
yoldas, line 119: LRC GEM-X 5p CITE SUP XP is not uploaded via sample manifest; It is child of LRC GEM-X 5p CITE SUP . The stock_plate flag should be false I think.

config/purposes/scrna_core_library_prep.yml
```diff
@@ -116,8 +116,9 @@ LRC GEM-X 5p CITE SUP:
...
+  :input_plate: true # Passed when submission is built.
```
yoldas, line 120: Should we close the request `limber_scrna_core_cdna_pcr_cite` first for this work because the submission will be adding the new request `limber_scrna_core_library_prep`?

app/models/labware_creators/custom_tagged_plate.rb
```diff
@@ -28,8 +29,10 @@ class CustomTaggedPlate < Base
           :tags_per_well,
           { substitutions: {} }
         ]
+        # transfers: [[:source_plate, :source_asset, :outer_request, :pool_index, { new_target: :location }]]
       }
     ]
+    # validates :transfers, presence: true
```
KatyTaylor, line 35: Delete the above 3 comments?

config/purposes/scrna_core_library_prep.yml
```diff
@@ -110,3 +110,46 @@ LRC GEM-X 5p GE Pool Norm:
...
   :default_printer_type: :tube
+LRC GEM-X 5p CITE SUP:
+  :asset_type: plate
+  :stock_plate: false
+  :input_plate: false
+LRC GEM-X 5p CITE SUP XP:
+  :asset_type: plate
+  :stock_plate: false
+  :input_plate:
+    false # Passed when submission is built.
```
yoldas, line 121: Can the value be moved to the previous line and the comment be removed?

app/models/labware_creators/custom_tagged_plate.rb
```diff
@@ -107,5 +109,100 @@ def create_labware!
...
+        .flat_map { |well| requests_for_well(well).map { |request| { request:, well: } } }
+        .map do |transfer|
+          {
+            source_asset: transfer[:well].uuid,
+            outer_request: transfer[:request].uuid,
+            target_asset: child_plate.wells.find { |child_well| child_well.location == transfer[:well].location }&.uuid
+          }
+        end
+    end
+    # rubocop:enable Metrics/AbcSize
```
KatyTaylor, line 206: Trivial point: I would call the method `build_transfer_requests_attributes` or something, as the current name makes me think it's creating stuff in the db.

Maybe more important thoughts:

If I'm reading it correctly, this code means that if the source well has more than one associated request, it's creating a transfer request (TR) for each of them. I don't think that's how TRs are meant to work. My understanding is that the TR tracks the physical transfer of liquid from a parent well to a child well - which happens once - so doesn't really make sense to have 2 TRs with the same source and target (unless maybe 2 separate transfers actually physically happened).

I think it should probably be more like:
```
# We've got to assume there's only one relevant request to be processed here, 
# because we wouldn't know what to do with more than one.
def request_for_well(well)
  well.active_requests.first
end
```

and

```
# This assumes a 'straight stamp' from source to destination plate,
# meaning well A1 goes to well A1, A2 to A2, etc.
def create_requests_and_transfers(child_plate)
  parent
    .wells
    .map do |well|
      {
        source_asset: well.uuid,
        outer_request: request_for_well(well).uuid,
        target_asset: child_plate.wells.find { |child_well| child_well.location == well.location }&.uuid
      }
    end
end
```

I haven't tested the above by the way!

I switched to using active_requests because that prioritises 'open / incomplete' requests over 'closed / complete' ones, whereas your code treated them the same. I think most of the logic under active_requests is good, although, related to the below, I'm a bit dubious about whether it should ever bring back 'closed / complete' requests.

Couple more concerns:
1. Does it work if some wells are empty? (maybe your code does and mine doesn't)
2. How does it behave if the source plate is one with a submission on, and the submission is 'partial' - so some source wells might have closed requests going into them (which would be in the aliquots and which we want to ignore), and some source wells might have a closed request (in the aliquot) and also an open request (in requests_as_source). I guess theoretically this could be the case on the `LRC GEM-X 5p CITE SUP XP` or `LRC GEM-X 5p CITE SUP Input` plate, if they did a partial submission.

Sorry for the essay. I think this is most of the way there. Hopefully between you and the other guys you can come up with the final answers.

config/purposes/scrna_core_library_prep.yml
```diff
@@ -110,3 +110,46 @@ LRC GEM-X 5p GE Pool Norm:
...
+    # close off the submission when the plate is created
+  :state_changer_class: StateChangers::AutomaticPlateStateChanger
+  :work_completion_request_type: ['limber_scrna_core_cdna_prep_gem_x_5p_gems_input', 'limber_scrna_core_cdna_pcr_cite']
```
KatyTaylor, line 124:
```suggestion
  :work_completion_request_type: ['limber_scrna_core_cdna_pcr_cite']
```

It shouldn't need to close off gems input requests, should it? Think they go down the left-hand (gene expression) branch and get closed off on the `LRC GEM-X 5p cDNA PCR XP` plate?

KatyTaylor (changes requested):
Couple of remaining questions (including an essay, sorry).

Well done on the marathon story 💪

app/models/labware_creators/custom_tagged_plate.rb
```diff
@@ -107,5 +109,100 @@ def create_labware!
...
+    #       target_asset: "target-uuid-1"
+    #     },
+    #     {
+    #       source_asset: "source-uuid-2",
+    #       outer_request: "request-uuid-2",
+    #       target_asset: "target-uuid-2"
+    #     }
+    #   ]
+    #
+    def create_requests_and_transfers(child_plate)
```
yoldas, line 194: I wonder whether all these could be achieved if we made a new WellFIlteredCustomTaggedPlate labware creator by adding WellFilterBehaviour.

app/models/labware_creators/custom_tagged_plate.rb
seenanair: Thanks a lot for the detailed review.

1. Transfer Requests Logic
You're right, there should only be one transfer request (TR) per source/target well pair. SS actually throws an error if duplicate TRs (with the same source and target assets) are submitted.

Originally, my code was filtering duplicates with:
`transfer_requests = create_requests_and_transfers(dest_plate).uniq { |transfer| [transfer[:source_asset], transfer[:target_asset]] }`
 Using well.active_requests.first handles this more cleanly by ensuring that only a single relevant request is considered, avoiding duplicates from the start. I’ve updated the code accordingly.

2. Use of active_requests
I was initially unsure whether active_requests was the best choice, but I trust your judgement here. I've updated the code to use it as suggested. Makes sense that it prioritises open/incomplete requests over closed ones.

3. Handling Empty Wells
Yes, the original code was handling empty wells properly. The SS side doesn’t accept null values for outer_request in the payload, so I’ve modified your code therefore it will only create transfer_requests for wells that do have a non-null active request. 

4. Partial Submissions (Pending)
I haven’t yet tested how this behaves with partial submissions. I'll check in with @yoldas on Monday to confirm the expected behaviour there

Aside from the partial submission case (which I’ll follow up on), I believe all your comments have now been addressed and reflected in the updated code.
Please let me know if there’s anything I’ve missed.

app/models/labware_creators/custom_tagged_plate.rb
seenanair: Done 👍

app/models/labware_creators/custom_tagged_plate.rb
seenanair: Thanks a lot for the detailed review.

Following are the updates for your comments.

1. Transfer Requests Logic:-  You're right, there should only be one transfer request (TR) per source/target well pair. SS actually throws an error if duplicate TRs (with the same source and target assets) are submitted.
Originally, my code was filtering duplicates with: `transfer_requests = create_requests_and_transfers(dest_plate).uniq { |transfer| [transfer[:source_asset], transfer[:target_asset]] } Using w`ell.active_requests.first` handles this more cleanly by ensuring that only a single relevant request is considered, avoiding duplicates from the start. I’ve updated the code accordingly.

2. Use of `active_requests`:- I was initially unsure whether `active_requests` was the best choice, but I trust your judgement here. I've updated the code to use it as suggested. Makes sense that it prioritises open/incomplete requests over closed ones.

3. Handling Empty Wells:- Yes, the original code was handling empty wells properly. The SS side doesn’t accept null values for outer_request in the payload, so I’ve modified your code therefore it will only create transfer_requests for wells that do have a non-null active request.

5. Partial Submissions (Pending):- I haven’t yet tested how this behaves with partial submissions. I'll check in with @yoldas  on Monday to confirm the expected behaviour there.

Aside from the partial submission case (which I’ll follow up on), I believe all your comments have now been addressed and reflected in the updated code. 

Please let me know if there’s anything I’ve missed.

config/purposes/scrna_core_library_prep.yml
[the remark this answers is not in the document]
seenanair: Done 👍

app/models/labware_creators/custom_tagged_plate.rb
seenanair: Thanks for the review @yoldas 
Hmm, I'm not sure. I'm on leave tomorrow (16th May) and @KatyTaylor  on leave next week. Would you mind discussing this with her may be tomorrow (16th May), and see whether it is a better approach. Then I can work on it next week.

app/models/labware_creators/custom_tagged_plate.rb
KatyTaylor: Brilliant thank you

yoldas (approved):

Issue #2074, opened 2024-11-18
Y24-461 - Tube Rack Banking Work Completion


KatyTaylor:
**User story**

Part of the epic https://github.com/sanger/limber/issues/1713. Check the 'work completion' code works properly, to automatically pass the cell extraction requests.

**Who are the primary contacts for this story**
Katy, Andrew, Abdullah

**Who is the nominated tester for UAT**
Once https://github.com/sanger/limber/issues/1713 and its dependencies are complete, we can ask Abby if she will UAT it. Could send some screenshots for earlier feedback.

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] When the transfer into an LRC Bank Seq or Spare tube rack is made, either by bed verification or 'manual transfer', the corresponding 'scRNA Core Cell Extraction' request is passed.
- [x] Update the integration suite tests if necessary.

It is not possible to release the implementation of this story in isolation. There are epic branches in use for SS, Limber and Int Suite to branch from and to.

**Dependencies**
Story relationships are shown in the Lucidchart diagram [TubeRack Stories](https://lucid.app/lucidchart/d88bdcfd-9628-4fec-8b5a-453f680ee06b/edit?viewport_loc=-1051%2C-316%2C2770%2C1486%2C0_0&invitationId=inv_13d74286-1cc2-4c2f-aaf0-30887764601c) .

**Additional context**
The versioned tube rack model is researched in https://github.com/sanger/limber/issues/1482 and the model is summarised in the Google Document[ DPL-1011 Re-use of tube racks](https://docs.google.com/document/d/1rvPXeCk5rBEf72VkchHCD2FRmWANboke2pln_Lt_j5U/edit?usp=sharing)

This story is part of re-implementation of https://github.com/sanger/limber/issues/1314 (Banking) using the versioned tube rack model.

KatyTaylor:
Link to details about state changer from Andrew - https://github.com/sanger/limber/issues/1700#issuecomment-2479124999

andrewsparkes:
A state change of the tubes (ie. to Passed) after the transfers should also be reflected in a state change of the Tube Rack, so viewing the Tube Rack in Limber should reflect that (i.e. the rack should appear Passed too).

dasunpubudumal:
**Some notes**

- For Robot verifications, currently the state changes start from `app/models/robots/bed/plate_to_tube_racks_bed.rb`. It does not seem to go through a designated state changer.
[...]

PR #2347, opened 2025-04-30
[dev] Add Rubocop spec plugins

StephenHulme:
Improves code quality by adding cops for specs

#### Changes proposed in this pull request

- Add rspec, rspec_rails, and capybara plugins
- Lint most of the safe-auto correct suggestions
- Add lint errors to the todo that were not easy to correct

#### Instructions for Reviewers

_[All PRs] - Confirm PR template filled_  
_[Feature Branches] - Review code_  
_[Production Merges to `main`]_  
 &nbsp; &nbsp; \- _Check story numbers included_  
 &nbsp; &nbsp; \- _Check for debug code_  
 &nbsp; &nbsp; \- _Check version_

PR #2355, opened 2025-05-02
Y24-461 Work Completion for tube racks

dasunpubudumal:
Closes https://github.com/sanger/limber/issues/2074

- Send the UUID of the tube rack and let Sequencescape do the state changes instead of looping through each tube.
- Complete work completion for tube racks.


dasunpubudumal:
> Morning, I merged in #2347 yesterday, which might break cause rubocop to fail in this PR when develop is merged in. I've found some work-arounds for common failing tests like multiple-expectations, so shout if Rubocop's being difficult.

Thanks @StephenHulme

app/models/state_changers.rb
```diff
@@ -162,33 +156,29 @@ def contents_for(target_state)
...
+      racked_tubes_locations_filtered =
+        v2_labware.racked_tubes.reject { |rt| rt.tube.state == 'failed' }.map(&:coordinate)
 
       # if no tubes are in the target state then no need to send the contents subset (state changer assumes all
       #  will change)
-      return nil if racked_tubes_locations_filtered.length == labware.racked_tubes.count
+      return nil if racked_tubes_locations_filtered.length == v2_labware.racked_tubes.count
 
       # NB. if all tubes are already in the target state then this method will return an empty array
       # TODO: is this correct behaviour?
```
andrewsparkes, line 167: Is this comment still true? is it right for an empty array to be returned?
I think I'd noticed that it was calling the SS api to change state on an empty array which seemed pointless.

app/models/state_changers.rb
```diff
@@ -201,47 +191,48 @@ def contents_for(target_state)
       return nil unless FILTER_FAILS_ON.include?(target_state)
 
       # determine list of well locations requiring the state change
-      well_locations_filtered = labware.wells.reject { |w| w.state == 'failed' }.map(&:location)
+      well_locations_filtered = v2_labware.wells.reject { |w| w.state == 'failed' }.map(&:location)
```
andrewsparkes, line 194: Maybe the 'failed' here should be a constant, like FILTER_FAILS_ON.

Also, I think I wondered before why on line 191 'failed' is in FILTER_FAILS_ON (so target_state can be 'failed', 'cancelled' or 'qc_complete') but then we filter out the 'failed' wells on line 194 (but not cancelled or qc_complete). Is that deliberate?
Does anyone understand that?

spec/support/api_url_helper.rb
```diff
@@ -172,11 +195,19 @@ def expect_tube_from_tube_creation
         [double(child: child_tube)] * tube_from_tubes_attributes.size
       )
     end
+
+    def expect_work_completion_creation
+      expect_api_v2_posts('WorkCompletion', work_completions_attributes)
+    end
+
+    def don_t_expect_work_completion_creation
```
andrewsparkes, line 203: prefer do_not to don_t (line 203) or dont (line 204)

andrewsparkes (commented):
Again, looks good, but left a few things to discuss.

PR #2398, opened 2025-06-09
Kinnex tube label printing

### LINKED ISSUE #2360: Kinnex tube barcode labels
**User story**

As a member of the Short Read team using the Kinnex Prep pipeline, I would like to make sure the tube labels contain a reference to the source plate barcode and well, to help with getting the manual transfers right.

Katy, Steve, Meg

Consider does this need _User_ testing or can it be done within the team. For guidance see Confluence page "Process for deciding on what user stories need UAT (User Acceptance Testing)".

Meg, Coline

**Acceptance criteria**

- [ ] Tube barcode labels contain the source plate barcode and well position
- [ ] Tube barcode labels can be printed individually from the tube page
- [ ] Ideally, tube barcode labels can be printed in bulk from the parent plate page (but not essential if difficult)

**Dependencies**

- https://github.com/sanger/sequencescape/issues/4897
- https://github.com/sanger/limber/issues/2358
(could probably work around blockers if needed)

dasunpubudumal:
Closes #2360

- Create new label class and override `workline_identifier` to fit the source.

image


config/label_templates.yml
```diff
@@ -105,3 +105,9 @@ templates:
     :printer_type: 96 Well Plate
     :pmb_template: sqsc_96plate_label_template_code39
     :sprint_template: plate_96_2d.yml.erb
+
+  kinnex_tube:
+    :label_class: Labels::TubeLabelKinnex
+    :printer_type: 96 Well Plate
+    :pmb_template: sqsc_96plate_label_template_code39
+    :sprint_template: plate_96_2d.yml.erb
```
KatyTaylor, line 113: I think you should send the photo of the label to the lab stakeholders to ask whether this is the correct type of label. I think we're transferring into 'normal' tubes so imagine we should be using a 'tube' label rather than a 'plate' label. If that means we can't have a 2D barcode then that's probably OK, I guess they can reprint from Traction.

app/sequencescape/sequencescape/api/v2/tube.rb
```diff
@@ -10,6 +10,7 @@ class Sequencescape::Api::V2::Tube < Sequencescape::Api::V2::Base
 
   DEFAULT_INCLUDES = [
     :purpose,
+    'transfer_requests_as_target.source_asset',
```
KatyTaylor, line 13: This makes me a little nervous as it's a generic query, presumably used in many places, so could cause performance issues? Any chance you can get this to include the transfer requests in a way that is more specific to this use case?

KatyTaylor (commented):
I think the labels look great, but (should have thought to mention this earlier, sorry) I assume they need to be 'tube' labels instead - see comments. Thanks

app/sequencescape/sequencescape/api/v2/tube.rb
dasunpubudumal: Made a small fix. I don't think it's very efficient but at least it won't slow down other places where it fetches `V2::Tube` instances.

app/models/labels/tube_label_kinnex.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class Labels::TubeLabelKinnex < Labels::TubeLabel # rubocop:todo Style/Documentation
+  attr_accessor :labware_with_custom_includes
```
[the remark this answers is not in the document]
dasunpubudumal: Yep, it's not required. I'll remove it.

PR #2476, opened 2025-07-28
2427-y25-396-mark_wells_as_under_represented

wendyyang:
Closes #

- add a new UI section for marking wells as under-represented
- new plate controller method to deal with the form action when submit selected wells to mark under-represented.


KatyTaylor (commented):
I've only looked at a few of the files so far, but publishing comments now in case they're useful.

app/controllers/plates_controller.rb
```diff
@@ -8,35 +8,82 @@ class PlatesController < LabwareController
...
   end
 
+  def process_mark_under_represented_wells # rubocop:todo Metrics/AbcSize
+    return redirect_to(limber_plate_path(params[:id]), notice: t('notices.no_wells_selected')) if selected_wells.empty?
+
+    begin
+      plate = fetch_plate_with_requests(params[:id])
+      mark_selected_wells_under_represented(plate)
+      redirect_to(limber_plate_path(params[:id]), notice: t('notices.wells_marked_under_represented'))
+    rescue StandardError => e
```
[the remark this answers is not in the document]
wendyyang: I am not sure if the error occurs often in production.  Just when I was testing during development, it crashed with little information on the page. So I thought adding error handling for debugging in case something goes wrong, and provide more friendly messages on the page to let user know the process is not successful.

app/views/plates/_mark_wells_under_represented.html.erb
```diff
@@ -0,0 +1,30 @@
+<% unless presenter.passed? %>
+  <div class="alert alert-danger" role="alert">
+    Notes!!! you can only mark wells under represented <strong>after</strong> passing the plate.
+  </div>
```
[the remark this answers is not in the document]
wendyyang: Thanks for the suggestion. will reword it. I am actually not sure if the alert is true, just doing the similar thing as fail well. Maybe @KatyTaylor can confirm if the alert is necessary or not.

dasunpubudumal:
image

image

image

image

image

This is very nice! I will go through the code :)

app/controllers/plates_controller.rb
```diff
@@ -8,35 +8,82 @@ class PlatesController < LabwareController
...
+      log_plate_error(e)
+      redirect_to limber_plate_path(params[:id]), alert: t('errors.messages.mark_wells_under_represented_failed')
+      nil
+    end
+  end
```
dasunpubudumal, line 46: I think we should refrain from todo-ing Rubocop for new code we add. It does seem that [`Metrics/AbcSize`](https://docs.rubocop.org/rubocop/cops_metrics.html) is too restrictive, but I suggest making it agree with Rubocop standards rather than todo-ing it. 


```suggestion
def process_mark_under_represented_wells
  return redirect_to_no_wells_selected if selected_wells.empty?
  mark_wells_under_represented_and_redirect
end

private

def redirect_to_no_wells_selected
  redirect_to(limber_plate_path(params[:id]), notice: t('notices.no_wells_selected'))
end

def mark_wells_under_represented_and_redirect
  plate = fetch_plate_with_requests(params[:id])
  mark_selected_wells_under_represented(plate)
  redirect_to(limber_plate_path(params[:id]), notice: t('notices.wells_marked_under_represented'))
rescue StandardError => e
  log_plate_error(e)
  redirect_to(limber_plate_path(params[:id]), alert: t('errors.messages.mark_wells_under_represented_failed'))
end
```

(Personally, I don't find your function harder to read or unmaintainable. It's just one of those Rubocop things that makes it too restrictive.)

app/controllers/plates_controller.rb
```diff
@@ -8,35 +8,82 @@ class PlatesController < LabwareController
...
+  def create_poly_metadatum_for_request(well)
+    aliquot = well.aliquots.first
+    # Get the request from the aliquot
+    request = aliquot.request
+
+    # If request is an array, get the first one?
+    request = Array(request).first
+
+    Sequencescape::Api::V2::PolyMetadatum.create!(
+      key: 'under_represented',
```
dasunpubudumal, line 76: It feels like this `under_represented` must be declared as a constant that other modules within Limber could also potentially access.

app/controllers/plates_controller.rb
```diff
@@ -8,35 +8,82 @@ class PlatesController < LabwareController
...
       )
-      redirect_to(
-        limber_plate_path(params[:id]),
-        notice: 'Selected wells have been failed' # rubocop:todo Rails/I18nLocaleTexts
-      )
+      redirect_to(limber_plate_path(params[:id]), notice: t('notices.wells_failed'))
+    rescue StandardError => e
+      log_plate_error(e)
+      redirect_to limber_plate_path(params[:id]), alert: t('errors.messages.fail_wells_failed')
+      nil
```
dasunpubudumal, line 26: Is this `nil` required?

app/helpers/plate_helper.rb
```diff
@@ -25,6 +25,29 @@ def fail_wells_presenter_from(form, presenter)
     WellFailingPresenter.new(form, presenter)
   end
 
+  # Proxy object wrapping the form alongside the presenter.
+  # This allows us to use the shared plate partial, but pass the form
+  # object through to the custom aliquot partial
+  # rubocop:disable Rails/HelperInstanceVariable
```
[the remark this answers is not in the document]
wendyyang: good point!

app/controllers/plates_controller.rb
wendyyang: removed.

app/controllers/plates_controller.rb
wendyyang: refactored.

dasunpubudumal (approved):
Nice! Thank you for the changes.

PR #2485, opened 2025-07-29
FIX - added uniq on id to prevent duplicates

andrewsparkes:
Closes #


yoldas:
Using uniq on the search results on the Limber side provides a workaround for duplicate results but the real problem is having duplicates in the search results on the Sequencescape side.

PR #2496, opened 2025-08-05
Y24-190-5 - Move sequencescape API to appropiate directories

StephenHulme:
Rails has an expected heirachy of directories for code and tests, see:
- https://rspec.info/features/7-1/rspec-rails/directory-structure/
- https://github.com/jwipeout/rails-directory-structure-guide

- Rename the `sequencescape/sequencescape/api/` directories to `integrations/sequencescape/api/`
- Rename the specs at `sequencescape/api/` to `integrations/sequencescape/api/`
- Lint files

This reduces duplicate namespacing and resulting confusion, along with mirroring the file paths between source code and test code.


seenanair:
The changes look fine. 
Just to be pedantic — renaming the sequencescape/sequencescape/api/ directories to external/sequencescape/api/ is definitely an improvement over the original structure. That said, I'm not sure if 'external' is intuitive enough. Maybe something like 'services' would be a bit clearer?

seenanair:
> > The changes look fine. Just to be pedantic — renaming the sequencescape/sequencescape/api/ directories to external/sequencescape/api/ is definitely an improvement over the original structure. That said, I'm not sure if 'external' is intuitive enough. Maybe something like 'services' would be a bit clearer?
> 
> Agreed. I thought of 'services' too, but this is described in [the above link ](https://github.com/jwipeout/rails-directory-structure-guide)as user facing, rather than app facing?
> 
> > Contains app services. A service object implements the user’s interactions with the application. It contains business logic that describe the connections with your domain objects.
> 
> I'm very happy to be conviced - I do agree that 'external' while better, is not ideal.

Thanks for the clarification — I wasn’t aware of that terminology convention, makes sense. In that case, external makes sense, and I’m happy to go with it if others are aligned as well.

andrewsparkes (approved):
Looks good.
I'm also not sure about 'external' as that suggests 'outside sanger' to me. But if that's the convention, as in 'external to this app' then I can live with it.

andrewsparkes:
As Ben said, careful when you merge this as it'll conflict with developments in progress.

andrewsparkes (approved):

PR #2599, opened 2025-10-08
Ultima combined

andrewsparkes:


app/views/labware/_pooled_tube.html.erb
```diff
@@ -11,8 +11,8 @@
           <%= tube.name %>
         <% end %>
       </dd>
-    <dt>Barcode</dt>
-    <dd><%= useful_barcode(tube.barcode) %></dd>
+    <dt>Barcode</dt><dd><%= useful_barcode(tube.barcode) %></dd>
+    <dt>Purpose</dt><dd><%= tube.purpose.name %></dd>
```
BenTopping: Do we have a test covering this? Can tubes have no purpose? (legacy?)

BenTopping (commented):
Looks good from a skim through. Once comment about adding purpose name to tube list.

BenTopping (approved):

config/pipelines/ultima_high_throughtput_pcr_free.wip.yml
```diff
@@ -20,3 +20,8 @@ Ultima Multiplexing:
   relationships:
     UPF Lib XP2: UPF EqVol Pool
     UPF EqVol Pool: UPF EqVol Norm
+Ultima Rebalancing:
+  pipeline_group: Ultima High Throughput PCR Free
+  relationships:
+    UPF Lib XP2: UPF Balanced Pool
+    UPF Balanced Pool: UPF Balanced Norm
```
sabrine33: @andrewsparkes - I don't think we still need this section - am I right ?

sabrine33: we don't have` Ultima Rebalancing:` anymore in SS - also the button is checking if the downstream sequencing request is passed to be visible/or not - well this is how I remember we made it work at the end.
I will deploy the PRs again and run the integ.suite

sabrine33: @andrewsparkes - you are right this part is necessary

PR #2645, opened 2025-11-27
Y25-598 - scRNA - add tube size options

andrewsparkes:
Closes #5389 

- adds new separated pipeline routes starting with 0.8 and 1.0ml input tubes, to accommodate different lab processes for those two types of input
- defines new tube and plate purposes and pipeline configuration for the new routes
- adds a new bed verifications for the 1ml  tube route and modifies an existing one for the 0.8ml route
- adds a new Cellaca count file for the 1ml tube route
- adds a new driver file for the 1ml tube route
- change to viability threshold value 65 -> 50
- added a validation in donor_pooling_validator.rb to check you scanned the parent you clicked 'add LRC PBMC Pool plate' on
- added a filter for export file_links in the donor_pooling_presenter so we don't show download buttons for driver files for parents we don't have
- added another filter in the donor_pooling_presenter on the bed verification buttons to show only relevant ones

### DIFF
--- a/app/helpers/exports_helper.rb
+++ b/app/helpers/exports_helper.rb
@@ -50,6 +50,9 @@ module ExportsHelper
   end
 
   # Helper method for scRNA PBMC Pools plate to create rows for the driver files
+  # Used in the following exports:
+  # - Hamilton LRC PBMC Aliquot to LRC PBMC Pools CSV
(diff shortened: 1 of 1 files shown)


StephenHulme (approved):
Glad to see the `.` removed from the purpose names!
There's a lot going on here and I'm not sure I fully understand everything, but what I do understand seems fine.
One small suggestion, but looks good otherwise.
Thanks for the helpful comments!

PR #2658, opened 2025-12-18
Y25-716 - address end-to-end testing feedback

sabrine33:
Closes Y25-716

This PR is been tested by the end user already

Integ.suite PR update : https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/300

- Removing XP1 from 'End prep XP1' plate purpose
- Remove Rebalancing terminology and renaming files and methods accordingly to  prevent any future confusion.
-  setting default printers for some plate purposes

### DIFF
--- a/config/purposes/ultima_pcr_free.wip.yml
+++ b/config/purposes/ultima_pcr_free.wip.yml
@@ -23,7 +23,6 @@ UPF Lib:
   :default_printer_type: :ultima_lib_plate
 UPF Lib XP2:
   :asset_type: plate
-  :label_template: plate_xp
   :presenter_class:
     name: Presenters::PlateDownstreamCompletedPresenter
     args:


lib/tasks/config.rake
```diff
@@ -38,6 +38,10 @@ namespace :config do
         configuration[:printers] = {
           plate_a: 'g316bc',
           plate_b: 'g311bc2',
+          plate_xp2: 'aa209bc2',
+          ultima_lib_plate: 'aa212bc',
+          ultima_balanced_tube: 'aa209bc1',
+          ultima_balanced_norm_tube: 'ssrtubebc-sq1',
```
KatyTaylor: This is setting a precedent - we don't normally set default printers at a plate purpose level. If we did, then this list would be a lot longer! 

It's a pretty neat solution though, so I don't feel strongly about it.

config/purposes/ultima_pcr_free.wip.yml
```diff
@@ -10,15 +10,17 @@ UPF Shear:
   :asset_type: plate
 UPF Post Shear:
   :asset_type: plate
-UPF End Prep XP1:
+UPF End Prep:
```
KatyTaylor: Might need to think through how this data change will work in UAT, where the purposes are already created.

I expect we'll end up with the 'UPF End Prep XP1' hanging around in UAT (as would be a pain to try to delete, linked to many objects), and this will create a new 'UPF End Prep' one on deployment.

KatyTaylor (approved):
Looks good I think. Well done getting the UAT feedback actioned so quickly. Approving, just left a few questions.

KatyTaylor (approved):

PR #2700, opened 2026-02-09
Y25-682 - Add RNAseq Workflow Branch

sabrine33:
Integ. suite PR: 
https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/304

### DIFF
--- a/config/robots.rb
+++ b/config/robots.rb
@@ -5378,7 +5378,6 @@ ROBOT_CONFIG =
       name: 'Bravo LCMT RNA OdT Beads First Verification',
       require_robot: true,
       verify_robot: true,
-      start_button_text: 'Bravo LCMT RNA OdT Beads First Verification',
       beds: {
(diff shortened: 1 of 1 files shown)


config/robots.rb
```diff
@@ -5352,4 +5328,224 @@
...
+        }
+      }
+    )
+    # LCMT RNA OdT beads Verification
+    custom_robot(
+      'bravo-rnaseq-odt-beads-first-verification',
+      name: 'Bravo LCMT RNA OdT Beads First Verification',
+      require_robot: true,
+      verify_robot: true,
+      start_button_text: 'Bravo LCMT RNA OdT Beads First Verification',
```
yoldas: start_button_text is implemented for the cases when the button text should be different from the name, such as preceding the name with "Finish "; It is redundant otherwise.

PR #2706, opened 2026-02-13
Y25-578 bed verification - modify js to check for bed scan first

### LINKED ISSUE #2575: Y25-578 - [BUG] Bed verification incorrectly succeeds without bed scan
**Describe the bug**
On the bed verification layout step, if the user does not scan the bed and only scans the plate barcode, the validation succeeds, when it should fail.

**RT Ticket Number**
N/A, but this was reported by Adulrahman.T when testing the Ultima story.

**To Reproduce**
1. Go to any labware page where a robot bed layout is requested (e.g., WGS pipeline, LB Post Shea plate).
2. Within the Robots section, click on LB Post Shea → LB End Prep.
3. On the bed verification page, scan only the plate barcode.
4. Click on Validate Layout.

**Expected behaviour**
The validation should fail if the bed is not scanned

**Screenshots**

Image

**Additional context**
This is can be tested by the developer.

andrewsparkes:
Closes #2575 

Modify JS to check for bed scan first


app/frontend/entrypoints/pages/bed_verification.spec.js
```diff
@@ -0,0 +1,46 @@
+import $ from 'jquery'
+import SCAPE from '@/javascript/lib/global_message_system.js'
+
+describe('bed_verification plate scan handler', () => {
+  let plateScan, bedScan, robotScan
+
+  // NB. use of beforeAll is not ideal but could not get beforeEach to work.
```
BenTopping: You could add an `afterEach` hook too to reset the values after each test if that is the main concern.

BenTopping (approved):
Works well and has clear user instructions 👍

PR #2732, opened 2026-03-05
Y26-005 - LCM Triomics - assign Project / Cost Code to automated submissions

KatyTaylor:
Closes #

- Quite functional code made by Github Copilot. Adds a field to the choose workflow component, which allows you to search by project id. Adds project uuid to the payload sent to Sequencescape to create the submission - this part doesn't work yet. To add - only display this field when configured to do so. Take hardcoded project out of submission template. Improve styling / UX.
- (Update: works now)


KatyTudor:
Someone requested me to review something on here but think it should be @KatyTaylor

StephenHulme:
There are a couple strange this about this PR... a Javascript test written as a Ruby spec is also strange... Can you comment on what has been created by co-pilot and what hasn't?

wendyyang:
Enclose a few screenshots:


**when enter a valid project code and click enter:**
Screenshot 2026-04-08 at 22 24 49



**when enter non-existing project code and click enter:**
Screenshot 2026-04-08 at 22 25 03



**when leave the field empty and click submission button:**
Screenshot 2026-04-08 at 22 25 18



**when the project code is valid and click the submission button:**
Screenshot 2026-04-08 at 22 26 42

app/frontend/entrypoints/pages/choose_workflow.js
```diff
@@ -1,3 +1,109 @@
...
+    event.preventDefault()
+    const projectCode = globalInput.value.trim()
+
+    if (!projectCode) {
+      showError('Project / Cost Code is required')
+      return
+    }
+
+    await lookupProject(projectCode)
+  })
```
StephenHulme, line 85: Nice, this is a good way of looking up the project. Some people might be using this on a tablet, or have a different manner of using forms. Unless there is a need to trigger on an enter key only (a barcode-scanner for example), may I suggest widening the events that would trigger the project lookup? Possible ones to include are [blur](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event), [onchange](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event), and maybe submit? (although that last one might take us into tricky form territory)

wendyyang: > Nice, this is a good way of looking up the project. Some people might be using this on a tablet, or have a different manner of using forms. Unless there is a need to trigger on an enter key only (a barcode-scanner for example), may I suggest widening the events that would trigger the project lookup? Possible ones to include are [blur](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event), [onchange](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event), and maybe submit? (although that last one might take us into tricky form territory)

good suggestion! added 'blur' option.

PR #2779, opened 2026-04-08
Y26-063 - Quad stamp scan fails when plate has no active requests

### LINKED ISSUE #2728: Y26-063 - [BUG] Bioscan - Quad stamp plate scans are valid despite no active requests
**Describe the bug**
The Quad (multistamp) stamping page for the PCR 1 plate creator in the Bioscan pipeline lets you scan Lysate plates without active requests on the wells and flags the plates as valid as you scan them. However, no wells are valid to be transferred so the view on the left shows a blank 384-well plate.

**To Reproduce**
1. Perform a bioscan run (complete the run, you can use the bioscan integration suite test)
2. Return to one of the Lysate plates
3. If https://github.com/sanger/limber/issues/2727 has not been done yet you can create a PCR 1 plate using the button, if it has then you'll need a library prep submission
4. Find yourself on the quad stamp page where you can scan up to 4 lysate plates
5. scan a plate that DOES NOT have an active library prep submission, it shows as valid but no wells are highlighted for the plate on the left

**Expected behaviour**
When you scan the lysate plate without any active requests it should be invalid, and give an error message that there are no wells with active requests 

**Additional context**
There is a well filter in multi-stamping to select for only wells with active requests. This is intended, to allow for partial submissions where you only want to submit some of the wells. 
The plate scan should error and say the plate does not have any active requests and is not ready for combining. There is a similar issue with another quad stamp step in another pipeline (BGE when combining XP2 plates) so we could fix it for both pipelines.

BenTopping:
Closes #2728

- Adds 'checkMinCountRequests' to MultiStamp to ensure at least one active library creation request exists on scanned plates.
- Adds 'requireActiveLibraryRequests' config attribute to multi stamp purposes. We need to do config driven instead of enforcing for every MultiStamp plate as some are not at the start of library prep so will fail the requests check.

### DIFF
--- a/app/frontend/javascript/multi-stamp/components/MultiStamp.spec.js
+++ b/app/frontend/javascript/multi-stamp/components/MultiStamp.spec.js
@@ -3,7 +3,8 @@ import { shallowMount } from '@vue/test-utils'
 import MultiStamp from './MultiStamp.vue'
 import itBehavesLikeMultiStamp from './shared_examples/multi_stamp_instance.shared_spec'
 import { flushPromises } from '@vue/test-utils'
-import { plateFactory } from '@/javascript/test_support/factories.js'
+import { plateFactory, wellFactory, requestFactory } from '@/javascript/test_support/factories.js'
+import { aggregate } from '@/javascript/shared/components/scanValidators.js'
 
 describe('MultiStamp', () => {
   itBehavesLikeMultiStamp({ subject: MultiStamp })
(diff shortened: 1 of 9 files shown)


app/frontend/javascript/shared/components/tubeScanValidators.spec.js
```diff
@@ -36,18 +36,6 @@ describe('checkDuplicates', () => {
     })
   })
 
-  it.skip('fails if there are duplicate tubes even when the parent has not been updated', () => {
```
BenTopping, line 39: I assume at one point this was dependent on the event but this is no longer the case so this is redundant. It has also been skipped for 7 years.

app/frontend/javascript/multi-stamp/components/MultiStamp.spec.js
```diff
@@ -23,6 +24,7 @@ describe('MultiStamp', () => {
           locationObj: options.locationObj,
           transfersLayout: 'quadrant',
           transfersCreator: 'multi-stamp',
+          requireActiveLibraryRequests: 'false',
```
andrewsparkes, line 27: nice, optional flag

app/frontend/javascript/multi-stamp/components/MultiStamp.vue
```diff
@@ -270,11 +280,19 @@ export default {
...
+
+      const validators = [
         checkSize(12, 8),
         checkDuplicates(currPlates),
         checkForUnacceptablePlatePurpose(this.acceptablePurposesArray),
       ]
+      // We can't enforce this check for all multi stamp plates because some may not be the first step of the workflow,
+      // and therefore may not have library requests yet. So we make it optional to check if the scanned plates have active library requests.
+      if (this.requireActiveLibraryRequests === 'true') {
+        validators.push(checkMinCountRequests(1))
```
andrewsparkes, line 292: I think the minimum number of requests that's acceptable for a scanned plate is always going to be 1, did you think of a scenario where it might be a bigger number?

andrewsparkes (approved):
Looks good, minor question about whether minimum number of requests would ever be more than 1.

app/frontend/javascript/multi-stamp/components/MultiStamp.vue
BenTopping: Im reusing this method from somewhere else so not 100% sure why it was made to accept a parameter (1). There is another similar method called `checkMaxCountRequests(x)` which makes more sense to have a parameter so maybe its keeping inline with that.

PR #2795, opened 2026-04-16
ci: reduce push triggering ci actions to just develop and master

BenTopping:

- Reduces ci test actions running on push to just master and develop

### DIFF
--- a/.github/workflows/compile_test.yml
+++ b/.github/workflows/compile_test.yml
@@ -2,6 +2,9 @@ name: Test compilation
 
 on:
   push:
+    branches:
+      - develop
+      - master
   pull_request:
 
 env:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,6 +5,9 @@ env:
 
 on:
   push:
+    branches:
+      - develop
+      - master
   pull_request:
 
 jobs:
--- a/.github/workflows/test_js.yml
+++ b/.github/workflows/test_js.yml
@@ -3,6 +3,9 @@ name: Unit tests
 on:
   # Warning: changing run conditions could prevent coverage report in PR, see codecov.yml for `after_n_builds`
   push:
+    branches:
+      - develop
+      - master
   pull_request:
 
 jobs:
--- a/.github/workflows/test_ruby.yml
+++ b/.github/workflows/test_ruby.yml
@@ -2,6 +2,9 @@ name: Unit tests
 
 on:
   push:
+    branches:
+      - develop
+      - master
   pull_request:
 
 env:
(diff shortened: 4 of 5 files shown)


StephenHulme:
I would suggest leaving Limber as is? It runs many fewer tests than sequencescape, for a shorter time - meaning that it doesn't block the runners nearly as much?

BenTopping:
> I would suggest leaving Limber as is? It runs many fewer tests than sequencescape, for a shorter time - meaning that it doesn't block the runners nearly as much?

I think this still has value by reducing redundant compute / actions. The push actions and pull request are running the same tests (except pull request is based on the merge with the base) so I dont see the need to run both.

StephenHulme:
> > I would suggest leaving Limber as is? It runs many fewer tests than sequencescape, for a shorter time - meaning that it doesn't block the runners nearly as much?
> 
> I think this still has value by reducing redundant compute / actions. The push actions and pull request are running the same tests (except pull request is based on the merge with the base) so I dont see the need to run both.

I wouldn't agree that they are redudant, in that they provide different information. A PR can pass for the `push` but fail for the `pull_request` if there are incompatible changes already in develop`. This particular bit of knowledge can help debug subtle problems arising from multiple people working on the repo simulatenously.

I'm happy to be overriden on this, but may I suggest we trial with the Sequencescape changes already merged in and see how it goes?
It's worth noting that we have [20 runners](https://github.com/organizations/sanger/settings/actions/hosted-runners). One push on Sequencescape can use all of them for at least 4 minutes. Whereas Limber only has [two tests that run longer than a minute](https://github.com/sanger/limber/pull/2794), leaving much spare capacity.

BenTopping:
> > > I would suggest leaving Limber as is? It runs many fewer tests than sequencescape, for a shorter time - meaning that it doesn't block the runners nearly as much?
> > 
> > 
> > I think this still has value by reducing redundant compute / actions. The push actions and pull request are running the same tests (except pull request is based on the merge with the base) so I dont see the need to run both.
> 
> I wouldn't agree that they are redudant, in that they provide different information. A PR can pass for the `push` but fail for the `pull_request` if there are incompatible changes already in develop`. This particular bit of knowledge can help debug subtle problems arising from multiple people working on the repo simulatenously.
> 
> I'm happy to be overriden on this, but may I suggest we trial with the Sequencescape changes already merged in and see how it goes? It's worth noting that we have [20 runners](https://github.com/organizations/sanger/settings/actions/hosted-runners). One push on Sequencescape can use all of them for at least 4 minutes. Whereas Limber only has [two tests that run longer than a minute](https://github.com/sanger/limber/pull/2794), leaving much spare capacity.

Okay thats makes sense. The counter argument to that is that we should always be update to date with the base when getting ready to merge which would make pull and push the same. Happy to trial the SS one and see how it goes.

BenTopping:
Closing in favour of keeping push to provide more instant feedback.

PR #2814, opened 2026-04-28
[js] Upgrade to Vite 8

StephenHulme:
Upgrades incrementally from Vite 5 to Vite 8.

While Vite 8 uses the new Rust-based `rolldown` to generate a significant compile-time decrease, we don't benefit from this performance improvement due to the css building that occurs.

- Upgrade brower targets to latest versions (might impact compiled code, but shouldn't be negatice)
- Upgrade vite-plugin-ruby
- Build JS for modern browsers as well as WinXP browsers
- Upgrade vitest and @vitest/coverage-v8 with config changes
- Repair unresolved import path

See [Sequencescape equivalent](https://github.com/sanger/sequencescape/pull/5730)

### DIFF
--- a/app/frontend/javascript/pool-xp-tube-panel/components/PoolXPTubeSubmitPanel.spec.js
+++ b/app/frontend/javascript/pool-xp-tube-panel/components/PoolXPTubeSubmitPanel.spec.js
@@ -417,7 +417,6 @@ describe('PoolXPTubeSubmitPanel', () => {
       mockTubeFoundResponse()
       wrapper = createWrapper()
       await flushPromises()
-      expect(global.fetch).toBeCalledTimes(1)
       expect(global.fetch).toHaveBeenCalledWith(wrapper.vm.tractionTubeCheckUrl)
 
       expect(wrapper.vm.state).toBe('tube_exists')
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ doc/
 /coverage
 /spec/examples.txt
 lcov.info
+# vitest test cache file
+results.json
 
 # Tempfiles
 tmp/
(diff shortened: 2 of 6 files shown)


StephenHulme:
Currently in [UAT](https://uat.limber.psd.sanger.ac.uk/plates/a2ba17c2-1620-11f0-9c83-02427e19310d) - and has passed int-suite

app/frontend/javascript/pool-xp-tube-panel/components/PoolXPTubeSubmitPanel.spec.js
```diff
@@ -417,7 +417,6 @@ describe('PoolXPTubeSubmitPanel', () => {
       mockTubeFoundResponse()
       wrapper = createWrapper()
       await flushPromises()
-      expect(global.fetch).toBeCalledTimes(1)
```
BenTopping, line 420: Was this removed because its called more than once or because it was redundant?

StephenHulme: It was called 55 times... I think because the built-in 1000ms delay has been short-circuited as part of the test and vitest 4 handles awaits and promises slightly differently? I wasn't able to figure a good solution in a short amount of time, and because the code is excellently documented and the behaviour isn't breaking (🤞) I figured it was easier to remove the check.

BenTopping: Okay, maybe worth checking manually in the int suite we aren't making lots of calls in this step.

StephenHulme: Yup, looks fine. About 1 call per second for about 100 seconds.

Screenshot 2026-04-29 at 10 54 30

BenTopping (approved):

PR #2828, opened 2026-05-05
feat(robots): adds scrna core bed verification

BenTopping:
Closes #2761

Int suite PR: https://gitlab.internal.sanger.ac.uk/psd/integration-suite/-/merge_requests/315

- Adds LRC GEM-X 5p GE LigXP => LRC GEM-X 5p GE PCR 2XP bed verification.


KatyTaylor (approved):

yoldas (approved):

yoldas:
Note that the config in this PR does not require entering the robot barcode (No require_robot: true). It is not requested in the story, so I assume that is correct. 

Neither it defines if they want to lock in to the same robot with the previous step (No verify_robot: true)

PR #2879, opened 2026-06-11
Y26-181 - Improve nanoseq performance

BenTopping:
SS Story https://github.com/sanger/sequencescape/issues/5839

- add metadatable_type to metadata lookup to improve query speed
- Pull all metadata for a request once instead of each request metadata key being a separate query.

Sequencecape PR: https://github.com/sanger/sequencescape/pull/5850

#### Notes
The main sluggishness comes from the poly_metadata checks and creation.
- Each polymetadata key / request combo produced a separate request to sequencescape. Since this was not taking advantage of the index of the table this was taking ~1s second per request in UAT.

The changes here allow us to make use of the Sequencescape index on poly_metadata. It also fixes (an unlikely) bug in which we are currently requesting on just metadatable_id which could be the same across classes and doesn't take into account type.
We also reduce N+1 queries by getting all polymetadata for a request at once instead of individually requesting one each time. This could cause issues if a request has _mass_ amounts of polymetadata but I cant see that so far in the existing records.

This process still takes ~20s which isn't ideal but before it was taking ~150s+.


spec/models/labware_creators/pcr_cycles_binned_plate_for_t_nano_seq_spec.rb
```diff
@@ -1561,6 +1559,7 @@
       end
 
       it 'raises an error when existing metadata fails to update' do
+        request_metadata['key1'] = 'new_value1'
```
BenTopping, line 1570: Not sure how this was passing before as there is a check to ensure update is only called if there is a new value.

StephenHulme (approved):
Looks good - thanks for the fix!

andrewsparkes (approved):

PR #2943, opened 2026-07-31
Pipeline visualiser

yoldas:
Closes #

- Pipeline visualiser MVP


app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.21.1/cytoscape.min.js"></script>
```
BenTopping: We shouldn't need to use a CDN as we should have these javascript libraries available via our node package manager (yarn)

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
...
+
+      <div id="graph" style="width: 100%; height: 600px; border: 1px solid #ccc;"></div>
+
+      <div id="error-message" class="alert alert-danger mt-3" style="display: none;"></div>
+      <div id="loading" class="alert alert-info mt-3" style="display: none;">Loading...</div>
+    </div>
+  </div>
+</div>
+
+<script>
```
BenTopping: Suggestion: I would move this to a separate javascript file and import that into this file.

app/views/pipeline_visualiser/show.html.erb
BenTopping, line 1: Suggestion: I wonder if this would be better as a vue component due to the amount of reactivity, it might make it easier to handle the requests etc.

app/controllers/pipeline_visualiser_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    return render_not_found unless labware
+
+    @labware_data = {
+      record: labware,
+      state: decide_state(labware),
+      ancestors: labware.ancestors
+    }
+
+    respond_to do |format|
+      format.html
```
BenTopping, line 11: When I first load the page using a barcode I am not being presented with the graph of the barcode straight away and I am having to search for it. Is this a bug? Should it be automatically populated?

app/controllers/pipeline_visualiser_controller.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+# Controller for visualizing a single labware's pipeline workflow
+class PipelineVisualiserController < ApplicationController
+  def show
+    barcode = params[:id]
+    labware = retrieve_labware_by_barcode(barcode)
+
+    return render_not_found unless labware
```
BenTopping: I think if you go to this URL without the barcode it should still load, I don't think we should require the barcode. Currently it throws an error

app/controllers/pipeline_visualiser_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+  def labware_to_cytoscape_graph(labware)
+    all_labware = build_labware_chain(labware)
+    nodes = all_labware.map { |item| build_node(item, searched: item.uuid == labware.uuid) }
+    edges = build_edges(all_labware)
+
+    { elements: nodes + edges }
+  end
+
+  def build_labware_chain(labware)
+    (labware.ancestors || []).reverse + [labware]
```
BenTopping: This will only shows ancestors (parent plates). Do we also want to show children plates or is that out of scope?

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
...
+          style: [
+            {
+              selector: 'node',
+              style: {
+                'label': 'data(label)',
+                'background-color': '#6272a4',
+                'color': '#fff',
+                'text-valign': 'center',
+                'text-halign': 'center',
+                'width': 180,
```
BenTopping: I have run into some cases where the barcode text overflows the container

Image

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
...
+        <div style="display: flex; align-items: center; gap: 6px;">
+          <div style="width: 14px; height: 14px; border-radius: 50%; background: #6272a4;"></div>
+          <span style="font-size: 13px;">Ancestor</span>
+        </div>
+      </div>
+
+      <div id="graph" style="width: 100%; height: 600px; border: 1px solid #ccc;"></div>
+
+      <div id="error-message" class="alert alert-danger mt-3" style="display: none;"></div>
+      <div id="loading" class="alert alert-info mt-3" style="display: none;">Loading...</div>
```
BenTopping: Suggestion: This alert and error-message might be easier to read if they are at the top of the page instead.

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
...
+      <h1>Pipeline Visualiser</h1>
+
+      <div class="form-group mb-3">
+        <label for="barcode-input">Enter Barcode:</label>
+        <div class="input-group">
+          <input
+            type="text"
+            id="barcode-input"
+            class="form-control"
+            placeholder="e.g., LP123456789">
```
BenTopping: Suggestion: real placeholder value - SQPP-12345

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,115 @@
...
+          <button id="search-btn" class="btn btn-primary">Search</button>
+        </div>
+      </div>
+
+      <div style="display: flex; gap: 12px; margin-bottom: 8px;">
+        <div style="display: flex; align-items: center; gap: 6px;">
+          <div style="width: 14px; height: 14px; border-radius: 50%; background: #e8a838;"></div>
+          <span style="font-size: 13px;">Searched plate</span>
```
BenTopping: Suggestion: `Searched labware` as it could be a plate or tube

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -0,0 +1,39 @@
...
+
+      <div style="display: flex; gap: 12px; margin-bottom: 8px;">
+        <div style="display: flex; align-items: center; gap: 6px;">
+          <div style="width: 14px; height: 14px; border-radius: 50%; background: #666; border: 2px solid #e8a838;"></div>
+          <span style="font-size: 13px;">Searched labware</span>
+        </div>
+      </div>
+
+      <div class="mb-2">
+        <strong class="small">Pipelines Key</strong>
```
BenTopping: Suggestion: These show the Labware purposes rather than the pipelines so might be worth renaming to `Purpose types`

BenTopping (changes requested):
Looks good, nice work. Works really nicely even with complex branched pipelines! Couple outstanding comments, happy to discuss.

app/controllers/pipeline_visualiser_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+        { plates: %w[uuid purpose labware_barcode state_changes updated_at parents children] },
+        { tubes: %w[uuid purpose labware_barcode state_changes updated_at parents children] }
+      )
+      .includes(:state_changes, :purpose, :parents, :children)
+      .where(barcode:)
+      .first
+  end
+
+  # Fetch a labware fresh by uuid so its parents/children associations are loaded
+  def fetch_with_relatives(uuid)
```
StephenHulme, line 40: suggestion: rename to `retrieve_labware_by_uuid` and update the comment of these functions since they seem to do near identical things.

config/routes.rb
```diff
@@ -102,5 +102,8 @@
 
   resources :sequencescape_submissions
 
+  resources :pipeline_visualiser, only: :show
+  get 'pipeline_visualiser', to: 'pipeline_visualiser#show'
```
StephenHulme, line 106: suggestion: try condense these into a single line like
```rb
get 'pipeline_visualiser(/:id)', to: 'pipeline_visualiser#show', as: :pipeline_visualiser
```
or add a comment explaining that the resources line handles the `/<id>` endpoint and the get line handles the no-id case.

README.md
```diff
@@ -254,6 +255,21 @@ This file is the start point for a set of files documenting how pipelines are co
...
+
+Search for a barcode to see:
+
+- the selected labware highlighted in the graph
+- parent labware shown above it
+- downstream labware shown below it
+- coloured arrows showing which pipeline connects each step
+
+Use the Pipelines Key to match each arrow colour to its pipeline name.
+
```
StephenHulme, line 272: praise: Yay docs 😁 
(and thanks for updating the table of contents 🙏)

app/controllers/pipeline_visualiser_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  def render_graph_json(barcode)
+    if barcode.blank?
+      render json: { graph_data: { elements: [] } }
+    elsif @labware
+      render json: { graph_data: labware_to_cytoscape_graph(@labware) }
+    else
+      render json: { error: 'Labware not found' }, status: :not_found
+    end
+  end
```
StephenHulme, line 26: suggestion: would be good to have these lines covered by tests (including an example output at a request level)

StephenHulme (approved):
This looks good and I can see it being really useful to both lab staff and developers 🎉 

Thanks for putting this together and for the changes. I've added a few comments, but they are all bonus items - I think there is more value in releasing this soon to help those in new roles than tweaking code to get the same result.

A feature-request for a follow-up PR: auto-scale and freeze the cytoscape scrollpane/canvas as I found I that scrolling down to the bottom of the page caused it to zoom out instead of move down.

StephenHulme:
(The ERB linting error can be easily fixed with the `--autocorrect` ERB lint command in https://github.com/sanger/limber#linting-and-formatting)

app/views/pipeline_visualiser/show.html.erb
```diff
@@ -2,6 +2,8 @@
   <div class="row">
     <div class="col-md-12">
       <h1>Pipeline Visualiser</h1>
+      <div id="error-message" class="alert alert-danger mt-3" style="display: none;"></div>
+      <div id="loading" class="alert alert-info mt-3" style="display: none;">Loading...</div>
```
BenTopping: Nice UX, good feedback for the user!

spec/spec_helper.rb
```diff
@@ -65,6 +68,7 @@
   options.add_argument('--window-size=1600,3200')
   options.add_argument('--no-sandbox')
   options.add_preference('profile.password_manager_leak_detection', false)
+  options.binary = ENV['CHROME_BINARY_PATH'] if ENV['CHROME_BINARY_PATH']
```
BenTopping: Is this the work around you mentioned for local CI?

I think our CI is pinned to an old version of chrome because at the time selenium didn't support the later versions properly. Suspect that is fixed now.

BenTopping (approved):
LGTM! Nice work.

spec/spec_helper.rb
yoldas: Yes, it is just a local workaround if used using an environment variable. I think, the pin is at https://github.com/sanger/limber/blob/pipeline_visualiser/.github/workflows/test_ruby.yml#L30 . I will look into that separately.

